PR #48, opened 2019-10-27
feat: added default test cased from official Vue repo

juzhiyuan:
Add default test cases from official Vue repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,12 +5,12 @@
   "author": "Chong Guo <armourcy@gmail.com>",
   "scripts": {
     "serve": "concurrently \"vue-cli-service serve\"",
+    "test:unit": "jest --clearCache && vue-cli-service test:unit",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -54,10 +55,12 @@
     "@vue/test-utils": "^1.0.0-beta.29",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.2",
+    "chromedriver": "78",
```
juzhiyuan: @moonming  We have some new dependencies for E2E test cases, so if I need to provide the License for them?

moonming: I think you can discuss license issues in mailing list.
 IMO, we only need add license for bundle codes in repo, not dependencies.

moonming: it's not need to add license if dependencies not in release packaege.

juzhiyuan: I see

PR #62, opened 2019-11-03
fix: add new features and function enhance

lilien1010:
1、when enter the router list ,the Action column fixed to the right and always inside view.
2、keep router upstream service list ordered by ID, which I think the ID is generated by ETCD,so it should be like ordered by created date desc, it will be helpful when routers or upstreams are more then 10,
3、support multi hosts and multi uris when create a router，which need APISIX lua be updated which my next PR，when current `schema_def.lua` is not support uris

fix: https://github.com/apache/incubator-apisix-dashboard/issues/59
fix: https://github.com/apache/incubator-apisix-dashboard/issues/58
fix: https://github.com/apache/incubator-apisix-dashboard/issues/57

### DIFF
--- a/src/views/schema/routes/edit.vue
+++ b/src/views/schema/routes/edit.vue
@@ -36,23 +36,45 @@
       </el-form-item>
 
       <el-form-item
-        label="URI"
-        prop="uri"
+        label="URIs"
+        prop="uris"
+        placeholder="can write multi uri here"
       >
-        <el-input
-          v-model="form.uri"
-          placeholder="URI"
-        />
+        <el-select
+          v-model="form.uris"
+          allow-create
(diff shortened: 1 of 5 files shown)


juzhiyuan:
Thanks for your contribution, PR will be merged once I build & test the codes successfully today.

lilien1010:
https://github.com/apache/incubator-apisix/pull/797
this PR need to be merged after this PR is merged, since the code in apisix doesn't support uris before my PR

juzhiyuan:
Tests will be in action when the APISIX demo instance is deployed with the latest codes.

juzhiyuan:
Hi @membphis, please let me know once you update the demo instance :)

PR #65, opened 2019-11-04
enhance: can choose chash keys from drop down list

lilien1010:
fix: https://github.com/apache/incubator-apisix-dashboard/issues/64

### DIFF
--- a/src/views/schema/upstream/edit.vue
+++ b/src/views/schema/upstream/edit.vue
@@ -56,11 +56,23 @@
       <el-form-item
         label="Key"
       >
-        <el-input
+        <el-select
           v-model="form.key"
           placeholder="Input a Key"
+          allow-create
+          filterable
+          default-first-option
+          value-key="form.key"
+          clearable
           :disabled="form.type !== 'chash'"
-        />
+        >
+          <el-option
+            v-for="item in defaultHashKeys"
+            :key="item.value"
+            :label="item.label"
+            :value="item.value"
+          />
+        </el-select>
       </el-form-item>
 
       <el-form-item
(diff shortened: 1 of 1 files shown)


src/views/schema/upstream/edit.vue
```diff
@@ -155,6 +167,49 @@ export default class extends Vue {
...
+    {
+      value: 'remote_port',
+      label: 'remote_port'
+    },
+    {
+      value: 'hostname',
+      label: 'hostname'
+    },
+    {
+      value: 'arg_id',
```
membphis: `arg_id` is wrong, here is the Nginx variable index: http://nginx.org/en/docs/http/ngx_http_core_module.html#var_arg_

`arg_name`: argument name in the request line, the value of the name here is not fixed, eg: `arg_id`, `arg_name`, `arg_age`.

juzhiyuan:
Thanks for your contribution, but does this change need API support? maybe API instance is using the older codes.

image

src/views/schema/upstream/edit.vue
lilien1010: I put the arg_id here just reminder  that arg_*** is able to input here

lilien1010:
@juzhiyuan  you need upgrade APISIX to master ,which conatin my lastest code

membphis:
> @juzhiyuan you need upgrade APISIX to master ,which conatin my lastest code

Please wait for more time, I need @moonming's help.

juzhiyuan:
> @juzhiyuan you need upgrade APISIX to master ,which conatin my lastest code

I see, so we need the demo instance to be updated.

juzhiyuan:
> > @juzhiyuan you need upgrade APISIX to master ,which conatin my lastest code
> 
> Please wait for more time, I need @moonming's help.

Sure, take your time :)

src/views/schema/upstream/edit.vue
juzhiyuan: It looks everything here is ok, I will approve soon.

PR #77, opened 2019-11-09
fix configuration not show in route edit page when route uri and host from old format

lilien1010:
fix: https://github.com/apache/incubator-apisix-dashboard/issues/74


lilien1010:
@juzhiyuan  check this as ASAP, other wise the former route configuration will not show in the route edit page while all `uri` change into `uris`,same to `hosts` field

PR #82, opened 2019-12-04
fix run "npm run serve" command, notice "Request failed with status c…

YazaiHu:
fix run "npm run serve" command, notice "Request failed with status code 500"

because the apisix server default listen 9080 on develop environment.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -28,7 +28,7 @@ module.exports = {
   productionSourceMap: false,
   devServer: {
     // https://cli.vuejs.org/zh/config/#devserver-proxy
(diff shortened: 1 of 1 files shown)


YazaiHu:
when NODE_ENV=development,use the command will notice error, because this set:
devServer: {
    // https://cli.vuejs.org/zh/config/#devserver-proxy
    proxy: 'http://127.0.0.1/apisix/admin/'
  },

proxy should set "proxy: 'http://127.0.0.1:9080/apisix/admin/'", because the apisix server default listen 9080 port.

PR #83, opened 2019-12-07
Fix Favicon doesn't show in history mode

juzhiyuan:
resolve https://github.com/apache/incubator-apisix/issues/939

### DIFF
--- a/src/layout/components/Sidebar/SidebarLogo.vue
+++ b/src/layout/components/Sidebar/SidebarLogo.vue
@@ -37,7 +37,7 @@
         to="/"
       >
         <img
-          src="favicon.ico"
+          src="/favicon.ico"
           class="sidebar-logo"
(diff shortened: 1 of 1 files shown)


lilien1010:
I think this won't fix the issue too, becos when people visit 
`http:/xxx.org/apisix/dashboard#/schema/routes/list`
the icon url changed into `http://xxx.org/favicon.ico`
 but the facicon.ico is in the dashboard directory.

I think we need change this from server side ,rediret
`apisix/dashboard` into `apisix/dashboard/`

lilien1010:
@membphis 
 should we add a `location` in `/usr/local/apisix`?
```
location =/favicon.ico {
   alias dashboard/
}
```

juzhiyuan:
static assets should be proxied properly.

PR #110, opened 2020-01-03
fix apache apisix icon not show issue

lilien1010:
get ico link from index.html ，so that the ico link keep Consistent


src/layout/components/Sidebar/SidebarLogo.vue
```diff
@@ -69,6 +69,29 @@ export default class extends Vue {
   @Prop({ required: true }) private collapse!: boolean
 
   private title = 'APISIX'
+
+  get getIco() {
```
juzhiyuan: We could use an elegant way to fix this issue... Let me check.

juzhiyuan:
After several attempts, I find that the icon only disappears when we use `history` mode, the Dashboard is using `hash` mode by default.

If using `history` mode suits your needs, please update the codes in `SidebarLogo.vue` file: Just use `favicon.ico` instead of `/favicon.ico`

image

src/layout/components/Sidebar/SidebarLogo.vue
lilien1010: sure thisPR is too complicated, but I think use the same link from link.rel will keep the Icon link consistent

juzhiyuan: I'd love to know if the latest version of dashboard is working on your side?

PR #158, opened 2020-03-17
Adding api key for the request

### LINKED ISSUE #137: Adding X-API-KEY for api request
Our API for dashboard is using API Key to auth the request, please add the API Key Header in global request handler. Just like [1] for reference. Please note, this key should be added by `config` file, such as `.env` file. I recommend fetching this key by `fetchAPIKey` API.

[1] https://github.com/apache/incubator-apisix-dashboard/commit/b3b306550e8e5153d42d943d2d652914e2ad88e3#diff-084c3d9c2786b7cd963be84e40a38725R32

sshniro:
Fix #137

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -4,6 +4,7 @@
  */
 import { extend } from 'umi-request';
 import { notification } from 'antd';
+import IConfig from '../../config/config'
 
 const codeMessage = {
   200: '服务器成功返回请求的数据。',
@@ -22,6 +23,10 @@ const codeMessage = {
   503: '服务不可用，服务器暂时过载或维护。',
   504: '网关超时。',
 };
+
+const fetchAPIKey = (() => {
+  return IConfig['API_KEY'];
+});
 
 /**
  * 异常处理程序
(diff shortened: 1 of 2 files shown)


src/utils/request.ts
```diff
@@ -51,6 +56,9 @@ const errorHandler = (error: { response: Response }): Response => {
 const request = extend({
   errorHandler, // 默认错误处理
   credentials: 'include', // 默认请求是否带上cookie
+  headers: {
+    'X-API-KEY': fetchAPIKey()
```
juzhiyuan: emm `IConfig` is just a typing, we could use:

```js
import {API_KEY} from '../../config/config'

headers: {
    'X-API-KEY': API_KEY
}
```

directly here.

juzhiyuan: and because the APISIX doesn't support fetching APIKEY , so we could use the fixed value now.

juzhiyuan: sorry, there should be

image

We just exported the API_KEY standalone, do not need put it in the default config value.

sshniro: Thank you for the detailed explanation!

juzhiyuan:
@sshniro Thank you for your contribution!

BTW, we'd better use `resolve #137` instead of `fix` :)

sshniro:
Sure! @juzhiyuan

PR #175, opened 2020-04-17
feature: add docker support 

### LINKED ISSUE #155: Docker support & pass env when building

bzp2010:
resolve #155 

To build and run dashboard in container.

> This is still a draft, and there are still some problems to be solved


bzp2010:
### Updates 2020-04-18: 
Instead of setting the Admin API URL through docker image `ENV`, set the page mode in the dashboard, and save the content in the local storage of the browser.

> Maybe you can review the dashboard modification first.

config/config.ts
```diff
@@ -172,6 +178,8 @@ export default {
     REACT_APP_ENV: REACT_APP_ENV || false,
     ANT_DESIGN_PRO_ONLY_DO_NOT_USE_IN_YOUR_PRODUCTION:
       ANT_DESIGN_PRO_ONLY_DO_NOT_USE_IN_YOUR_PRODUCTION || '', // preview.pro.ant.design only do not use in your production ; preview.pro.ant.design 专用环境变量，请不要在你的项目中使用它。
+    ADMIN_API: 'https://apisix.iresty.com/apisix/admin/',
```
juzhiyuan, line 181: `https://apisix.iresty.com/apisix/admin/` is for test locally, and we'd better use `http://127.0.0.1:9080/apisix/admin/` as default.

config/config.ts
```diff
@@ -172,6 +178,8 @@ export default {
...
+    API_KEY: 'edd1c9f034335f136f87ad84b625c8f1'
```
juzhiyuan, line 182: Please keep it empty in the repo :)

src/locales/en-US/settings.ts
```diff
@@ -1,60 +1,3 @@
...
-    'Messages from other users will be notified in the form of a station letter',
-  'app.settings.notification.messages': 'System Messages',
-  'app.settings.notification.messages-description':
-    'System messages will be notified in the form of a station letter',
-  'app.settings.notification.todo': 'To-do Notification',
-  'app.settings.notification.todo-description':
-    'The to-do list will be notified in the form of a letter from the station',
-  'app.settings.open': 'Open',
-  'app.settings.close': 'Close',
+  'app.settings.fieldAdminAPIInvalid': 'Invalid Admin API'
```
juzhiyuan, line 2: `app.settings.description.invalid-admin-api`

src/locales/zh-CN/settings.ts
```diff
@@ -1,55 +1,3 @@
...
-  'app.settings.binding.bind': '绑定',
-  'app.settings.notification.password': '账户密码',
-  'app.settings.notification.password-description': '其他用户的消息将以站内信的形式通知',
-  'app.settings.notification.messages': '系统消息',
-  'app.settings.notification.messages-description': '系统消息将以站内信的形式通知',
-  'app.settings.notification.todo': '待办任务',
-  'app.settings.notification.todo-description': '待办任务将以站内信的形式通知',
-  'app.settings.open': '开',
-  'app.settings.close': '关',
+  'app.settings.fieldAdminAPIInvalid': '非法的管理API'
```
juzhiyuan, line 2: ,

src/pages/Settings.tsx
```diff
@@ -0,0 +1,81 @@
...
+    offset: 2,
+  },
+};
+
+const Settings: React.FC = () => {
+  const [form] = useForm();
+
+  useEffect(() => {
+    form.setFieldsValue({
+      admin_api: getAdminAPI(),
```
juzhiyuan, line 29: snake format?

src/pages/Settings.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <PageHeaderWrapper>
+      <Card>
+        <Form {...layout} form={form} onFinish={onFinish}>
+          <Form.Item
+            label="Admin API"
+            name="admin_api"
+            rules={[
+              {required: true, message: formatMessage({id: 'app.settings.fieldAdminAPIInvalid'})},
```
juzhiyuan, line 53: ,

src/services/ssl.ts
```diff
@@ -2,19 +2,19 @@ import request from '@/utils/request';
 import { transformFetchListData, transformFetchItemData } from '@/transforms/global';
 import { SSL } from '@/models/ssl';
 
-export const fetchList = () => request('/api/ssl').then(data => transformFetchListData<SSL>(data));
```
juzhiyuan, line 5: the umi proxy will take care this part.

src/utils/utils.ts
```diff
@@ -76,3 +76,11 @@ export const getPageMode = (): PageMode => {
 
   return 'VIEW';
 };
+
+export const getAdminAPI = (): string => {
```
juzhiyuan, line 80: duplicate data origin for Admin API & Key, please keep getting api & key from config file.

juzhiyuan: Just keep it as simple as possible :)

src/typings.d.ts
```diff
@@ -37,6 +37,9 @@ declare let ANT_DESIGN_PRO_ONLY_DO_NOT_USE_IN_YOUR_PRODUCTION: 'site' | undefine
 
 declare const REACT_APP_ENV: 'test' | 'dev' | 'pre' | false;
 
+declare let ADMIN_API: '';
```
juzhiyuan, line 40: 🤔

juzhiyuan: keep getting config data from config file is a simple way in my opinion.

src/pages/Settings.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  useEffect(() => {
+    form.setFieldsValue({
+      admin_api: getAdminAPI(),
+      admin_api_key: getAdminAPIKey()
+    })
+  });
+
+  const onFinish = (values: any) => {
+    localStorage.setItem('admin_api', values.admin_api);
```
juzhiyuan, line 35: Please use global var like `GLOBAL_ADMIN_API` here.

juzhiyuan: and use camelCase format like `values.adminAPI` here.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+# phase-build
+FROM node:12-alpine as builder
```
juzhiyuan, line 2: as builder?

juzhiyuan:
Thanks for your contribution!

> Can we change base and publicpath to root?
We should access the dashboard by `http://xxx/dashboard` as default.

And I recommend passing API_KEY by Docker env or `.env` file, or maybe others cannot config API_KEY easily... I'm not sure 😳

config/config.ts
bzp2010: Ok

config/config.ts
bzp2010: The original value like this. XD
Changed it.

src/services/ssl.ts
bzp2010: But in fact, the compiled dashboard still request `/api/ssl`, and apisix can't respond correctly.
This may be a configuration error?

src/typings.d.ts
bzp2010: This is only used to declare constant types, not specific values.

src/utils/utils.ts
bzp2010: But this will cause every file used to be imported in the form of `import xxx from '../../config'` 
The `config.ts` is not in `src` fold. It can't be imported in `import xxx from '@/config.ts'` 
Define an function can help us to get configuration from one place and change it easy.

> In addition, constants in configuration files such as apikey are now automatically replaced at compile time by a umi plugin. That means we don't even need to import configuration files.

Dockerfile
bzp2010: That command is used to distinguish phase. Every `FROM` will be builded in different images.
The `FROM xxx as builder` and `COPY --from builder` is used to copy files in different images.

bzp2010:
> Thanks for your contribution!
> 
> > Can we change base and publicpath to root?
> > We should access the dashboard by `http://xxx/dashboard` as default.
> 
> And I recommend passing API_KEY by Docker env or `.env` file, or maybe others cannot config API_KEY easily... I'm not sure 😳

The way we use it now is to compile it first and then put it into nginx image. This means that we can only enter data into the image during compilation, not during runtime. At present, I only want to create the setting page and inject the configuration into the local storage of the browser to modify the configuration without recompiling the docker image. (The docker image used for compilation is very large. (1.3GB))

If there are better suggestions, we can continue the discussion.

src/pages/Settings.tsx
bzp2010: like `get_admin_api()`?

juzhiyuan: haha no, we use camelCase format here, so it should be `adminAPI: xxx`

src/services/ssl.ts
juzhiyuan: Oh yes, you are right 👍 

https://pro.ant.design/docs/deploy-cn

juzhiyuan: BTW, if you change this line, we should update the proxy config too, so it could work in development environment.

src/utils/utils.ts
juzhiyuan: ~how about adding alias for config file? so we could import it like '@config'? ~

all right, please go ahead,

src/typings.d.ts
juzhiyuan: then could you please add a constant.ts file?

Dockerfile
juzhiyuan: 👍

juzhiyuan:
> > Thanks for your contribution!
> > > Can we change base and publicpath to root?
> > > We should access the dashboard by `http://xxx/dashboard` as default.
> > 
> > 
> > And I recommend passing API_KEY by Docker env or `.env` file, or maybe others cannot config API_KEY easily... I'm not sure 😳
> 
> The way we use it now is to compile it first and then put it into nginx image. This means that we can only enter data into the image during compilation, not during runtime. At present, I only want to create the setting page and inject the configuration into the local storage of the browser to modify the configuration without recompiling the docker image. (The docker image used for compilation is very large. (1.3GB))
> 
> If you do not use the `/dashboard` path, I can simplify the rewrite configuration in `nginx.conf`.
> 
> If there are better suggestions, we can continue the discussion.

Ok, the only thing I'm not sure is that for users who use it first time, how they could config the Key? just by open Dev tools and set Key in localStorage manually?

src/utils/utils.ts
bzp2010: > like `@/utils/config`.  But, `/config` is compile time configuration, `/src/xxxx` is runtime.
> 
> In fact, I use DefinePlugin by webpack, [description link](https://webpack.js.org/plugins/define-plugin/).
> First, declare it in the configuration file, `{ADMIN_API: 'xxxxx'}`, it will automatically replace the part of the program that uses the `ADMIN_API` at compile time without import, Such as
> `let adminAPI = ADMIN_API` will be compiled to `let adminAPI = 'xxxxx'`.

> write this function `getAdminAPI()`, because it's used many times, in order to reduce redundant code, I think it's best to do so.

src/typings.d.ts
bzp2010: Oh, it's only used to declare a type. like `declare let ADMIN_API: string;`
It will not be compiled into code for IDE use only.

bzp2010:
> > > Thanks for your contribution!
> > > > Can we change base and publicpath to root?
> > > > We should access the dashboard by `http://xxx/dashboard` as default.
> > > 
> > > 
> > > And I recommend passing API_KEY by Docker env or `.env` file, or maybe others cannot config API_KEY easily... I'm not sure 😳
> > 
> > 
> > The way we use it now is to compile it first and then put it into nginx image. This means that we can only enter data into the image during compilation, not during runtime. At present, I only want to create the setting page and inject the configuration into the local storage of the browser to modify the configuration without recompiling the docker image. (The docker image used for compilation is very large. (1.3GB))
> > If you do not use the `/dashboard` path, I can simplify the rewrite configuration in `nginx.conf`.
> > If there are better suggestions, we can continue the discussion.
> 
> Ok, the only thing I'm not sure is that for users who use it first time, how they could config the Key? just by open Dev tools and set Key in localStorage manually?

You may be able to automatically open the settings page when you detect that the apisix admin api is not set in the local storage of the browser?

bzp2010:
You may be able to automatically open the settings page when you detect that the apisix admin api is not set in the local storage of the browser?

src/services/ssl.ts
bzp2010: I still need to study the different configurations in the development and the production.

juzhiyuan: https://pro.ant.design/docs/mock-api-cn

src/utils/utils.ts
juzhiyuan: ok, take your opinion :)

src/services/ssl.ts
juzhiyuan: Please keep the `api` prefix, then I will merge it soon :)

Here is an Nginx conf for reference: https://pro.ant.design/docs/deploy-cn

bzp2010:
Oh, I have not completed the current pull request. 😳

bzp2010:
I will create a new pull request for those changes.

juzhiyuan:
image

I noticed that this PR has been marked as Ready to review... 

Ok, please send a new PR 👍

bzp2010:
@juzhiyuan Build image for test if possible. Test whether dockers in different environments can run correctly.

juzhiyuan:
> @juzhiyuan Build image for test if possible. Test whether dockers in different environments can run correctly.

@moonming We event could build an image and send it to DockerHub? then `incubator-apisix` dons not need to include submodule?

PR #177, opened 2020-04-19
feature: add new setting items with i18n content

bzp2010:
add new setting items with i18n content for AdminAPIConfig

### DIFF
--- a/src/locales/zh-CN/settings.ts
+++ b/src/locales/zh-CN/settings.ts
@@ -1,3 +1,5 @@
 export default {
-  'app.settings.description.invalid-admin-api': '非法的管理API'
+  'app.settings.description.invalid-admin-api-schema': '非法的管理API协议',
+  'app.settings.description.invalid-admin-api-host': '非法的管理API地址',
+  'app.settings.description.invalid-admin-api-path': '非法的管理API路径',
 };
--- a/src/locales/zh-CN/settings.ts
+++ b/src/locales/zh-CN/settings.ts
@@ -1,5 +1,9 @@
 export default {
-  'app.settings.description.invalid-admin-api-schema': '非法的管理API协议',
-  'app.settings.description.invalid-admin-api-host': '非法的管理API地址',
-  'app.settings.description.invalid-admin-api-path': '非法的管理API路径',
+  'app.settings.item.admin-api-schema': '管理 API 协议',
+  'app.settings.item.admin-api-host': '管理 API 地址',
+  'app.settings.item.admin-api-path': '管理 API 路径',
+  'app.settings.item.admin-api-key': '管理 API 密钥',
+  'app.settings.description.invalid-admin-api-schema': '非法的管理 API 协议',
+  'app.settings.description.invalid-admin-api-host': '非法的管理 API 地址',
+  'app.settings.description.invalid-admin-api-path': '非法的管理 API 路径',
 };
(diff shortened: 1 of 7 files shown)


src/pages/Settings.tsx
```diff
@@ -1,14 +1,16 @@
 import React, {useEffect} from "react";
 import {useForm} from "antd/es/form/util";
-import {Button, Card, Form, Input, notification} from "antd";
+import {Button, Card, Form, Input, notification, Select} from "antd";
 import {formatMessage, FormattedMessage} from "umi-plugin-react/locale";
 import {router} from "umi";
 import {PageHeaderWrapper} from "@ant-design/pro-layout";
-import {getAdminAPI, getAdminAPIKey} from "@/utils/utils";
+import {getAdminAPIConfig} from "@/utils/utils";
```
juzhiyuan, line 7: It's not a good idea to name as `/utils/utils`, please use something like `/utils/setting` or `/utils/config` instead.

juzhiyuan: or use `/utils` file directly.

src/pages/Settings.tsx
```diff
@@ -17,22 +19,27 @@ const layout = {
...
-    offset: 2,
+    offset: 4,
   },
 };
 
 const Settings: React.FC = () => {
   const [form] = useForm();
 
   useEffect(() => {
+    let adminAPIConfig = getAdminAPIConfig();
```
juzhiyuan, line 30: please use ES6 destructing here https://es6.ruanyifeng.com/#docs/destructuring

```js
const {adminAPISchema, adminAPIHost, ....} = getAdminAPIConfig()
```

juzhiyuan: maybe use `schema, host, path, key` instead of `adminAPIxxx`?

src/pages/Settings.tsx
```diff
@@ -47,10 +54,33 @@ const Settings: React.FC = () => {
       <Card>
         <Form {...layout} form={form} onFinish={onFinish}>
           <Form.Item
-            label="Admin API"
-            name="adminAPI"
+            label="Admin API Schema"
+            name="adminAPISchema"
```
juzhiyuan, line 58: i18n

src/locales/zh-CN/settings.ts
```diff
@@ -1,3 +1,5 @@
 export default {
-  'app.settings.description.invalid-admin-api': '非法的管理API'
+  'app.settings.description.invalid-admin-api-schema': '非法的管理API协议',
```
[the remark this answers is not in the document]
bzp2010: Ok

src/pages/Settings.tsx
bzp2010: Although we should use the ES6 standard, this may cause the host or path to conflict with other content in the future.

bzp2010: `schema, host, path, key` will be changed to

src/pages/Settings.tsx
bzp2010: Ok

src/pages/Settings.tsx
bzp2010: use `/utils/setting`

juzhiyuan:
@bzp2010 Is this PR ready to merge?

bzp2010:
Yes, it's ready!

Issue #176, opened 2020-04-18
CD for Dashboard


juzhiyuan:
Holding for ZEIT's reply

bzp2010:
Have any reply?

juzhiyuan:
I'm not sure if GitHub pages support API rewrite 🤔just holding on and waiting for the Netlify's reply, thanks!

juzhiyuan:
Apache allows us to use Netlify!!! I will configure it those days!

https://issues.apache.org/jira/browse/INFRA-20262

bzp2010:
That's really good. I'm going to try out the features of Netlify.

juzhiyuan:
just for reference
1. https://docs.netlify.com/routing/redirects/rewrites-proxies/
2. https://github.com/ant-design/ant-design-pro/commit/8065f52038b6a4628317a35acecd604d7f9c09e8#diff-5db06fd2327543bbb72119cd1e5761cfL1

bzp2010:
I read the document of the `umi-request` carefully. I found that the `prefix` property can be set when creating the request instance. At the same time, the `prefix` value can be controlled through ENV. It can help us unify the path configuration in the service.

In this way, the `dev` and `deploy` environment do not need to create different service files.

I will do some tests and submit the changes of this part if there are no problems.

juzhiyuan:
Ok, go ahead!

PR #190, opened 2020-05-16
feature: compatible development and deployment environment request

bzp2010:
Add a request path prefix to compatible development and deployment environment.
Prepare For CD #176

### DIFF
--- a/src/services/ssl.ts
+++ b/src/services/ssl.ts
@@ -2,19 +2,19 @@ import request from '@/utils/request';
 import { transformFetchListData, transformFetchItemData } from '@/transforms/global';
 import { SSL } from '@/models/ssl';
 
-export const fetchList = () => request('/api/ssl').then(data => transformFetchListData<SSL>(data));
+export const fetchList = () => request('/ssl').then((data) => transformFetchListData<SSL>(data));
 
 export const fetchItem = (key: string) =>
-  request(`/api/ssl/${key}`).then(data => transformFetchItemData<SSL>(data));
+  request(`/ssl/${key}`).then((data) => transformFetchItemData<SSL>(data));
 
-export const remove = (key: string) => request.delete(`/api/ssl/${key}`);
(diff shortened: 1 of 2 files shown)


juzhiyuan:
Please rebase master so you could debug your changes on Netlify

juzhiyuan:
Also, I think we should still keep the `api` prefix:
1. we may use Mock server in the future;
2. we need use `api` prefix to let some services know which request wants access API server. 🤔

bzp2010:
[change here](https://github.com/apache/incubator-apisix-dashboard/pull/190/files#diff-084c3d9c2786b7cd963be84e40a38725R54)
The global request prefix is set here. The specific value is determined according to `process.env.NODE_ENV`. The `/api` will not be added forward in the deployment environment.
It won't interfere with the use of mock environment.

juzhiyuan:
🤔that's to say... the request will add `api` prefix in development, but not in production? then how could we distinct which request is accessing the API server?

bzp2010:
**For API server, using umi proxy to forward or request directly is transparent and will not affect API server.**

In development, we use `yarn run start` to start devServer, in there `process.env.NODE_ENV` value is `development`.
In deploy, we use `yarn run build` to compile, in there `process.env.NODE_ENV` value is `production`.
It will be automate completely.(Maybe)

juzhiyuan:
from your code changes, the request will add `api` prefix in development mode, but not in production mode, for any of those two modes, we still need to provide some prefix for Nginx or other services to check which one is accessing API server 🤔

Maybe I misunderstand your thoughts, we could have a WeChat call when you have time :)

bzp2010:
Yes, I also think we need to have in-depth communication on this issue. We can't accurately convey the original meaning by means of words.

Time can be today or tomorrow. I have free time at night.

juzhiyuan:
OK, you can call me directly between 09:00 and 22:00 tomorrow :)

PR #208, opened 2020-05-21
feature: change request prefix

bzp2010:
change request prefix for different environment

### DIFF
--- a/config/config.ts
+++ b/config/config.ts
@@ -4,7 +4,13 @@ import defaultSettings from './defaultSettings';
 import proxy from './proxy';
 import routes from './routes';
 
-const { REACT_APP_ENV } = process.env;
+const { REACT_APP_ENV, NETLIFY_DEMO, NODE_ENV } = process.env;
+
+const getRequestPrefix = () => {
+  if (NETLIFY_DEMO) return '/api';
+  if (NODE_ENV === 'development') return '/api';
+  return '';
+}
 
 export default defineConfig({
   hash: true,
@@ -37,6 +43,7 @@ export default defineConfig({
     ADMIN_API_HOST: '127.0.0.1:9080',
     ADMIN_API_PATH: '/apisix/admin/',
     API_KEY: '',
+    API_REQUEST_PREFIX: getRequestPrefix(),
   },
   // Theme for antd: https://ant.design/docs/react/customize-theme-cn
   theme: {
--- a/config/config.ts
+++ b/config/config.ts
@@ -10,7 +10,7 @@ const getRequestPrefix = () => {
   if (NETLIFY_DEMO) return '/api';
   if (NODE_ENV === 'development') return '/api';
   return '';
-}
+};
 
 export default defineConfig({
   hash: true,
@@ -41,7 +41,7 @@ export default defineConfig({
     REACT_APP_ENV: REACT_APP_ENV || false,
     ADMIN_API_SCHEMA: 'http',
     ADMIN_API_HOST: '127.0.0.1:9080',
-    ADMIN_API_PATH: '/apisix/admin/',
+    ADMIN_API_PATH: '/apisix/admin',
     API_KEY: '',
     API_REQUEST_PREFIX: getRequestPrefix(),
   },
--- a/config/config.ts
+++ b/config/config.ts
@@ -4,10 +4,9 @@ import defaultSettings from './defaultSettings';
 import proxy from './proxy';
 import routes from './routes';
 
-const { REACT_APP_ENV, NETLIFY_DEMO, NODE_ENV } = process.env;
+const { REACT_APP_ENV, NODE_ENV } = process.env;
 
 const getRequestPrefix = () => {
-  if (NETLIFY_DEMO) return '/api';
   if (NODE_ENV === 'development') return '/api';
   return '';
 };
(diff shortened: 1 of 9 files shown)


src/app.tsx
```diff
@@ -31,14 +31,14 @@ export async function getInitialState(): Promise<{
 }
 
 export const layout = ({
-  initialState,
-}: {
+                         initialState,
```
juzhiyuan, line 34: codes clean

config/config.ts
```diff
@@ -2,7 +2,13 @@ import { defineConfig } from 'umi';
 import defaultSettings from './defaultSettings';
 import proxy from './proxy';
 
-const { REACT_APP_ENV } = process.env;
+const { REACT_APP_ENV, NETLIFY_DEMO, NODE_ENV } = process.env;
+
+const getRequestPrefix = () => {
+  if (NETLIFY_DEMO) return '/api';
```
juzhiyuan, line 8: it seems not to work on Netlify...

image

juzhiyuan: image

src/app.tsx
```diff
@@ -31,14 +31,14 @@ export async function getInitialState(): Promise<{
...
+                       }: {
   initialState: { settings?: LayoutSettings };
 }): BasicLayoutProps => {
   return {
-    rightContentRender: () => <RightContent />,
+    rightContentRender: () => <RightContent/>,
     disableContentMargin: false,
-    footerRender: () => <Footer />,
+    footerRender: () => <Footer/>,
```
juzhiyuan, line 41: lint codes too

src/app.tsx
```diff
@@ -85,12 +85,10 @@ const errorHandler = (error: { response: Response; data: any }): Promise<Respons
 
 const adminAPIConfig = getAdminAPIConfig();
 export const request: RequestConfig = {
+  prefix: `${adminAPIConfig.schema}://${adminAPIConfig.host}${adminAPIConfig.path}${adminAPIConfig.prefix}`,
```
juzhiyuan, line 88: how about this?

```js
const {key, xxx, xxx} = getAdminAPIConfig()
```

src/typings.d.ts
```diff
@@ -37,11 +37,17 @@ declare let ANT_DESIGN_PRO_ONLY_DO_NOT_USE_IN_YOUR_PRODUCTION: 'site' | undefine
 
 declare const REACT_APP_ENV: 'test' | 'dev' | 'pre' | false;
 
-declare let ADMIN_API_SCHEMA; let ADMIN_API_HOST; let ADMIN_API_PATH; let API_KEY: '';
+declare let ADMIN_API_SCHEMA;
+declare let ADMIN_API_HOST;
+declare let ADMIN_API_PATH;
```
juzhiyuan, line 42: I'm not sure why ADMIN_API_PATH doesn't have initial data, but API_KEY has..

config/config.ts
juzhiyuan: https://deploy-preview-208--apisix-dashboard.netlify.app/dashboard/settings

juzhiyuan:
I still think this way is not good ... We could do some changes on the proxy to meet our needs, though we have to keep duplicate settings 🤔

src/app.tsx
bzp2010: It is expected that this function will be cancelled in the later modification of the setting page.

src/typings.d.ts
bzp2010: This is just a definition to help the IDE know the variable type.

config/config.ts
bzp2010: In fact, I think we can directly request the Admin API at the browser without using netlifly's 
request forwarding.

bzp2010:
> I still think this way is not good ... We could do some changes on the proxy to meet our needs, though we have to keep duplicate settings 🤔

Using forwarding leads to complexity in various deployment environments (self compiling, docker, nginx or other web server), and the cost of direct request in browser is low.
At the same time, when deploying in the Intranet environment, we should consider how to forward and authenticate on the Admin API, and should not modify the dashboard to meet the special needs.

juzhiyuan:
image

still not woking.

juzhiyuan:
https://deploy-preview-208--apisix-dashboard.netlify.app/dashboard/settings

juzhiyuan:
> > I still think this way is not good ... We could do some changes on the proxy to meet our needs, though we have to keep duplicate settings 🤔
> 
> Using forwarding leads to complexity in various deployment environments (self compiling, docker, nginx or other web server), and the cost of direct request in browser is low.
> At the same time, when deploying in the Intranet environment, we should consider how to forward and authenticate on the Admin API, and should not modify the dashboard to meet the special needs.

The way this PR using is not clear for new developers and users :(

bzp2010:
> image
> 
> still not woking.

What should it be like?🤔

juzhiyuan:
> > image
> > still not woking.
> 
> What should it be like?🤔

should be `https://apisix.iresty.com/apisix/admin/ssl` ...

`api` prefix is used for Nginx or Netlify to reforward the request to real API server.

juzhiyuan:
Because the current request is sent from Browser to API server directly, so just removed the `api` prefix I think.

bzp2010:
> Because the current request is sent from Browser to API server directly, so just removed the `api` prefix I think.

That's rigth. I will change it.

juzhiyuan:
Please handle the conflict files, thanks!

juzhiyuan:
Because we will separate every module, so we will use something like `SSLModule` namespace in ts.

PR #259, opened 2020-06-15
i18n: add Chinese and English tips for plugins.

moonming:


juzhiyuan:
image

@moonming Please take a look

PR #288, opened 2020-06-29
fix logo missing problem

ko-han:
1. Changing logo url path depends on node environment. This should avoid logo missing in production environment.
1. Little changes on login and logout behaviour.


juzhiyuan:
@ko-han Hi, would you like to have a try on the latest dashboard which is built with React and Ant Design? it's on the master branch.

PR #293, opened 2020-07-01
Fix incorrect format of some go source files

imjoey:
Use `gofmt` command to format codes, as well as re-organize the imports in a conventional way via `goimports` command. 

> BTW: I guess that would be great if we add these two command to make static check for go codes. Maybe that could be done in another PR.

### DIFF
--- a/api/conf/conf.go
+++ b/api/conf/conf.go
@@ -18,11 +18,12 @@ package conf
 
 import (
 	"fmt"
-	"github.com/tidwall/gjson"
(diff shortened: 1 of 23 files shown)


juzhiyuan:
@gxthrj

imjoey:
Hi @gxthrj , PTAL. Many thanks.

imjoey:
@juzhiyuan @gxthrj  Thank you very much.

PR #320, opened 2020-07-14
feat: bump dependencies version

bzp2010:
Upgrade dependencies with upstream Ant Design Pro project.


juzhiyuan:
from V5?

bzp2010:
> from V5?

Yes, It come from V5 branch.

package.json
```diff
@@ -44,10 +44,15 @@
...
+  "browserslist": [
+    "> 1%",
+    "last 2 versions",
+    "not ie <= 10"
+  ],
   "dependencies": {
-    "@ant-design/icons": "^4.2.1",
-    "@ant-design/pro-layout": "6.0.0-2",
-    "@ant-design/pro-table": "^2.3.3",
+    "@ant-design/icons": "^4.0.0",
```
juzhiyuan, line 53: I'm not sure if we should use some older versions...

bzp2010: `^` Should mean the maximum version greater than 4.0.0.

In [yarn.lock](https://github.com/apache/incubator-apisix-dashboard/blob/af555bc839759dca1b5268ecf523e789436a1fbc/yarn.lock#L30) can prove it.

PR #336, opened 2020-07-23
feat(i18n): set module

TikWind:
i18n set


src/pages/Setting/locales/en-US.ts
```diff
@@ -0,0 +1,8 @@
+export default {
+  'set.update.configuration.successfully': 'Update Configuration Successfully',
```
juzhiyuan, line 2: Starts with setting please

TikWind: I have changed.

PR #337, opened 2020-07-23
WIP:feat(i18n): other module

TikWind:
i18n other module


src/services/user.ts
```diff
@@ -8,8 +9,11 @@ export async function query() {
 }
 
 export async function queryCurrent(): Promise<API.CurrentUser> {
+
```
juzhiyuan, line 12: I think there will have an error...

juzhiyuan: Because useIntl is a hook, and hook could only be executed in Redux Function Component.

PR #343, opened 2020-07-28
feat(i18n): actionbar component

TikWind:
i18n actionbar component

### DIFF
--- a/src/locales/en-US.ts
+++ b/src/locales/en-US.ts
@@ -1,4 +1,5 @@
 import { PluginFormEnUS } from '@/components/PluginForm';
+import { ActionBarEnUS } from '@/components/ActionBar';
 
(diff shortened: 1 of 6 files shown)


src/locales/en-US.ts
```diff
@@ -1,4 +1,5 @@
 import { PluginFormEnUS } from '@/components/PluginForm';
+import { ActionBarEnUS } from '@/components/ActionBar';
```
juzhiyuan, line 2: Don't need this line any more, right?

TikWind: In this case, it is need.
image
line 4 and 5 don't need, but line 1 and 2 maybe need.

juzhiyuan: OK

PR #345, opened 2020-07-28
feat(i18n): pluginpage component

TikWind:
i18n pluginpage component


src/components/PluginPage/locales/en-US.ts
```diff
@@ -1,163 +1,172 @@
 export default {
-  'PluginForm.plugin.limit-conn.desc': 'Limit the number of concurrent connections',
-  'PluginForm.plugin.limit-conn.property.conn': 'conn',
-  'PluginForm.plugin.limit-conn.property.conn.extra': 'Maximum number of concurrent connections',
-  'PluginForm.plugin.limit-conn.property.burst': 'burst',
-  'PluginForm.plugin.limit-conn.property.burst.extra':
+  'pluginpage.plugin.limit-conn.desc': 'Limit the number of concurrent connections',
+  'pluginpage.plugin.limit-conn.property.conn': 'conn',
+  'pluginpage.plugin.limit-conn.property.conn.extra': 'Maximum number of concurrent connections',
```
juzhiyuan, line 4: Please use PluginPage instead.

TikWind: Can I use lower case ?

juzhiyuan: Better not, we may obey the `ModuleName.key1.key2` style.

TikWind: The demo in umijs maybe all lower case.

juzhiyuan: pluginpage is grouped by plugin with page, we need some ways to let developers know clearly what it is.

TikWind: OK

TikWind: I have changed "pluginpage" to "PluginPage".

Issue #341, opened 2020-07-28
route setting  bug


longshang:
Set advanced route matching criteria, such as access_ Token on MySQL is correct, but the front end displays access.
mysql  as 
![image](https://user-images.githubusercontent.com/17903581/88616601-253c4e80-d0c7-11ea-8bce-32f987a18819.png)

the front end as
![image](https://user-images.githubusercontent.com/17903581/88616629-35ecc480-d0c7-11ea-89ab-cdfa564d7e86.png)

membphis:
@litesun it seems a bug

LiteSun:
Problem solved.

PR #347, opened 2020-07-29
fix: transform vars error

LiteSun:
- fix: MatchingRules wrong  unequal value
- fix: transform vars error

#341


membphis:
you can add an action name like `fix` or `close`, then the related issue will be closed automatically after your PR was merged. 

```
fix #341
```

image

PR #370, opened 2020-08-10
feat: Unified access entrance, only the dashboard port is exposed to …

gxthrj:
…the outside

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bugfix
- Description

- How to fix?

___
- Unified access entrance, only the dashboard port is exposed to the outside.

### DIFF
--- a/compose/dashboard_conf/nginx.conf
+++ b/compose/dashboard_conf/nginx.conf
@@ -1,0 +1,21 @@
+server {
+    listen 80;
+    # gzip config
+    gzip on;
+    gzip_min_length 1k;
+    gzip_comp_level 9;
+    gzip_types text/plain text/css text/javascript application/json application/javascript application/x-javascript application/xml;
+    gzip_vary on;
+    gzip_disable "MSIE [1-6]\.";
+
+    root /usr/share/nginx/html;
+    include /etc/nginx/mime.types;
+
+    location / {
+        try_files $uri $uri/ /index.html;
+    }
+
+    location /apisix/admin {
+        proxy_pass http://manager:8080/apisix/admin;
+    }
+}
--- a/compose/dashboard_conf/nginx.conf
+++ b/compose/dashboard_conf/nginx.conf
@@ -19,3 +19,4 @@ server {
         proxy_pass http://manager:8080/apisix/admin;
     }
 }
+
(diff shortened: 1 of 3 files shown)


compose/dashboard_conf/nginx.conf
```diff
@@ -0,0 +1,21 @@
...
+    include /etc/nginx/mime.types;
+
+    location / {
+        try_files $uri $uri/ /index.html;
+    }
+
+    location /apisix/admin {
+        proxy_pass http://manager:8080/apisix/admin;
+    }
+}
```
membphis, line 21: need a blank line at the end of file

compose/dashboard_conf/nginx.conf
```diff
@@ -0,0 +1,21 @@
+server {
+    listen 80;
+    # gzip config
+    gzip on;
```
membphis, line 4: why do we need to enable `gzip` by default? I think `off` is better for us.

only CDN need to open `gzip` by default.

juzhiyuan: It's ok to use gzip here

https://github.com/ant-design/ant-design-pro/blob/master/docker/nginx.conf#L4

compose/dashboard_conf/nginx.conf
```diff
@@ -0,0 +1,21 @@
...
+
+    root /usr/share/nginx/html;
+    include /etc/nginx/mime.types;
+
+    location / {
+        try_files $uri $uri/ /index.html;
+    }
+
+    location /apisix/admin {
+        proxy_pass http://manager:8080/apisix/admin;
```
juzhiyuan, line 19: if we need X-Forwarded-Proto here?

compose/dashboard_conf/nginx.conf
gxthrj: done

compose/dashboard_conf/nginx.conf
gxthrj: use ant-design default configuration temporarily

compose/dashboard_conf/nginx.conf
gxthrj: I think `http` is enough  for intranet

PR #432, opened 2020-08-31
Feat: manage api  support route group

liuxiran:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
fix issue: #431 
___
### Bugfix
- Description

- How to fix?

___

- add a new module routeGroup: 

  -  create table `route_group`

  - support CRUD and list route group

- create/update route support add attribute route group id

### DIFF
--- a/api/route/route.go
+++ b/api/route/route.go
@@ -101,11 +101,15 @@ func listRoute(c *gin.Context) {
 		db = db.Where("upstream_nodes like ? ", "%"+ip+"%")
 		isSearch = false
 	}
+	if rgid, exist := c.GetQuery("route_group_id"); exist {
+		db = db.Where("route_group_id equal ?", rgid)
+		isSearch = false
+	}
 	// search
 	if isSearch {
 		if search, exist := c.GetQuery("search"); exist {
 			s := "%" + search + "%"
-			db = db.Where("name like ? or description like ? or hosts like ? or uris like ? or upstream_nodes like ? ", s, s, s, s, s)
+			db = db.Where("name like ? or description like ? or hosts like ? or uris like ? or upstream_nodes like ? or route_group_id = ?", s, s, s, s, s, search)
 		}
 	}
 	// mysql
@@ -288,6 +292,7 @@ func findRoute(c *gin.Context) {
 			}
 			result.Script = script
 
+			result.RouteGroupId = route.RouteGroupId
 			resp, _ := json.Marshal(result)
 			c.Data(http.StatusOK, service.ContentType, resp)
 		}
--- a/api/route/route.go
+++ b/api/route/route.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/apisix/manager-api/conf"
 	"github.com/apisix/manager-api/errno"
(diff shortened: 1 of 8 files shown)


api/route/route.go
```diff
@@ -101,11 +101,15 @@ func listRoute(c *gin.Context) {
 		db = db.Where("upstream_nodes like ? ", "%"+ip+"%")
 		isSearch = false
 	}
+	if rgid, exist := c.GetQuery("route_group_id"); exist {
+		db = db.Where("route_group_id equal ?", rgid)
```
nic-chen, line 105: does `equal` works  ?

api/route/route_group.go
```diff
@@ -0,0 +1,195 @@
...
+		e := errno.FromMessage(errno.RouteGroupRequestError, err.Error())
+		logger.Error(e.Msg)
+		c.AbortWithStatusJSON(http.StatusInternalServerError, e.Response())
+		return
+	} else {
+		result := &service.ListResponse{Count: count, Data: routeGroupList}
+		resp, _ := json.Marshal(result)
+		c.Data(http.StatusOK, service.ContentType, resp)
+	}
+}
```
nic-chen, line 105: need a blank line between two functions.

api/route/route_group.go
```diff
@@ -0,0 +1,195 @@
...
+func listRouteGroupName(c *gin.Context) {
+	db := conf.DB()
```
nic-chen, line 107: i think we should make paging for this api.

api/service/route_group.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func (rd *RouteGroupDao) UpdateRouteGroup() error {
+	db := conf.DB()
+	return db.Model(&RouteGroupDao{}).Update(rd).Error
+}
+
+func (rd *RouteGroupDao) DeleteRouteGroup() error {
+	return conf.DB().Delete(&rd).Error
```
nic-chen, line 73: we should check if there is a route bound before delete

nic-chen (changes requested):

api/service/route_group.go
liuxiran: yeap，and it has been checked in `function deleteRouteGroup` in `api/route/route_group.go ` file

liuxiran: this check has been put in the RoutegroupDao before, and tansfered to the service after updated :)

api/route/route_group.go
liuxiran: this is used in select routegroup name in selectbox, and I found that the same method in upstream has not made paging too. so would it need to modified at the same time? @nic-chen

nic-chen: yes if you have time to do that :)

juzhiyuan:
@nic-chen any update?

api/route/route_group.go
liuxiran: ok, I will open other pr to make paging about  these two apis :)

api/route/route_group.go
liuxiran: done

api/route/route.go
liuxiran: done

api/route/route.go
```diff
@@ -209,6 +223,15 @@ func updateRoute(c *gin.Context) {
 			}
 		}()
 		logger.Info(rd)
+		if isCreateGroup {
+			if err := tx.Model(&service.RouteGroupDao{}).Create(routeGroup).Error; err != nil {
+				tx.Rollback()
```
nic-chen, line 228: why we need a transaction here ?

juzhiyuan: @liuxiran Please take look :D

liuxiran: Refer to the original design, route must be in a group, if the group has  not exited, create with the route. so here if the group create with error, do rollback :) @nic-chen

nic-chen: @liuxiran  Sorry, it looks like a separate transaction.

nic-chen (approved):

PR #485, opened 2020-09-17
feat: Route debug

### LINKED ISSUE #469: Feature: debug a route online

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirements
- [x] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

___
### Bug

- What happened?

___
### Requirement or improvement

**Online debugging would be convenient for users to test the route：**

1. support authentication

1. support to edit request params

1. support to show the response info

1. friendly UI  (swagger UI maybe a good choice,  Investigating～～)

liuxiran:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
fix #469 
___
### Bugfix
- Description

- How to fix?

___

### DIFF
--- a/api/route/route.go
+++ b/api/route/route.go
@@ -39,7 +39,7 @@ func AppendRoute(r *gin.Engine) *gin.Engine {
 	r.DELETE("/apisix/admin/routes/:rid", deleteRoute)
 	r.GET("/apisix/admin/notexist/routes", isRouteExist)
 	r.PUT("/apisix/admin/routes/:rid/offline", offlineRoute)
-	r.GET("/apisix/admin/routeinfos/:rid", getRouteAndApisixUrl)
+	r.GET("/apisix/admin/routes/:rid/debuginfo", getRouteWithApisixUrl)
 	return r
 }
 
@@ -482,7 +482,7 @@ func findRoute(c *gin.Context) {
 	}
 }
 
-func getRouteAndApisixUrl(c *gin.Context) {
+func getRouteWithApisixUrl(c *gin.Context) {
 	rid := c.Param("rid")
 	var count int
 	if err := conf.DB().Table("routes").Where("id=?", rid).Count(&count).Error; err != nil {
--- a/api/route/route.go
+++ b/api/route/route.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"regexp"
 
 	"github.com/apisix/manager-api/conf"
 	"github.com/apisix/manager-api/errno"
@@ -38,6 +39,7 @@ func AppendRoute(r *gin.Engine) *gin.Engine {
 	r.DELETE("/apisix/admin/routes/:rid", deleteRoute)
 	r.GET("/apisix/admin/notexist/routes", isRouteExist)
 	r.PUT("/apisix/admin/routes/:rid/offline", offlineRoute)
+	r.GET("/apisix/admin/routeinfos/:rid", getRouteAndApisixUrl)
 	return r
 }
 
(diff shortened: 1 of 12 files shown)


src/pages/Route/transform.ts
```diff
@@ -252,3 +252,131 @@ export const transformRouteData = (data: RouteModule.Body) => {
...
+    in: 'path',
+    description: `enter your path param`,
+    required: true,
+    type: 'string',
+  };
+  const requestBodyMethod = ['POST', 'PUT', 'PATCH'];
+
+  protocols.forEach((protocol) => {
+    if (protocol !== 'websocket') {
+      servers.push({
```
juzhiyuan, line 329: `push` will take some side effects, we would recommend using `concat` instead of `push`

src/pages/Route/transform.ts
```diff
@@ -252,3 +252,131 @@ export const transformRouteData = (data: RouteModule.Body) => {
...
+        url: `${protocol}://${url}`,
+      });
+      // FIXME
```
juzhiyuan, line 332: what happened here?

src/pages/Route/transform.ts
```diff
@@ -252,3 +252,131 @@ export const transformRouteData = (data: RouteModule.Body) => {
...
+  const tags: RouteModule.TagSchema[] = [
+    {
+      name: `Route-${name}`,
+      description: desc,
+    },
+  ];
+  const servers: RouteModule.Server[] = [];
+  const responses: RouteModule.ResponseSchema = {
+    // default response code
+    '200': {
```
juzhiyuan, line 279: could we use `ok` here? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200

src/pages/Route/typing.d.ts
```diff
@@ -157,5 +157,70 @@ declare namespace RouteModule {
       [name: string]: any;
     };
     script: Record<string, any>;
+    url?: string;
+  };
+
+  // for route debug
+  interface Server {
```
juzhiyuan, line 164: please use `type` instead of `interface` to prevent from type override :D

api/route/route.go
```diff
@@ -480,6 +482,58 @@ func findRoute(c *gin.Context) {
 	}
 }
 
+func getRouteAndApisixUrl(c *gin.Context) {
```
juzhiyuan, line 485: could we use camelCase here? e.g `getRouteWithAPISIXURL` .

juzhiyuan (approved):
Looks good to me anyway :D

src/pages/Route/transform.ts
liuxiran: This is a place I want to discuss with you:

In the docker environment， the apisix's url returned by the manager-api is the docker ip: 192.17.5.11, which is unavailable here， so I added a url got from `window. location` as a remedy. It's easy but  not flexible🤦‍♀️，
looking forward to your proposal @juzhiyuan

juzhiyuan: I'm not family with Docker, this issue could be discussed with @nic-chen @gxthrj @bzp2010 , if this trick way works, I think it's ok to use here.

moonming (approved):

src/pages/Route/transform.ts
liuxiran: got it 👍,thank you, and both of concat and spread syntax  have the same effect here , I perfer the latter :)

src/pages/Route/transform.ts
liuxiran: done

src/pages/Route/typing.d.ts
liuxiran: done

juzhiyuan (approved):
LGTM

api/route/route.go
liuxiran: Refer to the naming of Apisix in other documents, updated to getRouteWithApisixUrl :)

PR #486, opened 2020-09-19
feat: add consumer CURD refactoring

nic-chen:

add consumer CURD refactoring

### DIFF
--- a/api/internal/handler/consumer/consumer.go
+++ b/api/internal/handler/consumer/consumer.go
@@ -1,4 +1,4 @@
-package route
+package consumer
 
 import (
 	"reflect"
@@ -17,16 +17,16 @@ import (
 )
 
 type Handler struct {
-	routeStore *store.GenericStore
+	consumerStore *store.GenericStore
 }
 
 func NewHandler() (handler.RouteRegister, error) {
 	s, err := store.NewGenericStore(store.GenericStoreOption{
-		BasePath: "/apisix/routes",
-		ObjType:  reflect.TypeOf(entity.Route{}),
+		BasePath: "/apisix/consumers",
+		ObjType:  reflect.TypeOf(entity.Consumer{}),
 		KeyFunc: func(obj interface{}) string {
-			r := obj.(*entity.Route)
-			return r.ID
+			r := obj.(*entity.Consumer)
+			return r.Username
 		},
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


api/internal/handler/consumer/consumer.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/shiningrush/droplet"
+	"github.com/shiningrush/droplet/data"
+	"github.com/shiningrush/droplet/wrapper"
+	wgin "github.com/shiningrush/droplet/wrapper/gin"
```
juzhiyuan, line 11: wgin?

api/internal/handler/consumer/consumer.go
```diff
@@ -0,0 +1,135 @@
...
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.GET("/apisix/admin/consumers/:id", wgin.Wraps(h.Get,
+		wrapper.InputType(reflect.TypeOf(GetInput{}))))
+	r.GET("/apisix/admin/consumers", wgin.Wraps(h.List,
+		wrapper.InputType(reflect.TypeOf(ListInput{}))))
+	r.POST("/apisix/admin/consumers", wgin.Wraps(h.Create,
+		wrapper.InputType(reflect.TypeOf(entity.Consumer{}))))
+	r.PUT("/apisix/admin/consumers/:id", wgin.Wraps(h.Update,
+		wrapper.InputType(reflect.TypeOf(UpdateInput{}))))
+	r.DELETE("/apisix/admin/consumers", wgin.Wraps(h.BatchDelete,
```
juzhiyuan, line 54: BatchDelete means it will delete all consumers? when we want to delete only 1 item, we should pass only 1 id, right?

api/route/base.go
```diff
@@ -37,19 +38,20 @@ func SetUpRouter() *gin.Engine {
 	r := gin.New()
 	store := cookie.NewStore([]byte("secret"))
 	r.Use(sessions.Sessions("session", store))
-	r.Use(filter.CORS(), filter.Authentication(), filter.RequestId(), filter.RecoverHandler())
+	r.Use(filter.CORS(), filter.Authentication(), filter.RequestId(), filter.RecoverHandler()) //
```
juzhiyuan, line 41: `//`?

api/route/base.go
```diff
@@ -37,19 +38,20 @@ func SetUpRouter() *gin.Engine {
...
 
 	AppendHealthCheck(r)
 	AppendAuthentication(r)
 	//AppendRoute(r)
 	AppendSsl(r)
 	AppendPlugin(r)
 	AppendUpstream(r)
-	AppendConsumer(r)
+	//AppendConsumer(r)
```
juzhiyuan, line 49: just removed the extra comment line

juzhiyuan (changes requested):

api/internal/handler/consumer/consumer.go
ShiningRush: Yes

api/internal/handler/consumer/consumer.go
ShiningRush: this is  a gin wrapper for the droplet

api/route/base.go
ShiningRush: No, just keep this until all API migrating completed.
This can help us compare the previous code

api/internal/handler/consumer/consumer.go
```diff
@@ -0,0 +1,135 @@
...
+	consumerStore *store.GenericStore
+}
+
+func NewHandler() (handler.RouteRegister, error) {
+	s, err := store.NewGenericStore(store.GenericStoreOption{
+		BasePath: "/apisix/consumers",
+		ObjType:  reflect.TypeOf(entity.Consumer{}),
+		KeyFunc: func(obj interface{}) string {
+			r := obj.(*entity.Consumer)
+			return r.Username
```
ShiningRush, line 29: I think we should use "ID" as key, keep same as admin.I will add a unique check function for generic store,so you can use it to check username's unique.

ShiningRush (changes requested):

api/route/base.go
nic-chen: removed. @juzhiyuan

api/internal/handler/consumer/consumer.go
nic-chen: @ShiningRush 
username is used as key for consumer not ID.
see:
https://github.com/apache/apisix/blob/master/apisix/admin/consumers.lua#L77

juzhiyuan (approved):

ShiningRush (approved):

PR #489, opened 2020-09-19
feat: Upstream health check

### LINKED ISSUE #470: Proposal: Health Checks for Upstream
image

## Why
Doc: https://github.com/apache/apisix/blob/master/doc/health-check.md

## How
Active and passive checks exist in parallel in the upstream module as separate configurations, they should obey the following rules.
1. All disabled: Only show the Ant Design Switcher components for those two fields, no more configuration form；
2. Enable active check: show its configuration form；
3. Enable passive check
	1. Check if the active check is enabled, if not, just show a warning notification by using Ant Design’s Notification component, and tell users to enable the active check first.
	2. If the active check is enabled, then users could enable the passive check and show its form.

## API
1. Test cases: https://github.com/apache/apisix/blob/master/t/admin/health-check.t#L61

## Reference
1. Logic: https://github.com/apache/apisix/blob/master/apisix/admin/upstreams.lua

liuxiran:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
fix #470 
___
### Bugfix
- Description

- How to fix?

___


juzhiyuan:
waiting for https://github.com/apache/apisix-dashboard/issues/470#issuecomment-695333690

src/pages/Upstream/components/Step1.tsx
```diff
@@ -43,9 +47,38 @@ const initialValues = {
...
+      },
+      unhealthy: {
+        interval: 1,
+        http_failures: 2,
+      },
+      req_headers: [''],
+    },
+    passive: {
+      healthy: {
+        http_statuses: [undefined],
```
juzhiyuan, line 69: could we use an empty array here?

src/pages/Upstream/components/Step1.tsx
```diff
@@ -171,6 +204,390 @@ const Step1: React.FC<Props> = ({ form, disabled }) => {
 
   const renderTimeUnit = () => <span style={{ margin: '0 8px' }}>ms</span>;
 
+  function activeChange() {
```
juzhiyuan, line 207: We'd better use `const fn = () => {}` to have a unified style :D

juzhiyuan: Also, for onChange handler, I would prefer using some name like `handleXXX`

src/pages/Upstream/components/Step1.tsx
```diff
@@ -171,6 +204,390 @@ const Step1: React.FC<Props> = ({ form, disabled }) => {
...
+    if (isActive) {
+      onChange(!isActive, false);
+      form.setFieldsValue({ ...form.getFieldsValue(), passive: false });
+      return;
+    }
+    onChange(!isActive, isPassive);
+    form.setFieldsValue({ ...form.getFieldsValue(), active: !isActive });
+  }
+  function passiveChange() {
```
juzhiyuan, line 216: same here

juzhiyuan (approved):
LGTM from codes style

src/pages/Upstream/components/Step1.tsx
liuxiran: empty array will lose  the first item  while rendering Form.List 
we tried to use [''], but the type of the http_statuses: number[], so we use [undefined] here

src/pages/Upstream/components/Step1.tsx
liuxiran: done

src/pages/Upstream/components/Step1.tsx
liuxiran: done

src/pages/Upstream/components/Step1.tsx
juzhiyuan: got it, thanks!

juzhiyuan (approved):

LiteSun (approved):

Issue #509, opened 2020-09-24
create a consumer only enable openid-connect plugin return a warning blocks the next step


liuxiran:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

I tried to create a consumer only enabled `openid-connect` plugin, and press [nextStep], then got a warning: `Please enable at least one authentication plugin`. 
![image](https://user-images.githubusercontent.com/2561857/94093591-2502ca00-fe50-11ea-8272-800659961753.png)

and I checked the code: https://github.com/apache/apisix-dashboard/blob/69199d22dd30e47a71c191060a0402d43633addf/src/pages/Consumer/Create.tsx#L71

`openId-connect` does not match the rule indeed, **so change the plugin-name or change the match rule or just keep the degin？**

___
### Bug
- Which version of Apache APISIX Dashboard, OS, and Browser?

- What happened?
If possible, provide a way to reproduce the error.

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

juzhiyuan:
yes, could you help modify this checker? just like a whitelist?

liuxiran:
> yes, could you help modify this checker? just like a whitelist?

OK

PR #512, opened 2020-09-24
fix: update valify rules about whether enabled auth plugin in create/…

liuxiran:
…update a consumer

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fix: #509 
___
### Bugfix
- Description

- How to fix?
update valify rules to create/update a consumer must enable one of the five auth plugins.
(when enables two or above auth plugins , it will return a error from Apisix: `only one auth plugin is allowed`
2020-09-24 18-27-50屏幕截图
2020-09-24 18-27-11屏幕截图

) 

___

### DIFF
--- a/src/pages/Consumer/Create.tsx
+++ b/src/pages/Consumer/Create.tsx
@@ -18,7 +18,7 @@ import React, { useState, useEffect } from 'react';
 import { PageContainer } from '@ant-design/pro-layout';
 import { Card, Steps, notification, Form } from 'antd';
 import { history, useIntl } from 'umi';
-import { PluginPage, PluginPageType } from '@api7-dashboard/plugin';
+import { PluginPage, PluginPageType, PLUGIN_MAPPER_SOURCE } from '@api7-dashboard/plugin';
 
 import ActionBar from '@/components/ActionBar';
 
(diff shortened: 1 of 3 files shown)


src/pages/Consumer/Create.tsx
```diff
@@ -68,8 +68,18 @@ const Page: React.FC = (props) => {
...
+        'basic-auth',
+        'jwt-auth',
+        'key-auth',
+        'authz-keycloak',
+      ];
+      const currentAuthPlugin = Object.keys(plugins).filter((plugin) =>
+        authPluginNames.includes(plugin),
+      );
+      const currentAuthPluginLen = currentAuthPlugin.length;
+      if (currentAuthPluginLen > 1 || currentAuthPluginLen === 0) {
```
juzhiyuan, line 82: Maybe there have some mistakes here? I think this message will only popup when there has no 1 auth plugin :D

src/pages/Consumer/Create.tsx
```diff
@@ -68,8 +68,18 @@ const Page: React.FC = (props) => {
         setStep(nextStep);
       });
     } else if (nextStep === 3) {
-      const isValid = Object.keys(plugins).some((name) => name.includes('auth'));
-      if (!isValid) {
+      const authPluginNames = [
```
juzhiyuan, line 71: oh my mistake, you could import an plugin array from https://github.com/api7/dashboard-components/blob/master/packages/plugin/src/data.tsx#L4

juzhiyuan:
```js
import {PLUGIN_MAPPER_SOURCE} from '@api7-dashboard/plugin;
```

https://github.com/api7/dashboard-components/blob/master/packages/plugin/src/data.tsx#L37

src/pages/Consumer/Create.tsx
juzhiyuan: let's invite @membphis @moonming to check this condition :D

juzhiyuan (changes requested):
Please use the `PLUGIN_MAPPER_SOURCE` from `plugin` module :D (my mistaken 😂)

juzhiyuan:
I noticed the response from that snapshot, yes, we only allow to enable 1 plugin.

src/pages/Consumer/Create.tsx
liuxiran: got it~thanks, already done 😊

juzhiyuan (approved):
You got it!

LiteSun (approved):

PR #561, opened 2020-10-16
feat: compatible with PUT method of `admin api` and nodes of upstream

nic-chen:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___

compatible with PUT method of `admin api` and nodes of upstream

### DIFF
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -38,7 +38,7 @@ type Interface interface {
 	Get(key string) (interface{}, error)
 	List(input ListInput) (*ListOutput, error)
 	Create(ctx context.Context, obj interface{}) error
-	Update(ctx context.Context, obj interface{}) error
+	Update(ctx context.Context, obj interface{}, createOnFail bool) error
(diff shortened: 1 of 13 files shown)


api/internal/core/store/store.go
```diff
@@ -267,6 +274,9 @@ func (s *GenericStore) Update(ctx context.Context, obj interface{}) error {
 	}
 	oldObj, ok := s.cache.Load(key)
 	if !ok {
+		if createOnFail {
+			return s.Create(ctx, obj)
```
ShiningRush, line 278: We should init `create_time`、`update_time`  and `id`，Hmmm，we can add two function for baseinfo: `creating` and `updating` to do some logic about `BaseInfo`, then add `BaseInfoSetter` interface to use it.

ShiningRush (changes requested):

ShiningRush (approved):

juzhiyuan (approved):
Not sure if logics are ok, but looks good to me from codes format.

PR #563, opened 2020-10-19
feat: compatible with HTTP status of `admin api`

nic-chen:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___


juzhiyuan (approved):
Not sure if logics are ok, but looks good to me from codes style.

moonming (changes requested):
No test cases

api/internal/core/store/validate.go
```diff
@@ -125,17 +125,17 @@ func (v *APISIXJsonSchemaValidator) Validate(obj interface{}) error {
...
 			s, err := gojsonschema.NewSchema(gojsonschema.NewStringLoader(schemaDef))
 			if err != nil {
-				return fmt.Errorf("new schema failed: %w", err)
+				return fmt.Errorf("scheme validate failed: %w", err)
 			}
 
 			ret, err := s.Validate(gojsonschema.NewGoLoader(pluginConf))
 			if err != nil {
-				return fmt.Errorf("validate failed: %w", err)
+				return fmt.Errorf("scheme validate failed: %w", err)
```
liuxiran, line 138: I found that the same error messages repeatedly appeared in the files, would it be better to  manage error messages in one place just like [error.go](https://github.com/apache/apisix-dashboard/blob/master/api/errno/error.go) before :)?

membphis:
> No test cases

https://github.com/apache/apisix-dashboard/pull/563/commits/580b88e10edd82d4c02e3d583ce56aff4123f1d1

@moonming you missed them

api/internal/core/store/validate.go
ShiningRush: Agree, but we can not do it in this version for now. When the first rc version is released, we will optimize it in subsequent versions

ShiningRush (approved):

api/internal/core/store/validate.go
liuxiran: +1 @ShiningRush

liuxiran (approved):

juzhiyuan:
Let's merge it now

Issue #542, opened 2020-10-09
feature: [refactor] embedding static files into Go binaries


nic-chen:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirements
- [x] Feature or performance improvement
- [ ] Other

___
### Requirement or improvement

We need to embed static files into Go binaries, 

so that we don't need to deploy an additional nginx ,

and it will be very easy to use.

juzhiyuan:
@bzp2010 is working on this issue, please let @nic-chen know whenever there have any blocker issues.

bzp2010:
Work branch: [Click Here](https://github.com/bzp2010/apisix-dashboard/pull/1)
Github Workflow: [Click Here](https://github.com/bzp2010/apisix-dashboard/blob/master/.github/workflows/api_newci.yml)

## feature
- [x] Packing (It worked on my own computer)
> It's use [rakyll/statik](https://github.com/rakyll/statik) library.
- [ ] build by Github Actions

## problem 
- The binary program provided by `rakyll/statik` is used for packaging, but after executing the go get instruction in GitHub Actions, the program is not in the path directory and needs to be processed.

juzhiyuan:
cc @nic-chen @gxthrj Could anyone help @bzp2010 to resolve this problem?

PR #588, opened 2020-10-22
feat: embedding static files into Go binaries

bzp2010:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#542 

Provides the ability to package static files into Go binaries.

### DIFF
--- a/api/main.go
+++ b/api/main.go
@@ -14,6 +14,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+//go:generate statik -src=./dashboard -f
+
 package main
 
 import (
--- a/api/go.mod
+++ b/api/go.mod
@@ -26,4 +26,5 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	go.etcd.io/etcd v3.3.25+incompatible
 	go.uber.org/zap v1.16.0
+	github.com/rakyll/statik v0.1.7
 )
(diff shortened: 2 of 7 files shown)


api/main.go
```diff
@@ -14,6 +14,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+//go:generate statik -src=./dashboard -f
```
membphis, line 18: It means it is optional, right?

bzp2010: It describes the instructions that will be executed when you run `go generate` directly. Users can use the static command instead of go generate.
This command does not affect the original compilation process.

juzhiyuan:
@bzp2010 Could we keep both https://github.com/apache/apisix-dashboard/pull/594 and https://github.com/apache/apisix-dashboard/pull/588?

Two ways to deploy the dashboard:
- Go with static files
- Go and static files

bzp2010:
> @bzp2010 Could we keep both #594 and #588?
> 
> Two ways to deploy the dashboard:
> 
> * Go with static files
> * Go and static files

Perhaps we can use two compile scripts to implement them separately, or use configuration files to select built-in static files or external folders at Golang runtime.

juzhiyuan:
ok then @ShiningRush what do you think?

ShiningRush:
this pr is not production ready for those reason:
- static file path is `/dashboard`
- it do not contain gin's `no route action` to handle `History Mode`
- `statik` may be not the best choice for us, we can have a talk before we implement this(more detail plz refer to #594 description).

It is better to embedding files after #594 merged, we can implement this feature base on it.

juzhiyuan:
Just to make sure, @bzp2010 Is this PR still in development?

juzhiyuan:
I have to close this PR for there may have many conflicts with v2.0.

PR #666, opened 2020-11-02
fix: promethus incorrect value when update route

### LINKED ISSUE #627: Route update prompts: scheme validate failed: (root): Additional property enabled is not allowed

- [ ] Question or discussion
- [x] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

___
### Bug

Bug with routing updates. When the prometheus plugin is enabled, update route request prompts: `scheme validate failed: (root): Additional property enabled is not allowed`

路由更新时的 bug，当 prometheus 插件开启的时候，更新路由时会提示：`scheme validate failed: (root): Additional property enabled is not allowed`

- What happened?

___
### Requirements or improvement

LiteSun:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bugfix
- Description
 Put incorrect Prometheus value when update route. 
 resolved #627 

- How to fix?

___

### DIFF
--- a/frontend/src/pages/Route/transform.ts
+++ b/frontend/src/pages/Route/transform.ts
@@ -152,10 +152,6 @@ export const transformRouteData = (data: RouteModule.Body) => {
 
   const { plugins, script } = data;
 
-  if (plugins && plugins.prometheus) {
-    plugins.prometheus = { enabled: true };
-  }
-
   const step3Data: RouteModule.Step3Data = {
     plugins,
     script,


frontend/src/pages/Route/transform.ts
```diff
@@ -152,10 +152,6 @@ export const transformRouteData = (data: RouteModule.Body) => {
 
   const { plugins, script } = data;
 
-  if (plugins && plugins.prometheus) {
-    plugins.prometheus = { enabled: true };
```
juzhiyuan, line 156: If we remove this line, and just enabled this plugin, then how about this plugin's status when we use Edit mode? Is it still enabled?

LiteSun: > If we remove this line, and just enabled this plugin, then how about this plugin's status when we use Edit mode? Is it still enabled?

When we in Edit mode, API will return `plugins:{promethues:{}}`if we enabled it before and it works as expected.

image

juzhiyuan (approved):

PR #711, opened 2020-11-06
fix: deploy doc

nic-chen:
remove useless description
fix config description.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -117,7 +117,7 @@ $ api/run.sh &
 
 2. `conf.etcd.endpoints` is used to set ETCD's instances address, it supports multiple instances mode.
 
-```json
+```yaml
   etcd:
     endpoints:
       - 127.0.0.1:2379
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -81,7 +81,7 @@ The bundled files are located in the root directory `/output/html`.
 
 ## Run
 
-1. According to your deploy environment, check the related configurations in `api/conf/conf.yaml`, modify those variables if needed.
+1. According to your deploy environment, check the related configurations in `output/conf/conf.yaml`, modify those variables if needed.
 
 Example:
 
@@ -90,7 +90,6 @@ conf:
   listen:
     host: 127.0.0.1
     port: 8080
-  dag_lib_path: ''
   etcd:
     endpoints:
       - 127.0.0.1:2379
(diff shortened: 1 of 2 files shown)


docs/deploy.md
```diff
@@ -81,7 +81,7 @@ The bundled files are located in the root directory `/output/html`.
 
 ## Run
 
-1. According to your deploy environment, check the related configurations in `api/conf/conf.yaml`, modify those variables if needed.
+1. According to your deploy environment, check the related configurations in `output/conf/conf.yaml`, modify those variables if needed.
```
juzhiyuan, line 84: Why? The output folder is generated only when we run the shell.

docs/deploy.md
```diff
@@ -114,18 +113,15 @@ $ api/run.sh &
...
 ```json
-{
-  "etcd": {
-    "endpoints": "127.0.0.1:2379,127.0.0.1:3379"
-  }
-}
+  etcd:
+    endpoints:
+      - 127.0.0.1:2379
+      - 127.0.0.1:3379
```
juzhiyuan, line 124: image

juzhiyuan (changes requested):

docs/deploy.md
nic-chen: The binary program we run is in the `output` directory, so the config file it reads is `output/conf/conf.yaml`

juzhiyuan: But we don't havr that directory in source codes, and users can choose not to run the binary file in that folder.

juzhiyuan (commented):
Enmm

docs/deploy.md
nic-chen: Then he should change the config file in the directory he run. After  run `api/build.sh`, the directory of the binary file is `output`, changing `api/conf/conf.yaml` will not take effect.

nic-chen: or what's your suggestion ? @juzhiyuan

juzhiyuan: 1. Before building the binary file, we should update the config file under the `/api/conf/` directory, right?
2. After building successfully, we have the output directory.
3. When users want to update the config file without rebuilding the binary file, they must modify the file under the `/output` directory.

juzhiyuan: According to your opinion, we should add some tips in the default config file, to let users know they can update the config only when the binary is generated.

moonming, line 84: This is unacceptable, there can only be one configuration file

moonming (changes requested):
This is unacceptable, there can only be one configuration file

nic-chen:
@juzhiyuan @moonming 

Actually, users only need to care about `/output/conf/`, only developers need to care about `/api/conf/`

juzhiyuan:
> @juzhiyuan @moonming 
> 
> Actually, users only need to care about `/output/conf/`, only developers need to care about `/api/conf/`
> 
> 

But users don't know

membphis:
> @juzhiyuan @moonming
> 
> Actually, users only need to care about `/output/conf/`, only developers need to care about `/api/conf/`

that is a bad way. 

after build, we can generate `dag-to-lua` and `manager-api` in `/api`.

```
$ tree ./ -L 1
./
├── conf
├── dag-to-lua
├── Dockerfile
├── filter
├── go.mod
├── go.sum
├── internal
├── log
├── logs
├── manager-api
├── README.md
└── test
```

nic-chen:
> > @juzhiyuan @moonming
> > Actually, users only need to care about `/output/conf/`, only developers need to care about `/api/conf/`
> 
> that is a bad way.
> 
> after build, we can generate `dag-to-lua` and `manager-api` in `/api`.
> 
> ```
> $ tree ./ -L 1
> ./
> ├── conf
> ├── dag-to-lua
> ├── Dockerfile
> ├── filter
> ├── go.mod
> ├── go.sum
> ├── internal
> ├── log
> ├── logs
> ├── manager-api
> ├── README.md
> └── test
> ```


We have other considerations, that is, users can directly copy the `output` directory to any location they want.

juzhiyuan:
@membphis Yes, we couldn't build it under the api directory due to users may copy them and move them to another place.

juzhiyuan:
Why not use a template-configuration? No matter use a standalone file or just added them to codes.

membphis:
I think we need refractory the deploy doc.

Let us discuss in issue first.

gxthrj:
I create a new issue to impove #715

nic-chen:
so this pr should be closed now.

PR #719, opened 2020-11-07
bugfix: dashboard 2.0 failed to fetch ssl certificate not found 

### LINKED ISSUE #697: dashboard 2.0 failed to fetch ssl certificate not found

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [ ] Other

___
### Question
- What do you want to know?

___
### Bug

dashboard 2.0 
centos 7
chrome 83

- What happened?
证书已经上传了，但是错误日志中报下面这个截图
image

___
### Requirements or improvement

nic-chen:
close #697


api/test/e2e/ssl_test.go
```diff
@@ -0,0 +1,115 @@
...
+			caseDesc:     "create ssl fail - key and cert not match",
+			Object:       MangerApiExpect(t),
+			Method:       http.MethodPost,
+			Path:         "/apisix/admin/ssl",
+			Body:         string(invalidBody),
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusBadRequest,
+		},
+		{
+			caseDesc:     "create ssl successful",
```
juzhiyuan, line 57: `successfully`, *adv*

api/internal/handler/ssl/ssl.go
```diff
@@ -147,6 +147,7 @@ func (h *Handler) Create(c droplet.Context) (interface{}, error) {
 	}
 
 	ssl.ID = input.ID
+	ssl.Status = 1
```
moonming, line 150: Is this the cause of the bug?  Can you explain the reason?

.github/workflows/e2e-test-backend.yml
```diff
@@ -28,7 +28,10 @@ jobs:
           sleep 5
           docker logs docker_managerapi_1
 
+      - name: set hosts
+        working-directory: ./api/test/e2e
+        run: echo "127.0.0.1 www.test2.com" | sudo tee -a /etc/hosts
```
moonming, line 33: why need to change local file?

api/test/e2e/ssl_test.go
```diff
@@ -0,0 +1,122 @@
...
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusOK,
+		},
+		{
+			caseDesc:     "delete ssl",
+			Object:       MangerApiExpect(t),
+			Method:       http.MethodDelete,
+			Path:         "/apisix/admin/ssl/1",
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusOK,
```
moonming, line 115: We need to check again after delete route and ssl cert

moonming (changes requested):

api/internal/handler/ssl/ssl.go
nic-chen: yes, it's the cause.
That is because in Go if does not explicitly assign a value, it will automatically assign a 0 value. So the SSL is disable.

.github/workflows/e2e-test-backend.yml
nic-chen: That's because we can't make https request like this:

```
curl https://127.0.0.1:9443  -H 'Host: www.test2.com'
```

the SSL of www.test2.com will not match by this way. so need to set host and request www.test2.com directly.

api/test/e2e/ssl_test.go
nic-chen: ok

api/internal/handler/ssl/ssl.go
membphis: that is the magic number. we should use a `const` variable for a better name.

.github/workflows/e2e-test-backend.yml
membphis: `curl --resolve 'www.test2.com:9443:127.0.0.1'  https://www.test2.com:9443`

you can make a try in this way.  ^_^

api/internal/handler/ssl/ssl.go
membphis: and I think we'd better add some comment here, it should be useful for code reader

nic-chen: OK.

.github/workflows/e2e-test-backend.yml
nic-chen: thanks. I will try for it. but it's an example, we are not use curl to test.

membphis: ok, got it. we can write some comment first, then we can fix it in a better way.

nic-chen: It works!

api/internal/handler/ssl/ssl.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/shiningrush/droplet/wrapper"
 	wgin "github.com/shiningrush/droplet/wrapper/gin"
 
+	"github.com/apisix/manager-api/conf"
```
membphis, line 37: we'll move this module to another place, is it right?

api/internal/handler/ssl/ssl.go
```diff
@@ -147,6 +148,8 @@ func (h *Handler) Create(c droplet.Context) (interface{}, error) {
 	}
 
 	ssl.ID = input.ID
+	//set default value for SSL status, if not set, it will be 0 which means disable.
+	ssl.Status = conf.SSLDefaultStatus
 	if err := h.sslStore.Create(c.Context(), ssl); err != nil {
 		return handler.SpecCodeResponse(err), err
```
membphis, line 154: Did we catch this error and write it to the error log?  please confirm it. if not, we can create a new issue to fix it

api/internal/handler/ssl/ssl.go
```diff
@@ -169,6 +172,9 @@ func (h *Handler) Update(c droplet.Context) (interface{}, error) {
 	if input.ID != "" {
 		ssl.ID = input.ID
 	}
+
+	//set default value for SSL status, if not set, it will be 0 which means disable.
+	ssl.Status = conf.SSLDefaultStatus
 	if err := h.sslStore.Update(c.Context(), ssl, true); err != nil {
 		return handler.SpecCodeResponse(err), err
```
membphis, line 179: ditto

api/test/e2e/ssl_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for _, tc := range tests {
+		testCaseCheck(tc)
+	}
+
+	//try again after deleting SSL, make a HTTPS request
+	time.Sleep(time.Duration(20) * time.Millisecond)
+	_, err = http.Get("https://www.test2.com:9443")
+	assert.NotNil(t, err)
+	assert.EqualError(t, err, "Get https://www.test2.com:9443: remote error: tls: internal error")
```
moonming, line 137: why return ssl error? The router has been deleted

api/internal/handler/ssl/ssl.go
nic-chen: yes. when we adjust the code dir.

api/internal/handler/ssl/ssl.go
nic-chen: sure, I will create an issue.

api/test/e2e/ssl_test.go
nic-chen: because APISIX exit when no SSL match:
https://github.com/apache/apisix/blob/master/apisix/init.lua#L174-L180

moonming: Then why you need to delete router?https://github.com/apache/apisix-dashboard/pull/719/files#diff-c2b7edfd95881ffeb7a02870b56d8c228b7b1f0f83d78be56f4462091bf76dc7R112-R117
 just delete ssl is enough

nic-chen: just clean test data.

moonming: Please move delete router to the end, and add comments

nic-chen: updated.

api/conf/conf.go
```diff
@@ -40,16 +40,17 @@ const (
...
+	Schema           gjson.Result
+	WorkDir          = "."
+	ServerHost       = "127.0.0.1"
+	ServerPort       = 80
+	ETCDEndpoints    = []string{"127.0.0.1:2379"}
+	ErrorLogLevel    = "warn"
+	ErrorLogPath     = "logs/error.log"
+	UserList         = make(map[string]User, 2)
+	AuthConf         Authentication
+	SSLDefaultStatus = 1 //enable ssl by default
```
membphis, line 53: `CONST` variable? avoid magic number

api/internal/handler/ssl/ssl.go
membphis: if you created, you should better leave an issue id here.

membphis (approved):
LGTM, except the minor issues that I have written.

gxthrj (approved):

moonming (approved):

api/conf/conf.go
membphis: @nic-chen please create a new issue about this point, we will fix this later

Issue #734, opened 2020-11-09
update_time from manager api shouldn't be 0


juzhiyuan:
This issue would be fall back on the frontend according to @membphis 's advice.

membphis:
if the `update_time` or `change_time` is `0`, we'll show `None` at dashboard.

tip: this is a temporary hack, we'll remove this process finally.

juzhiyuan:
what's change_time?

membphis:
https://github.com/apache/apisix/blob/master/apisix/schema_def.lua#L289-L290
![image](https://user-images.githubusercontent.com/6814606/98545055-d8743080-22cf-11eb-8d8b-7e1491094f62.png)

juzhiyuan:
hmm, you mean updated time?

membphis:
> what's change_time?

sorry, it should be `create_time`

PR #736, opened 2020-11-09
fix: invalid values from the manager-api

juzhiyuan:
resolve https://github.com/apache/apisix-dashboard/issues/734

### DIFF
--- a/web/src/helpers.tsx
+++ b/web/src/helpers.tsx
@@ -18,6 +18,7 @@ import React from 'react';
 import { notification } from 'antd';
 import { MenuDataItem } from '@ant-design/pro-layout';
 import { history } from 'umi';
+import moment from 'moment';
 
 import { codeMessage } from './constants';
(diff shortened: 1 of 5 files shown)


web/src/helpers.tsx
```diff
@@ -96,3 +97,12 @@ export const getUrlQuery: (key: string) => string | false = (key: string) => {
   }
   return false;
 };
+
+export const timestampToLocaleString = (timestamp: number) => {
+  if (!timestamp) {
+    // TODO: i18n
+    return "None"
```
LiteSun, line 104: I think it's wired to show text `None` in the time column and some of user maybe don't know what's it mean. How about return empty string ``? 

image

juzhiyuan: This is a issue from backend, please see the related issue.

LiteSun (approved):

PR #738, opened 2020-11-09
test: add e2e test cases for upstream

nic-chen:
add e2e test cases for upstream

and fix bugs found during testing


membphis (changes requested):
as we talked before, one PR for one thing, if you found new bug, we need to fix them in a new PR.

membphis:
we should only add E2E code in this PR

api/internal/handler/upstream/upstream.go
```diff
@@ -144,8 +146,33 @@ type BatchDelete struct {
 
 func (h *Handler) BatchDelete(c droplet.Context) (interface{}, error) {
 	input := c.Input().(*BatchDelete)
+	ids := strings.Split(input.IDs, ",")
 
-	if err := h.upstreamStore.BatchDelete(c.Context(), strings.Split(input.IDs, ",")); err != nil {
+	for _, id := range ids {
```
membphis: why do we need those codes?

they are not E2E test codes.

nic-chen: removed.

membphis (approved):
LGTM, good luck

membphis:
merged, many thx

PR #753, opened 2020-11-09
chore: move the Dockerfile that may be confusing

nic-chen:
move `api/Dockerfile` into `api/test/docker`


membphis:
This is a temporary solution, we will put it completely in `api/test/docker` later, right?

Please confirm.

nic-chen:
> This is a temporary solution, we will put it completely in `api/test/docker` later, right?
> 
> Please confirm.

updated.

membphis (approved):

Issue #787, opened 2020-11-11
makefile: for `make release-src`, it should release the source code


membphis:
https://github.com/apache/apisix-dashboard/blob/v2.0/Makefile#L93

that is wrong, for `make release-src`, it should release the source code. 

but we released binary file now, this is wrong.

```
.PHONY: release-src
release-src:
	tar –cf dashboard.tar ./output/*

membphis:
@gxthrj you can take a look at this PR

gxthrj:
ok, we can call it `make release`,the release package contains the binary file of `manager-api` and static files of `web`.

juzhiyuan:
make-release is more semantic and clearer.

gxthrj:
@membphis Can we close this issue now? 
The document has changed the original purpose of packaging the `output` directory.
`make release-src` is for source release now. #816

juzhiyuan:
We could close this one IMO.

PR #807, opened 2020-11-15
chore: rename makefile release-src to release

gxthrj:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,9 @@ build
 manager-api
 
 output
+release
 default.etcd
 api/build-tools/apisix
 /*.zip
+/*.tar
+/*.tgz
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,7 @@ build
 manager-api
 
 output
-release
+publish
 default.etcd
 api/build-tools/apisix
 /*.zip
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ UNAME ?= $(shell uname)
 YARN_EXEC ?= $(shell which yarn)
 GO_EXEC ?= $(shell which go)
 VERSION ?= latest
-RELEASE = apache-apisix-dashboard-${VERSION}
+PUBLISH = apache-apisix-dashboard-${VERSION}
 
 export GO111MODULE=on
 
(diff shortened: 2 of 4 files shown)


membphis (approved):

Makefile
```diff
@@ -89,7 +89,7 @@ endif
 	.actions/openwhisk-utilities/scancode/scanCode.py --config .actions/ASF-Release.cfg ./
 
 
-.PHONY: release-src
-release-src:
+.PHONY: release
```
juzhiyuan, line 92: Does release only do a package work? 

cc @moonming @ShiningRush

gxthrj: Do you want to include the build process?

moonming (changes requested):
please refer to apisix's make release

gxthrj:
Add sha512 and mv the pkgs to release directory .
Refer to apisix's make release.

Makefile
```diff
@@ -91,5 +93,10 @@ endif
 
 .PHONY: release
 release:
-	tar –cf dashboard.tar ./output/*
+	tar -zcvf $(RELEASE).tgz ./output/*
```
moonming, line 96: NO.
`release` is ONLY for release manager, who ONLY release source code packge.

moonming (changes requested):

Makefile
gxthrj: Ok, It seems that `make release` should not be used here.
The `output` directory is not source code. They are the result of compilation, including manager and web.
Maybe `make publish` is more suitable. Ref to some web projects , Such as elementUI.

.gitignore
```diff
@@ -42,6 +42,9 @@ build
 manager-api
 
 output
+publish
```
juzhiyuan, line 45: We use `publish` as a folder is not a good idea, can you provide some examples of other popular projects?

The act of ordering this line is to package the product, and we need to think about whether we need to do this here. 

If we need a package as the final product, then just packaged it in the build process.
If we use a package just to let users move it easily, let's call it `make package`, but I don't think it's needed and will take the confusion with `make build`.

cc @moonming @membphis any ideas?

juzhiyuan (changes requested):

.gitignore
moonming: agreed. @gxthrj please refer to other repos first.
To be honest, I don’t know what PR does

gxthrj: It is for the issue #787 
It seems that I didn't understood the meaning of issue.

gxthrj:
Close the PR,  #816 has been merged. 
It seems no necessary to package the output directory.

PR #840, opened 2020-11-20
feat: add a hanlder unit test for upstream and remove init

ShiningRush:
This pr is a demo about how to do a real `unit test` in handler.
I also remove the `init` in `conf` and `log`, go is recommended to use explicit initial function instead of implicit `init`

### DIFF
--- a/api/internal/handler/consumer/consumer_test.go
+++ b/api/internal/handler/consumer/consumer_test.go
@@ -19,6 +19,10 @@ package consumer
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/shiningrush/droplet/data"
+	"github.com/stretchr/testify/mock"
+	"net/http"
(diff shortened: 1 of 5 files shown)


ShiningRush:
Oh,it is automatically merged?what happened...
Maybe my misoperation : (

moonming:
please revert it

api/internal/handler/consumer/consumer_test.go
```diff
@@ -19,6 +19,10 @@ package consumer
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/shiningrush/droplet/data"
```
tokers, line 23: I think it's better to keep the convention about `import`: the first part is all standard libs, the second part is your local packages in your go module, and the last part is all third party packages.

PR #846, opened 2020-11-22
ci: enable CI for all release branches

nic-chen:


.github/workflows/backend-cli-test.yml
```diff
@@ -4,11 +4,11 @@ on:
   push:
     branches:
       - master
-      - v2.0
+      - v**
```
juzhiyuan, line 7: Someone may don't know what happened here, could we add a README instead of multiple comments under .github folder?

nic-chen: But they may not read the README, I prefer to write comments directly, although it is more cumbersome.
What do you think?

juzhiyuan: hmmm not sure, `v**` is too mixed to understand, just waiting for others' comments.

ShiningRush: Why we not develop in `master` or `development`?

juzhiyuan: Good idea, could we merge 2.1 into 2.0 now? We should develop on the master branch indeed. 

cc @mooming

nic-chen: en, we should always develop on master branch,

moonming: yes, 
```
branches:
      - master
```
is good enough.

moonming:
@nic-chen please submit PR based on `master`, and merge to `master` too

nic-chen:
will close this pr and create a new one later.

Issue #851, opened 2020-11-23
bug: manager-api do not support add consumer with custom create time or update time


idbeta:
# Bug report

## Describe the bug
use manager-api add a consumer with custom create time or update time, 
but the consumer actually created does not use the custom time.

starsz:
Hi, @idbeta, I can't get your point. 
Can you give me an example or steps to reproduce the issue?

idbeta:
hi, @starsz you can refer tohttps://github.com/apache/apisix-dashboard/pull/830/files#diff-d6674616f38aabf4e6e7f91f04b057982213b9b3758fc363ffc140843db957e3R344-R385
this case is failed, it will return the current timestamp.

starsz:
OK, I reproduce the issue.

ShiningRush:
I don't think it is a bug, audit field should maintain by server not client.So we should not accept client audit field such as CreatedAt, CreatedBy etc.

juzhiyuan:
Not sure if it's a bug indeed, what's your opinion? I would be more prefer with @ShiningRush's opinion 🤔

cc @membphis @nic-chen @liuxiran @tokers

liuxiran:
> I don't think it is a bug, audit field should maintain by server not client.So we should not accept client audit field such as `CreatedAt`, `CreatedBy` etc.

agree +1.
If the user customizes these fields, it goes against the original intent of using these fields to record the time.

nic-chen:
> I don't think it is a bug, audit field should maintain by server not client.So we should not accept client audit field such as `CreatedAt`, `CreatedBy` etc.

agree +1

starsz:
OK. So the conclusion is that we should not accept the client audit fields such as CreatedAt, CreatedBy.
If the client put those fields, just ignore them.

juzhiyuan:
Should we add this tip to FAQ?

idbeta:
At the beginning, I referred to the processing of adminAPI. After reading your comments, I think that manager-api does not need to be the same as adminAPI at this point. I will modify the related test cases.

PR #860, opened 2020-11-23
fix: add consumer with custom create time or update time

starsz:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

fix: #851 
___
### Bugfix
- Description

info.CreateTime will be covered.

- How to fix?

If info.CreateTime is not zero value, don't new time value.


membphis (changes requested):
unit test case

ShiningRush:
I don't think it is a bug, audit field should maintain by server not client.So we should not accept client audit field such as `CreatedAt`, `CreatedBy` etc.

starsz:
Closed. For more details see: #851

PR #899, opened 2020-11-28
chore: add middleware to transform error and rename placeholder.txt

### LINKED ISSUE #568: [refactor][bug] bind params to struct error should response error with 400 status
now it response normal.

here is an example:

```
$ curl http://127.0.0.1:8080/apisix/admin/services/1  -X PUT -i -d '
{
    "upstream": {
        "nodes": {
            "127.0.0.1:8080": 1
        },
        "type": "roundrobin"
    },
    "labels": {
        "env": ["production", "release"]
    },
    "desc": "new service"
}'

```

ShiningRush:
- Add middleware to transform error, resolve #568 
- Rename `placeholder.txt` to `.gitkeep`, It is more suitable

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/lestrrat-go/strftime v1.0.3 // indirect
 	github.com/natefinch/lumberjack v2.0.0+incompatible
 	github.com/satori/go.uuid v1.2.0
-	github.com/shiningrush/droplet v0.2.2
+	github.com/shiningrush/droplet v0.2.3
 	github.com/shiningrush/droplet/wrapper/gin v0.2.0
 	github.com/sirupsen/logrus v1.7.0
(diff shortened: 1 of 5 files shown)


juzhiyuan (approved):
.gitkeep looks good to me

nic-chen (approved):

idbeta:
@ShiningRush I found that you only added test cases to the routed (uri).  should we add test cases in these situations? https://github.com/apache/apisix-dashboard/issues/568#issuecomment-734043715

idbeta:
> @ShiningRush I found that you only added test cases to the routed (uri). should we add test cases in these situations? [#568 (comment)](https://github.com/apache/apisix-dashboard/issues/568#issuecomment-734043715)

I have tested the problematic scenes, and there are indeed no problems.

PR #773, opened 2020-11-10
chore: introduce corba as the cli scafford

tokers:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance

- Related issues

https://github.com/apache/apisix-dashboard/issues/739

PR #900, opened 2020-11-28
feat: add version info into manager-api 

### LINKED ISSUE #853: feat: show some important information after started the `manager-api`
# Feature request

After starting manager-api, no information is output.
Make new users easily confused.

We can print some tips and make this step more friendly.

image

Show some important things for the user. 

For example: `manager-api is running, and listen on: http://127.0.0.1:8080.`

### LINKED ISSUE #866: feat: manager-api need show version info
# Feature request

Hi, I think manager-api needs version info. 

Like: `./manager-api version`

And when `manager-api` start, show the manager-api version.

releated #853 

In golang, there are some ways to manage version.

1. Create a file, named `version.go` and write the current version in this file.

2. Use go build ldflags. Such as 
`go build -ldflags "-X main.VERSION=1.0.0 -X 'main.BUILD_TIME=date' -X 'main.GO_VERSION=go version'"`
We can add this into Makefile.

I prefer the second solution.

starsz:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance

- Related issues

fix #866
fix #853


membphis (changes requested):
missing test case

api/VERSION
```diff
@@ -0,0 +1 @@
+v2.1-rc1
```
juzhiyuan, line 1: 🤔 Do we have to modify this file before releasing it? The release manager may forget this point sometimes. We should update docs or add some tips somewhere 😂

starsz: Yeah. It's important. I had checked the apisix project. It needs to update `version.lua` for every release.
The same as apisix-dashboard. 

Maybe we can add a checklist to the PR template?

juzhiyuan: Got it, we may have this tip to release guide 😃

starsz:
> missing test case

added.

nic-chen (approved):

api/build.sh
```diff
@@ -19,6 +19,9 @@ set -ex
 export ENV=local
 pwd=`pwd`
 
+VERSION=$(cat ./api/VERSION)
```
ShiningRush, line 22: We also need go version, it is useful to debug problem, and if we add the git version, it would be better to add branch name. and build date is also useful.
You can refer [here](https://github.com/m3db/m3/blob/master/scripts/go-build-ldflags.sh).

api/main.go
```diff
@@ -35,6 +35,16 @@ import (
 	"github.com/apisix/manager-api/log"
 )
 
+var Version string
+
+func printInfo() {
+	fmt.Fprint(os.Stdout, "The manager-api is running successfully!\n\n")
```
ShiningRush, line 41: Just keep version info short, welcome is unnecessary

api/main.go
```diff
@@ -35,6 +35,16 @@ import (
...
+	fmt.Fprintf(os.Stdout, "%-8s: %s\n", "Version", Version)
+	fmt.Fprintf(os.Stdout, "%-8s: %s:%d\n", "Listen", conf.ServerHost, conf.ServerPort)
+	fmt.Fprintf(os.Stdout, "%-8s: %s\n", "Loglevel", conf.ErrorLogLevel)
```
ShiningRush, line 44: Log info is described in config file, I think here show listen info is enough because this is the most important running info to user, how do you think?

api/build.sh
starsz: I agree with add go version. But why are branch name and build date useful, since we had git commit info.

In fact, I refer the [mosn](https://github.com/mosn/mosn/blob/2adbcbc57994288fa4bdeb86bf43973443233a0b/Makefile#L94)

api/main.go
starsz: Sorry for the PR title. It's not only the version info. It also fixes #853.
And `-v` will be implemented at https://github.com/apache/apisix-dashboard/pull/773. Manager-api need a cli scafford.

api/build.sh
ShiningRush: If here are many branch, how do you know to which branch the git version belong?

api/main.go
starsz: The listen info is also described in the config file. But we show it.
So I think it's more convenient for users to get the log file path location if the file path is described in `relative path` format.

api/main.go
ShiningRush: This is a different line with `listen on:`, this line dose not offer any helpful info to user.
The original issues demo is just only one line `manager-api is running, and listen on: http://127.0.0.1:8080.`
BTW: I think `manager-api is listen on : xxxx`  would be more simple,how do you think? @membphis @nic-chen

api/build.sh
starsz: Hi, I write the `abbreviated commit hash`.
If the commit hash is the same, we can consider the code is the same.

api/main.go
ShiningRush: Output listen info have two meanings:
- Tell user, manager api is running normal
- Show the serve info

If we output the log info, why we not output other config? We should give a standard to tell other developer what should be print with begin info instead of component's log,I  think `relative path` is not reasonable.

api/build.sh
ShiningRush: Code is same is not enough, user could build it with different build tag and option(such as enable static link and disable cgo etc.).

starsz: Got it. BTW, why we need to add the branch name?

ShiningRush: > If here are many branch, how do you know to which branch the git version belong?

For this, help us to locate at which branch git version is

nic-chen:
ping @starsz @ShiningRush

membphis:
@starsz I helped you resolve the conflict

membphis (approved):

membphis:
@starsz many thx, merged right now

Issue #638, opened 2020-10-29
add e2e test for config route with management fileds


nic-chen:
detail:

 - [ ]  add test case for config route with name and desc

 - [ ]  verify the route just configured with correct name and desc, and has create_time and update_time

 - [ ]  add test case for update route  with other name and desc

 - [ ]  verify the route just configured with correct name and desc, update_time has changed.

 - [ ]  add test case for config route with labels and filter_func

 - [ ]  add test case for config route with  non-existent fields

 - [ ]  verify routing correctness under multi routes

For background and description, please see:

https://github.com/apache/apisix-dashboard/issues/607

idbeta:
Hi, please assign it to me.

idbeta:
The frequency of using `filter_func` is low, it is recommended to move to the 2.2 milestone. What do you think? cc @nic-chen

nic-chen:
agree with that.

idbeta:
` verify routing correctness under multi routes`, this test point need to 
Now, there are few effective uris provided by the upstream service docker, can you provide an upstream, similar to nginx `location /`? In that case, no matter how many routes are created, they can all use different valid uris.

Issue #912, opened 2020-11-30
chore: use `Desc` to replace of `caseDesc`


membphis:
![image](https://user-images.githubusercontent.com/6814606/100576857-e854c280-3319-11eb-9fd4-495bc5ad8e2e.png)

keep one name style, same as others.

starsz:
Hi, @juzhiyuan. I checked it, and I think it was fixed.

See line 166.

https://github.com/apache/apisix-dashboard/blob/40d8f97de8770187599ff5963617049603020e92/api/test/e2e/base.go#L165-L181

PR #927, opened 2020-11-30
test: config route with management fileds (E2E)

idbeta:

- [x] E2E test
- [ ] New feature provided
- [ ] Improve performance

- Related issues
resolve https://github.com/apache/apisix-dashboard/issues/638

### DIFF
--- a/api/test/e2e/route_with_management fileds_test.go
+++ b/api/test/e2e/route_with_management fileds_test.go
@@ -30,7 +30,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 	tests := []HttpTestCase{
 		{
 			caseDesc: "config route with name and desc (r1)",
-			Object:   MangerApiExpect(t),
+			Object:   ManagerApiExpect(t),
 			Path:     "/apisix/admin/routes/r1",
 			Method:   http.MethodPut,
 			Body: `{
@@ -61,7 +61,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 		},
 		{
 			caseDesc:     "verify the route's content (r1)",
-			Object:       MangerApiExpect(t),
+			Object:       ManagerApiExpect(t),
 			Path:         "/apisix/admin/routes/r1",
 			Method:       http.MethodGet,
 			Headers:      map[string]string{"Authorization": token},
@@ -77,7 +77,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 	tests = []HttpTestCase{
 		{
 			caseDesc:     "delete the route (r1)",
-			Object:       MangerApiExpect(t),
+			Object:       ManagerApiExpect(t),
 			Method:       http.MethodDelete,
 			Path:         "/apisix/admin/routes/r1",
 			Headers:      map[string]string{"Authorization": token},
@@ -92,7 +92,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 	tests = []HttpTestCase{
 		{
 			caseDesc: "config route with name, desc, create_time, update_time",
-			Object:   MangerApiExpect(t),
+			Object:   ManagerApiExpect(t),
 			Path:     "/apisix/admin/routes/r1",
 			Method:   http.MethodPut,
 			Body: `{
@@ -125,7 +125,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 		},
 		{
 			caseDesc:     "verify the route's detail (r1)",
-			Object:       MangerApiExpect(t),
+			Object:       ManagerApiExpect(t),
 			Path:         "/apisix/admin/routes/r1",
 			Method:       http.MethodGet,
 			Headers:      map[string]string{"Authorization": token},
@@ -154,7 +154,7 @@ func TestRoute_with_name_desc(t *testing.T) {
 	tests = []HttpTestCase{
 		{
 			caseDesc: "update the route (r1)",
-			Object:   MangerApiExpect(t),
+			Object:   ManagerApiExpect(t),
 			Path:     "/apisix/admin/routes/r1",
 			Method:   http.MethodPut,
 			Body: `{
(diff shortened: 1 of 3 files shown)


api/test/e2e/route_with_management fileds_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/tidwall/gjson"
+)
+
+func TestRoute_with_name_desc(t *testing.T) {
+	tests := []HttpTestCase{
+		{
+			caseDesc: "config route with name and desc (r1)",
```
moonming, line 32: Why is the first letter lowercase of caseDesc?

api/test/e2e/route_with_management fileds_test.go
```diff
@@ -0,0 +1,282 @@
...
+			caseDesc: "config route with name, desc, create_time, update_time",
+			Object:   MangerApiExpect(t),
+			Path:     "/apisix/admin/routes/r1",
+			Method:   http.MethodPut,
+			Body: `{
+					"uri": "/hello",
+					"name": "jack",
+					"create_time": 1604046556,
+					"update_time": 1604046556,
+					"desc": "config route with name and desc",
```
moonming, line 103: Bad desc

api/test/e2e/route_with_management fileds_test.go
```diff
@@ -0,0 +1,282 @@
...
+			Object:       APISIXExpect(t),
+			Method:       http.MethodGet,
+			Path:         "/hello",
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusOK,
+			ExpectBody:   "hello world",
+			Sleep:        sleepTime,
+		},
+		{
+			caseDesc:     "verify the route's detail (r1)",
```
moonming, line 127: How is this test case different from the first one?

api/test/e2e/route_with_management fileds_test.go
```diff
@@ -0,0 +1,282 @@
...
+	time.Sleep(time.Duration(100) * time.Millisecond)
+	//get the route
+	basepath := "http://127.0.0.1:8080/apisix/admin/routes"
+	request, _ := http.NewRequest("GET", basepath+"/r1", nil)
+	request.Header.Add("Authorization", token)
+	resp, _ := http.DefaultClient.Do(request)
+	respBody, _ := ioutil.ReadAll(resp.Body)
+	createtime := gjson.Get(string(respBody), "data.create_time")
+	updatetime := gjson.Get(string(respBody), "data.update_time")
+	assert.NotEqual(t, createtime.String(), "")
```
moonming, line 151: Why not compare the time here?

moonming (changes requested):

api/test/e2e/route_with_management fileds_test.go
```diff
@@ -0,0 +1,282 @@
...
+	tests = []HttpTestCase{
+		{
+			caseDesc: "config route with name, desc, create_time, update_time",
+			Object:   MangerApiExpect(t),
+			Path:     "/apisix/admin/routes/r1",
+			Method:   http.MethodPut,
+			Body: `{
+					"uri": "/hello",
+					"name": "jack",
+					"create_time": 1604046556,
```
membphis, line 101: we should forbidden to specify `create_time`, `update_time`

api/test/e2e/route_with_management fileds_test.go
idbeta: This route is newly created, so the timestamp here is the current time, not the incoming time, so it is not convenient to compare specific values.

api/test/e2e/route_with_management fileds_test.go
idbeta: Now the case does not require the first letter of caseDesc has to be capitalized, all existing use cases are in lower case, I think it can be regarded as "standard".

api/test/e2e/route_with_management fileds_test.go
idbeta: This is just used to compare the case results, so it's good as long as it does not exist before.

api/test/e2e/route_with_management fileds_test.go
idbeta: But if manager-api wants to support the command line to create routes in the future, it probably should not be forbidden to specify `create_time`, `update_time`.

api/test/e2e/route_with_management fileds_test.go
idbeta: This case is to verify that input `create_time`, `update_time` can also create routes.

api/test/e2e/route_with_management fileds_test.go
idbeta: ping @moonming Can you review it when you have time?

api/test/e2e/route_with_management fileds_test.go
membphis: > @idbeta I think it can be regarded as "standard".

that is totally wrong. I created a related issue yesterday: https://github.com/apache/apisix-dashboard/issues/912
we'll fix this later. 

`caseDesc` is not a good field name

api/test/e2e/route_with_management fileds_test.go
membphis: `create_time` and `update_time` are fields maintained by manager-api, and we do not need the caller to care about them.

so I think we can delete `create_time` and `update_time` here.

api/test/e2e/route_with_management fileds_test.go
idbeta: ok, I see, we can fix it in `2.2`, what do you think? @nic-chen

api/test/e2e/route_with_management fileds_test.go
membphis: I created a new issue, we can it later: https://github.com/apache/apisix-dashboard/issues/933

idbeta: ok, I delete that part.

membphis:
@idbeta your branch is wrong, we can close this PR then you can submit a new one

membphis:
https://github.com/apache/apisix-dashboard/pull/927/files#diff-ea228bf7b6d1ad6b5bf1fa49208640eba9cfbe76f2451d923b104bc8fab20a6d

this is wrong commit

PR #947, opened 2020-12-02
fix: snowflake remove network dependency

### LINKED ISSUE #938: panic: dial udp 8.8.8.8:80: connect: network is unreachable
./manager-api   内网不能联外网，启动提示connect: network is unreachable
panic: dial udp 8.8.8.8:80: connect: network is unreachable
goroutine 1 [running]:
github.com/apisix/manager-api/internal/utils.GetOutboundIP(0x0, 0x0, 0x0)
	/root/apisix-dashboard/api/internal/utils/utils.go:61 +0x133
github.com/apisix/manager-api/internal/utils.init.0.func1(0x0, 0xecb95ae00, 0x0)
	/root/apisix-dashboard/api/internal/utils/utils.go:41 +0x2b
github.com/sony/sonyflake.NewSonyflake(0x0, 0x0, 0x0, 0xc0000e5eb0, 0x0, 0x3)
	/root/go/pkg/mod/github.com/sony/sonyflake@v1.0.0/sonyflake.go:75 +0x14a
github.com/apisix/manager-api/internal/utils.init.0()
	/root/apisix-dashboard/api/internal/utils/utils.go:39 +0xb4

gxthrj:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fix #938 
___
### Bugfix
- Description
remove network dependency for snowflake 

- How to fix?
Get IP in local

### DIFF
--- a/api/internal/utils/utils.go
+++ b/api/internal/utils/utils.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"strconv"
@@ -37,10 +38,13 @@ func init() {
 		}
 		salt = uint16(i)
 	}
-
+	ips, err := getLocalIPs()
+	if err != nil {
+		panic(err)
+	}
 	_sf = sonyflake.NewSonyflake(sonyflake.Settings{
 		MachineID: func() (u uint16, e error) {
-			return sumIP(GetOutboundIP()) + salt, nil
+			return sumIP(randomIP(ips)) + salt, nil
 		},
 	})
 	if _sf == nil {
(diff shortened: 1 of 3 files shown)


api/internal/utils/utils.go
```diff
@@ -37,10 +38,13 @@ func init() {
...
 	}
-
+	ips, err := getLocalIPs()
+	if err != nil {
+		panic(err)
+	}
 	_sf = sonyflake.NewSonyflake(sonyflake.Settings{
 		MachineID: func() (u uint16, e error) {
-			return sumIP(GetOutboundIP()) + salt, nil
+			return sumIP(randomIP(ips)) + salt, nil
```
nic-chen, line 47: how about sum all local ips?

membphis:
@ShiningRush please take a look at this RP when you have time ^_^

we need to fix this issue today

api/internal/utils/utils.go
gxthrj: I think one is enough, it is for machineID, one or all IPs are the same :)

nic-chen: may it meet 127.0.0.1 ？

gxthrj: No, `127.0.0.1` is filtered out

nic-chen: 👌

nic-chen (approved):

api/internal/utils/utils.go
ShiningRush: Here is a risk that different instance may be get the same IP in multiple network environment  (such as virtual network ), calculating all IP would be better when we can not confirm which is physical network interface.

membphis:
@gxthrj please take a look at the output of CI

image

api/internal/utils/utils.go
gxthrj: OK, you convinced me.
Done.

gxthrj:
> @gxthrj please take a look at the output of CI
> 
> image

Fixed.

api/internal/utils/utils.go
```diff
@@ -37,35 +38,50 @@ func init() {
...
+			ips = append(ips, ipNet.IP)
+		}
 	}
-	defer conn.Close()
+	return ips, nil
+}
 
-	localAddr := conn.LocalAddr().(*net.UDPAddr)
-	return localAddr.IP
+func randomIP(xs []net.IP) net.IP {
```
ShiningRush: So we can remove unused code and test case.
BTW: this is not a really `random`, because here use the default random seed,  so it will always output a certain int sequence.If we need to pick a ip, pick the first one would be better, it help to debug program.

ShiningRush (approved):

PR #948, opened 2020-12-02
fix: omit remote_addrs when users don't input it

### LINKED ISSUE #940: bug: when user not fill in `remote address` , `remote_addrs` should be [], but not [""]
# Bug report

## How to Reproduce

1. use dashboard to create a route without filling in `remote address`

2. wait a few minutes, will find some error log like 
```
2020/12/01 22:44:48 [error] 34404#36781528: *230358 [lua] config_etcd.lua:509: failed to fetch data from etcd: failed to check item data of [/apisix/routes] err:property "remote_addrs" validation failed: failed to validate item 1: object matches none of the requireds,  etcd key: /apisix/routes, context: ngx.timer
2020/12/01 22:44:48 [error] 34408#36781532: *230380 [lua] config_etcd.lua:509: failed to fetch data from etcd: failed to check item data of [/apisix/routes] err:property "remote_addrs" validation failed: failed to validate item 1: object matches none of the requireds,  etcd key: /apisix/routes, context: ngx.timer
2020/12/01 22:44:48 [error] 34409#36781533: *230352 [lua] config_etcd.lua:509: failed to fetch data from etcd: failed to check item data of [/apisix/routes] err:property "remote_addrs" validation failed: failed to validate item 1: object matches none of the requireds,  etcd key: /apisix/routes, context: ngx.timer

```

3. the route doesn't work

LiteSun:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bugfix
- Description

remote_addrs should not be [""]

close #940 

- How to fix?

___

### DIFF
--- a/web/src/pages/Route/transform.ts
+++ b/web/src/pages/Route/transform.ts
@@ -82,6 +82,7 @@ export const transformStepData = ({
       !Object.keys(step3DataCloned.script || {}).length ? 'script' : '',
       form1Data.hosts.filter(Boolean).length === 0 ? 'hosts' : '',
       form1Data.redirectOption === 'disabled' ? 'redirect' : '',
+      data.remote_addrs?.filter(Boolean).length === 0 ? 'remote_addrs' : '',
     ]);
   }
 
@@ -97,8 +98,8 @@ export const transformStepData = ({
     'redirect',
     'vars',
     'plugins',
-    'remote_addrs',
     form1Data.hosts.filter(Boolean).length !== 0 ? 'hosts' : '',
+    data.remote_addrs?.filter(Boolean).length !== 0 ? 'remote_addrs' : '',
   ]);
 };
 


web/src/pages/Route/transform.ts
```diff
@@ -82,6 +82,7 @@ export const transformStepData = ({
       !Object.keys(step3DataCloned.script || {}).length ? 'script' : '',
       form1Data.hosts.filter(Boolean).length === 0 ? 'hosts' : '',
       form1Data.redirectOption === 'disabled' ? 'redirect' : '',
+      data.remote_addrs?.filter(Boolean).length === 0 ? 'remote_addrs' : '',
```
juzhiyuan, line 85: Very ugly and not easy to maintain on the frontend's side, I agree with no matter empty array or omit that field should be valid for the backend.

cc @membphis @liuxiran @moonming @ShiningRush @tokers

moonming:
bad title

web/src/pages/Route/transform.ts
membphis: @juzhiyuan you can create a new issue at APISIX, we can discuss this first.

we can use a better way in milestone `2.2`.

liuxiran: fot the data validation, I agree with  @juzhiyuan , this will also be flexible for people who use api.

If Fe must comply with omit field which is not configed, we may choose more elegant way to implement this transform, e.g: 
write a until func to handle omit params.

juzhiyuan: Those ugly fallbacks would be dropped according to https://github.com/apache/apisix/issues/2924 , we use current way to be compatible with APISIX 2.1.

spacewander: Why `empty array` is equal to `omit that field`? They are quite different things. The former one is a `[]`, while the latter one is `nil`.

juzhiyuan: They are different indeed, but the frontend doesn't care about which array-like parameters should be omitted before sending request as well.

spacewander: > the frontend doesn't care about which array-like parameters should be omitted before sending requests

Yes, but the backends care. This change just shift the duty to the backend.
There are three problems:
1. We have two backends. It is hard to sync the validation.
2. jsonschema treat `empty array` and `omit array field` different. We have to write code manually twice.
3. What if one day we want to distinguish `[]` from `nil`?

juzhiyuan: resolved and discussing here[1]

[1] https://github.com/apache/apisix/issues/2924#issuecomment-737037777

juzhiyuan (approved):
Looks good to me from codes, but a more semantic title is needed.

membphis (approved):

juzhiyuan:
I merged this PR on the mobile app but it  doesn't request me to update commit message, so the original message is merged into master branch. Do I need to revert this?

cc @moonming @membphis

membphis:
`revert` is useless, it can not update the commit log. pay attention next time ^_^

Issue #839, opened 2020-11-20
feat: support to connect etcd with `username` and `password`


membphis:
`etcd` can specify access username/password and assign different permissions, which is more secure.

`manager-api` needs to support this feature.

membphis:
related PR: https://github.com/apache/apisix-dashboard/pull/811

juzhiyuan:
Will this be resolved in 2.2?

membphis:
I think `2.2` is fine. I add this issue to `2.2` milestone

juzhiyuan:
Who is glad to resolve this issue?

nic-chen:
close by #951

PR #951, opened 2020-12-02
feat: add etcd basic auth support

JieTrancender:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/apisix-dashboard/issues/839#issue-747461530
___
### Bugfix
- Description

- How to fix?

___


nic-chen:
@JieTrancender  need test case for it

JieTrancender:
> @JieTrancender need test case for it

i will finish it as soon as possible

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,30 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
+curl -L http://localhost:2379/v3/auth/role/add \
+  -X POST -d '{"name": "root"}'
+
+# grant root role to root user
+curl -L http://localhost:2379/v3/auth/user/grant \
+  -X POST -d '{"user": "root", "role": "root"}'
+
+# enable auth
+curl -L http://localhost:2379/v3/auth/enable -X POST -d '{}'
+
```
nic-chen, line 113: I think we could try to run `./manager-api` that should fail before modify etcd config.

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,30 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
+# modify etcd auth config
+sed -i '1,$s/# username: "root" # ignore this argument if not enable auth/username: "root"/g' conf.yaml
+sed -i '1,$s/# password: "123456" # ignore this argument if not enable auth/password: "apisix-dashboard"/g' conf.yaml
+
+./manager-api &
+sleep 3
```
nic-chen, line 119: we could try some curl here.

JieTrancender: it's right, i have a lazy just now.

JieTrancender:
@nic-chen  if there is any convenient method which i can make a test like the auto ci tool?
i have a test in macOS and centos, it's all not right by execing cli_test.sh

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,48 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
+  -X POST -d '{"name": "root"}'
+
+# grant root role to root user
+curl -L http://localhost:2379/v3/auth/user/grant \
+  -X POST -d '{"user": "root", "role": "root"}'
+
+# enable auth
+curl -L http://localhost:2379/v3/auth/enable -X POST -d '{}'
+
+./manager-api &
```
nic-chen, line 114: how about `./manager-api > output.log 2>&1` here ?

nic-chen:
> @nic-chen if there is any convenient method which i can make a test like the auto ci tool?
> i have a test in macOS and centos, it's all not right by execing cli_test.sh

I think centos is ok.

api/test/shell/cli_test.sh
```diff
@@ -111,12 +111,9 @@ curl -L http://localhost:2379/v3/auth/user/grant \
 # enable auth
 curl -L http://localhost:2379/v3/auth/enable -X POST -d '{}'
 
-echo "runing after etcd auth enable"
 ./manager-api &
 sleep 3
-# pkill -f manager-api
```
JieTrancender: @nic-chen  It's this code makes wrong. when init etcd fail, this process gets dead.  And after i try again, you are right, centos is ok.

nic-chen: Congratulations, good to hear that!

api/internal/handler/consumer/consumer_test.go
```diff
@@ -19,6 +19,7 @@ package consumer
 
 import (
 	"encoding/json"
+	"github.com/apisix/manager-api/conf"
```
nic-chen, line 22: Should be under the standard library.

nic-chen (approved):
LGTM

Thanks.

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,49 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
     echo "failed: failed to write log on right level"
     exit 1
 fi
+
+
+# etcd basic auth
+# add root user
+curl -L http://localhost:2379/v3/auth/user/add \
+  -X POST -d '{"name": "root", "password": "apisix-dashboard"}'
```
juzhiyuan, line 101: A little strange to use this password 🤔

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,49 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
+
+# modify etcd auth config
+sed -i '1,$s/# username: "root" # ignore this argument if not enable auth/username: "root"/g' conf/conf.yaml
+sed -i '1,$s/# password: "123456" # ignore this argument if not enable auth/password: "apisix-dashboard"/g' conf/conf.yaml
+
+./manager-api &
+sleep 3
+
+# validate process is right by requesting login api
+resp=$(curl http://127.0.0.1:9000/apisix/admin/user/login -X POST -d '{"username":"admin", "password": "admin"}')
```
ShiningRush, line 131: This API does not access store, we would better add a case to access store such as `try create a consumer`

api/test/shell/cli_test.sh
JieTrancender: so, want more complex like **xdRX3UOis2CY5AWI** or more easy like **root**.
this is just a auth verification test, i think it's all ok.

api/test/shell/cli_test.sh
JieTrancender: yes, this api does't access store. but it's enough for verifing this test, if not this process has been killed.

JieTrancender: if do other things like create a consumer, clear dirty is necessary, making easy thing gets complex.

ShiningRush: I think the discussion here is based on our goals:
- If we only need to confirm that the process is alive, then accessing the `/ping` is a better choice, the `/login` has nothing to do with the modification this time
- If we want to verify that the program can access ETCD actually, we should access an interface related to it. Because now e2e test is run with docker-compose, so I think not to clean dirty is fine, if you ensure consumer id is not same with other test cases. 

I have not encountered a situation where the manager-api alive but cannot access the ETCD normally, so I think it is acceptable to just verify the process survival (although the latter would be better)

JieTrancender: ye, you are right, i have done it.

api/test/shell/cli_test.sh
juzhiyuan: more semantic password like root or password.

JieTrancender: although i disagreen with this, i also choise modifing it, just to pass it. 😄 😄

api/test/shell/cli_test.sh
```diff
@@ -93,3 +93,57 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
     echo "failed: failed to write log on right level"
     exit 1
 fi
+
+
+# etcd basic auth
+# add root user
+curl -L http://localhost:2379/v3/auth/user/add \
+  -X POST -d '{"name": "root", "password": "root"}'
```
tokers, line 101: Actually when you use `-d` option, the method is set to `POST` by default.

JieTrancender: ok

ShiningRush (approved):

nic-chen:
merged. thanks very much. @JieTrancender

PR #964, opened 2020-12-04
WIP: Refactor online debug

liuxiran:

- [x] Bugfix

- Related issues
fix: https://github.com/apache/apisix-dashboard/issues/835
_

Based on the design of proposal, there two problems found in the actual implementation process:

1.  The **original** design was to **create a new page**. Once implemented, I experienced myself, I have to switch pages to check out the route url I want to debug.
Finally, borrowed from the plugin config, I used `Draw` to show the debug page, so that use can do debug in the route list page

2020-12-04 13-42-20屏幕截图

2020-12-04 13-53-40屏幕截图

2. A new API  `POST /apisix/admin/debug-request-forwarding` has been added to the manager-api to help FE forward the debug request. Due to browser‘s same origin policy, some interfaces are not set `Access-Control-Allow-*` headers , simple FE requests cannot succeed.


api/internal/handler/route/route_test.go
```diff
@@ -985,6 +985,32 @@ func TestRoute(t *testing.T) {
...
+	  "uri": "/get"
+	}`
+	err = json.Unmarshal([]byte(reqBody), route7)
+	assert.Nil(t, err)
+	ctx.SetInput(route7)
+	ret, err = handler.Create(ctx)
+	assert.NotNil(t, err)
+	portTransmit := &ParamsInput{}
+	// FIXME
+	reqBody = `{"url":"http://127.0.0.1:9080/get","method":"GET"}`
```
liuxiran, line 1008: needs @nic-chen's help~

the fixed address is not appropriatly here, however, it is also not appropriate to add a configuration item just for test case, do you have any suggestions?

nic-chen: @liuxiran 
I think this should be E2E test, not unit test.

we could finish the job on APISIX then go ahead with this.

nic-chen: And I think this PR should be split into two, api and web separate, what do you think?

liuxiran: thank for your advice @nic-chen , let me try to update the test case, and split the pr.

PR #1002, opened 2020-12-10
fix: spelling

jbampton:

### DIFF
--- a/api/conf/schema.json
+++ b/api/conf/schema.json
@@ -3217,21 +3217,21 @@
 				"minProperties": 1,
 				"properties": {
 					"body": {
-						"description": "new body for repsonse",
+						"description": "new body for response",
 						"type": "string"
 					},
 					"body_base64": {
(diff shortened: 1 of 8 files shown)


moonming (approved):

nic-chen (approved):

api/conf/schema.json
```diff
@@ -3217,21 +3217,21 @@
 				"minProperties": 1,
 				"properties": {
 					"body": {
-						"description": "new body for repsonse",
+						"description": "new body for response",
```
membphis, line 3220: I think we need to update the schema of APISIX too

@jbampton how did you find those problems? Any tools?

membphis (approved):

api/conf/schema.json
jbampton: Hey @membphis I use various tools like linters, spell checkers and also the find and replace function in my editor 👍

PR #1024, opened 2020-12-11
fix: remove the dependency on git from the build script

nic-chen:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

___
### Bugfix
- Description

The source code package we will release does not contain git info, but the build script relies on the git command to obtain the git log version, which cause an error.

- How to fix?

remove the dependency on git from the build script

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -34,7 +34,7 @@ jobs:
           VERSION=$(cat ./api/VERSION)
           mkdir release
           git clean -Xdf
-          tar -zcvf apache-apisix-dashboard-${VERSION}-src.tgz \
+          tar -zcvf release/apache-apisix-dashboard-${VERSION}-src.tgz \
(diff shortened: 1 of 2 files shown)


membphis (approved):
LGTM

membphis:
rerun the CI ?

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,67 @@
...
+      - uses: actions/checkout@v2
+
+      - name: setup go
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.13'
+
+      - name: release and build
+        run: |
+          VERSION=$(cat ./api/VERSION)
```
juzhiyuan, line 34: This is a duplicate file like `Makefile`, could we use only one of them?

juzhiyuan (approved):
This CI would be improved in some ways.

Issue #546, opened 2020-10-09
[refactor] license issue of refactored `api` 


nic-chen:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirements
- [ ] Feature or performance improvement
- [x] Other

___
### Requirement or improvement

need to use the right way to fix license issue because some code is copied from Kubernetes dashboard, and current way is not right.

#535 
https://github.com/apache/apisix-dashboard/pull/535/files#r499468721

moonming:
ok, I will fix it

nic-chen:
add:
https://github.com/apache/apisix-dashboard/pull/551/files#r503025206

juzhiyuan:
Should notice this issue before releasing, cc @LiteSun \

LiteSun:
We need to remove `api/internal/core/store/validate_mock.go` and `api/internal/core/storage/storage_mock.go` manually before releasing.

PR #1063, opened 2020-12-17
fix: when create a route without ID in request body by Admin API, it can't be displayed in Manager API

### LINKED ISSUE #1041: failed to load the route in etcd
here are the routes in my `etcd`, I created them by a python script:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
import grequests

payloads = {}
for id in range(100):
    payload = {
                "uri": "/test_%s"%id,
                "upstream": {
                    "type": "roundrobin",
                    "nodes": {
                        "127.0.0.1:7000": 1
                    }
                }
    }
    payloads[id]=payload
unsent_request = []
header = {"X-API-KEY":"edd1c9f034335f136f87ad84b625c8f1"}
for id in payloads:
    print(id)
    unsent_request.append(grequests.post("http://127.0.0.1:9080/apisix/admin/routes/" ,json=payloads[id], headers = header))
    responses = grequests.map(unsent_request)
for response in responses:
    print(response.content)
    response.close()
```

the Manager API can not load them, here are the routes in etcd:

[routes.log](https://github.com/apache/apisix-dashboard/files/5684342/mm.log)

nic-chen:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

fix #1041
___
### Bugfix
- Description

When using Admin API to create resources such as route, the ID may not be placed in the content body. 
Manager API uses ID as the key, and will consider all data without ID value in the content body to be the same record, so that only one record is displayed and the others are not displayed.

- How to fix?

when a resource without ID value in the content body, get ID from resource key.


membphis (approved):

membphis (changes requested):
CI failed, E2E test cases are fine

membphis (approved):

api/internal/core/storage/storage_mock.go
```diff
@@ -1,5 +1,19 @@
-// Code generated by mockery v1.0.0. DO NOT EDIT.
-
+/*
```
tokers, line 1: It is a auto-generated file, why not re-generate it.

api/test/e2e/id_compatible_test.go
```diff
@@ -335,3 +335,73 @@ func TestID_Crossing(t *testing.T) {
...
+			Method:       http.MethodGet,
+			Path:         "/hello",
+			ExpectStatus: http.StatusNotFound,
+			Sleep:        sleepTime,
+		},
+		{
+			caseDesc: "create route that has no ID in request body by admin api",
+			Object:   APISIXExpect(t),
+			Method:   http.MethodPut,
+			Path:     "/apisix/admin/routes/r1",
```
gxthrj: Let me remind the issue in https://github.com/apache/apisix-ingress-controller/issues/125.
This is a scene like this: 
1. Add route with POST method ( admin api), APISIX will generate an id , such as `00000000000000000274 `;
2. And then read the Route from dashboard, the id is empty( id = "");

I think we should test the POST scenario.

api/internal/core/storage/storage.go
```diff
@@ -33,6 +33,11 @@ type WatchResponse struct {
 	Canceled bool
 }
 
+type Data struct {
```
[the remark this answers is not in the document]
nic-chen: OK, it is clearer.

api/internal/core/storage/storage_mock.go
nic-chen: for [license issue](https://github.com/apache/apisix-dashboard/issues/546#issuecomment-718371164), we'd better maintain the mock file manually.

tokers:
@nic-chen The e2e cases failed.

api/test/e2e/id_compatible_test.go
nic-chen: added.

api/internal/core/store/store.go
```diff
@@ -54,7 +54,7 @@ type GenericStore struct {
 type GenericStoreOption struct {
 	BasePath   string
 	ObjType    reflect.Type
-	KeyFunc    func(obj interface{}) string
+	KeyFunc    func(obj interface{}, key string) string
```
ShiningRush, line 57: Wow, this is not a good way, `KeyFunc` is responsible to ensure resource key, not the place to do compatible logic.
We can do it in `StringToObjPtr`.

nic-chen: updated.

juzhiyuan (approved):

Issue #566, opened 2020-10-20
bug: need to check ID between path and body


nic-chen:
## Describe the bug

When using the put method to modify or create routes or other resources, there are two places where you can pass ID, path and body. 
If the IDS passed in these two places are inconsistent, we cannot be sure which one is correct.

Example:
```
$ curl http://127.0.0.1:9000/apisix/admin/routes/1 -H 'Authorization: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2MDcwNDc1NTksImlhdCI6MTYwNzA0Mzk1OSwic3ViIjoiYWRtaW4ifQ.CnrVJEVp6hsSZvhUOmkqM3z1Y4_12deI2jgdJem4OdU' -X PUT -i -d '
{
    "id": "2",
    "uri": "/index.html",
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "39.97.63.215:80": 1
        }
    }
}'
```

## Solution

If the IDs passed by path and body are inconsistent, we should reject and return an error message.

BTW, `route`, `upstream`, `ssl`, `service`, `consumer` all have this issue.

jinchizhou:
I'll play around with this issue tomorrow!

nic-chen:
> I'll play around with this issue tomorrow!

that's great. thanks.

jinchizhou:
Hey @nic-chen , 

When setting up, I ran into the issue 
```
make api-run
cd api/ && go run .
{"level":"warn","ts":"2020-12-07T11:15:29.958-0800","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-b1e38523-73fc-4b26-a698-0f0d447376d4/127.0.0.1:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:2379: connect: connection refused\""}
panic: etcd get failed: context deadline exceeded`

goroutine 1 [running]:
main.main()
        /Users/jordanlee/Documents/jinchi/apisix-dashboard/api/main.go:53 +0x7c5
exit status 2
make: *** [api-run] Error 1
```

[...]

PR #1067, opened 2020-12-18
fix: error check to ensure path id doesn't conflict body id

jinchizhou:

- [x ] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fix https://github.com/apache/apisix-dashboard/issues/566
___
### Bugfix
- Description
Ensure Path ID doesn't conflict with Body ID
- Added test case for id to be string and int
- refactored code a bit to reuse a variable.
- 
- How to fix?
Check both the IDs if they both exist
___


juzhiyuan (approved):
LGTM from logics

juzhiyuan (commented):
Do we have Testcase to cover this?

jinchizhou:
Nope, I'll add some tests.

jinchizhou:
@nic-chen could you please take a look thanks!

jinchizhou:
FYI, I'm unable to request reviews

api/internal/handler/route/route_test.go
```diff
@@ -754,6 +754,20 @@ func TestRoute(t *testing.T) {
...
+	errRoute := &UpdateInput{}
+	errRoute.ID = "2"
+	err = json.Unmarshal([]byte(reqBody), errRoute)
+	assert.Nil(t, err)
+	ctx.SetInput(errRoute)
+	ret, err = handler.Update(ctx)
+	assert.NotNil(t, err)
+	assert.EqualError(t, err, "ID on path (2) doesn't match ID on body (1)")
+	assert.Equal(t, http.StatusBadRequest, ret.(*data.SpecCodeResponse).StatusCode)
+
```
nic-chen, line 770: For historical reasons, id supports int and stringing.
so we need test case for int ID in body.

jinchizhou: I think this test should support

api/internal/handler/route/route_test.go
```diff
@@ -408,352 +408,380 @@ func TestRoute(t *testing.T) {
...
-	          },
-	          "scale":0.832,
-	          "selected":{
-	              "id":"b93d622c-92ef-48b4-b6bb-57e1ce893ee3",
-	              "type":"node"
-	          }
-	      }
-	  }
-	}`
+	  ` + reqBodyCopy
```
nic-chen, line 750: we could use a simple route config
```go
reqBody = `{
    "id": "1",
    "uri": "/index.html",
    "upstream": {
        "type": "roundrobin",
        "nodes": [{
            "host": "www.a.com",
            "port": 80,
            "weight": 1
        }]
    }
}`
```

api/internal/handler/route/route_test.go
```diff
@@ -408,352 +408,380 @@ func TestRoute(t *testing.T) {
...
+	errRoute2 := &UpdateInput{}
+	errRoute2.ID = "2"
+	err = json.Unmarshal([]byte(reqBodyErr), errRoute2)
+	assert.Nil(t, err)
+	ctx.SetInput(errRoute2)
+	ret, err = handler.Update(ctx)
+	assert.NotNil(t, err)
+	assert.EqualError(t, err, "ID on path (2) doesn't match ID on body (%!s(float64=1))")
+	assert.Equal(t, http.StatusBadRequest, ret.(*data.SpecCodeResponse).StatusCode)
+
```
nic-chen, line 784: we should have a test case for IDs are same but ID in body is int:
```sh
curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "id": 1,
    "uri": "/index.html",
    "upstream": {
        "type": "roundrobin",
        "nodes": [{
			"host": "www.a.com",
			"port": 80,
			"weight": 1
		}]
    }
}'
```

jinchizhou:
Not sure why these CI tests are blocking. I verified that these tests are failing locally even without my changes.

nic-chen:
> Not sure why these CI tests are blocking. I verified that these tests are failing locally even without my changes.

I think it's due to https://github.com/apache/apisix-dashboard/pull/1067/files#diff-c45dc0115f67204842506d9284c9142195bbca6e06f449fe9fa367c7652e8c74R354-R358

We should judge whether it is a string, and convert it to a string if it's not.

jinchizhou:
I am only converting to a string if the type is float. The body id are guarantee to be a string or a float I assume? I use a new variable so that shouldn't be breaking any existing test cases.

membphis:
@nic-chen need your help ^_^

nic-chen (approved):

nic-chen:
merged. thanks @jinchizhou

PR #1072, opened 2020-12-18
feat: support labels list 

### LINKED ISSUE #861: feat: Route group feature
# why
In order to allow users to filter out routes with a specified group from a large number of routes, I propose to add the route groups feature.

# how
## front-end
In order to finish this feature. There are 3 steps need to do.
- Adding group options to the edit or create route page.
image

- Route group list page
image

- Create or Edit route group page
image

## back-end
Create or update API to match front-end requirements.
- Create route and update route API need to update.
- Route Group List API.
- Create and update route group API.

# APISIX
Need to discuss how to implement route group requirements.

Free to comment here.

cc @juzhiyuan @liuxiran @gxthrj @nic-chen @membphis

starsz:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance

- Related issues
fix #861 
___

Support labels list.


api/internal/handler/label/label.go
```diff
@@ -0,0 +1,260 @@
...
+
+	typ := input.Type
+	reqLabels, err := utils.GenLabelMap(input.Label)
+	if err != nil {
+		return &data.SpecCodeResponse{StatusCode: http.StatusBadRequest},
+			fmt.Errorf("%s: \"%s\"", err.Error(), input.Label)
+	}
+
+	var items []interface{}
+	switch typ {
```
juzhiyuan, line 142: shortname?

starsz: Because the `type` is a keyword in golang.So I often use `typ` instead of `type`.

api/internal/handler/label/label.go
```diff
@@ -0,0 +1,260 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package label
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/shiningrush/droplet/data"
```
tokers, line 22: Put this package after the standard libs.

api/internal/handler/label/label.go
```diff
@@ -0,0 +1,260 @@
...
+
+var _ json.Marshaler = Pair{}
+
+type Pair struct {
+	Key string
+	Val string
+}
+
+func (p Pair) MarshalJSON() ([]byte, error) {
+	res := fmt.Sprintf("{\"%s\":\"%s\"}", p.Key, p.Val)
```
tokers, line 55: It's not a good way, the JSON string will be broken if `p.Key` or `p.Val` has `"` in contents.

api/internal/handler/label/label.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.GET("/api/labels/:type", wgin.Wraps(h.List,
+		wrapper.InputType(reflect.TypeOf(ListInput{}))))
+}
+
+type ListInput struct {
+	Type  string `auto_read:"type,path" validate:"required"`
+	Label string `auto_read:"label,query"`
+	store.Pagination
```
tokers, line 77: Well, we should put embed member at the top of struct definition.

api/internal/handler/label/label.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func getMatch(reqLabels, labels map[string]string) map[string]string {
```
tokers, line 80: You can reference the label implementation of istio: https://github.com/istio/istio/blob/0c0cf2d6fb/pkg/config/labels/instance.go#L55.

Encapsulate the getMatch as `SubsetOf` will be more semantic.

api/internal/handler/label/label.go
starsz: Oh No. I think my goland plugin doesn't work.
Let me fix it.

api/internal/handler/label/label.go
starsz: Good idea.

api/internal/handler/label/label.go
starsz: I think it's ok because `key` and `val` come from req args. I will add some tests to ensure it.
And do you have some suggestion?

starsz: I think it's ok because `key` and `val` come from req args. I will add some tests to ensure it.
And do you have some suggestions?

api/internal/handler/label/label.go
starsz: OK.

api/internal/handler/label/label.go
starsz: I think it's not a problem. If '"' in contents, Json.Marshal will return an error.

api/test/e2e/label_test.go
```diff
@@ -0,0 +1,229 @@
...
+			caseDesc:     "get service label",
+			Object:       ManagerApiExpect(t),
+			Method:       http.MethodGet,
+			Headers:      map[string]string{"Authorization": token},
+			Path:         "/api/labels/service",
+			ExpectStatus: http.StatusOK,
+			ExpectBody:   "{\"build\":\"16\"},{\"env\":\"production\"},{\"extra\":\"test\"},{\"version\":\"v2\"}",
+		},
+		{
+			caseDesc:     "get all label",
```
nic-chen, line 164: need test cases for paging and both searching and paging.

starsz: Added.

api/internal/handler/label/label.go
tokers: No, it's a bug if you don't handle the special characters correctly in server side.

That's OK if you can make sure that no `"` in key and value.

starsz: fixed.And add some tests.

api/internal/handler/label/label_test.go
```diff
@@ -0,0 +1,312 @@
...
+func TestPair_MarshalJSON(t *testing.T) {
+	p := Pair{Key: "test_key", Val: `test_val`}
+	content, err := json.Marshal(p)
+	assert.Nil(t, err, nil)
+	assert.Equal(t, `{"test_key":"test_val"}`, string(content))
+
+	p = Pair{Key: "test_key", Val: `test_val"`}
+	content, err = json.Marshal(p)
+	assert.Nil(t, err, nil)
+	assert.Equal(t, `{"test_key":"test_val\""}`, string(content))
```
tokers, line 48: Can we also use `json.Unmarshal` to verify the manual constructed JSON?

starsz: A piece of cake.

api/internal/handler/label/label.go
```diff
@@ -0,0 +1,261 @@
...
+		routeStore:    store.GetStore(store.HubKeyRoute),
+		serviceStore:  store.GetStore(store.HubKeyService),
+		upstreamStore: store.GetStore(store.HubKeyUpstream),
+		sslStore:      store.GetStore(store.HubKeySsl),
+		consumerStore: store.GetStore(store.HubKeyConsumer),
+	}, nil
+}
+
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.GET("/api/labels/:type", wgin.Wraps(h.List,
```
nic-chen, line 71: why not use prefix `/apisix/admin` here ?

juzhiyuan: @starsz maybe not family with that rule? Let's help to update it.

juzhiyuan: oh no I find there have a lot of APIs that start with `/api/`, just waiting for @starsz's reply.

starsz: Oh.Sorry. I remember that we discuss use `/api/`?

https://github.com/apache/apisix-dashboard/issues/861#issuecomment-738519671.

Let me fix it.

tokers (approved):

nic-chen (approved):

PR #1073, opened 2020-12-18
ci: fix e2e test cases unstable

nic-chen:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

___
### Bugfix
- Description

When testing balancing, the statistics on different upstream accesses often deviate from expectations, causing test cases to fail.

- How to fix?

use single worker APISIX to test balancing.


api/test/docker/apisix_config2.yaml
```diff
@@ -42,6 +42,7 @@ apisix:
 
 nginx_config:
   error_log_level: "debug"
+  worker_processes: "1"
```
juzhiyuan, line 45: 1 is the default value for worker_processes, why we need to add it? 🤔

nic-chen: default is 'auto'

juzhiyuan: fine, I find it https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L126

juzhiyuan (approved):

LiteSun (approved):

PR #1049, opened 2020-12-14
feat: support patch in route module in order to support publish/offline

liuxiran:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [x] New feature provided

- Related issues

BE of https://github.com/apache/apisix-dashboard/issues/837
support for https://github.com/apache/apisix-dashboard/pull/991
borrow from https://github.com/apache/apisix-dashboard/pull/1005
depends on APISIX release with https://github.com/apache/apisix/issues/2737 and sync `https://github.com/apache/apisix-dashboard/blob/master/api/conf/schema.json`

PR #1081, opened 2020-12-20
Refactor publish offline be 

liuxiran:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance

- Related issues
split from #1049 

---

support patch in route module


api/filter/schema.go
```diff
@@ -136,6 +144,25 @@ func handleSpecialField(resource string, reqBody []byte) ([]byte, error) {
...
 
+func handleDefaultValue(resource string, reqBody []byte) ([]byte, error) {
+	// go jsonschema lib doesn't support setting default values, so we need to set for some fields necessary
+	if resource == "routes" {
+		var route map[string]interface{}
+		err := json.Unmarshal(reqBody, &route)
+		if err != nil {
+			return reqBody, fmt.Errorf("read request body failed: %s", err)
+		}
+		if _, ok := route["status"]; !ok {
```
juzhiyuan, line 155: Just to make sure, 
1. `!ok` will be truthy when `ok` is `falsy`, If one route is not enabled, then after this filter, the status will be?
2. The `status` now is enabled by default after created, if that route has something wrong, we have to offline it after published? 🤔 I'm not sure if it's ok.

nic-chen: `if _, ok := route["status"]; !ok {` is checking whether there is a `status` field in the route config, if not, use the default value, otherwise use the submitted `status` value.

juzhiyuan: Got it! Another question is should we enable it by default? If there has something wrong with that route, we have to offline it 🤔

juzhiyuan: Q3: Could we debug an offline route in `OnlineDebug` feature? I'm not sure if `OnlineDebug` means debug online routes 😂

liuxiran: > 2. The `status` now is enabled by default after created, if that route has something wrong, we have to offline it after published? 🤔 I'm not sure if it's ok.

1. we will pass `status: 1` when create a route via fe, if user did not disable this field
2. create route without `status`,  golang will treat this field as 0, which goes against the original intent defined in APISIX so @nic-chen add these code to solve this problem.

juzhiyuan: Got it, thanks!

liuxiran: > feature

we can debug a online route with `APISIX_IP:PORT/url` via Online debug function
and we can debug a offline route with `upstream_IP:PORT/url` via Online debug function

nic-chen (approved):

api/internal/core/entity/entity.go
```diff
@@ -83,6 +83,7 @@ type Route struct {
 	ServiceProtocol string                 `json:"service_protocol,omitempty"`
 	Labels          map[string]string      `json:"labels,omitempty"`
 	EnableWebsocket bool                   `json:"enable_websocket,omitempty"`
+	Status          uint8                  `json:"status"`
```
tokers, line 86: We have type `Status`, so why use `uint8` here?

liuxiran: `status` only expect 0 or 1, so define the type of `Status` is `uint8`

nic-chen: fixed.

juzhiyuan (approved):
LGTM from E2E test

PR #1112, opened 2020-12-23
feat: add test cases for rejecting request data if it contains an empty array

### LINKED ISSUE #984: reject request data if it contains an empty array, eg: "remote_addrs": []
To add a route, if the `remote_addrs` field is an empty array, the matching behavior is ambiguous.

In order to avoid ambiguity and keep the API design simple and clear, we should reject the value as an empty array.

Here is an example, should return `400`:

```
$ curl http://127.0.0.1:9080/apisix/admin/routes/1 -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -X PUT -i -d '
{
    "uri": "/index.html",
    "remote_addrs": [],
    "enable_websocket": true,
    "upstream": {
        "type": "roundrobin",
        "nodes": {
            "39.97.63.215:80": 1
        }
    }
}'
HTTP/1.1 400 Bad Request
... ...
```

**Here is the todo list(from the discussion)**

* `Admin API`: reject request data if it contains an empty array, eg: `"remote_addrs": []`. `Code + Doc`
* `Manager API`: same as `Admin API`.
* `Dashbaord`: keep adding hacking codes to the frontend. `Code`

The field `hosts`, `uris` and `methods`, they have the same issue, we should fix them in the same way too.

original comment: https://github.com/apache/apisix/issues/2924#issuecomment-740405671

related issue: https://github.com/apache/apisix/issues/2991

nic-chen:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [x] Test case

issue has resolved by https://github.com/apache/apisix/pull/3008
just add e2e test cases for it

- Related issues
close #984

### DIFF
--- a/api/test/e2e/route_test.go
+++ b/api/test/e2e/route_test.go
@@ -491,25 +491,6 @@ func TestRoute_With_Empty_Array(t *testing.T) {
 			ExpectStatus: http.StatusBadRequest,
 			ExpectBody:   `{"code":10000,"message":"schema validate failed: (root): Must validate one and only one schema (oneOf)\n(root): Must validate all the schemas (allOf)\nremote_addrs: Array must have at least 1 items"}`,
 		},
-		//{
-		//	caseDesc: "create route with empty remote_addrs and remote_addr",
-		//	Object:   ManagerApiExpect(t),
-		//	Path:     "/apisix/admin/routes/r1",
-		//	Method:   http.MethodPut,
-		//	Body: `{
-		//		"uri": "/hello",
-		//		"methods": [],
-		//		"upstream": {
-		//			"nodes": {
-		//				"172.16.238.20:1980": 1
-		//			},
-		//			"type": "roundrobin"
-		//		}
-		//	}`,
-		//	Headers:      map[string]string{"Authorization": token},
-		//	ExpectStatus: http.StatusBadRequest,
-		//	ExpectBody:   `{"code":10000,"message":"schema validate failed: (root): Must validate one and only one schema (oneOf)\n(root): Must validate all the schemas (allOf)\nmethods: Array must have at least 1 items"}`,
-		//},
 		{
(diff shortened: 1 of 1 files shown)


api/test/e2e/route_test.go
```diff
@@ -392,3 +392,135 @@ func TestRoute_Patch(t *testing.T) {
 		testCaseCheck(tc)
 	}
 }
+
+//uris methods remote_addrs
+func TestRoute_With_Empty_Array(t *testing.T) {
+	tests := []HttpTestCase{
+		{
+			caseDesc: "create route with empty hosts and host",
```
tokers, line 400: typo: with empty hosts and hosts?

api/test/e2e/route_test.go
```diff
@@ -392,3 +392,135 @@ func TestRoute_Patch(t *testing.T) {
...
+						"172.16.238.20:1980": 1
+					},
+					"type": "roundrobin"
+				}
+			}`,
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusBadRequest,
+			ExpectBody:   `{"code":10000,"message":"schema validate failed: (root): Must validate one and only one schema (oneOf)\n(root): Must validate all the schemas (allOf)\nremote_addrs: Array must have at least 1 items"}`,
+		},
+		//{
```
tokers, line 494: Remove these commented lines if we don't need them.

api/test/e2e/route_test.go
nic-chen: they are two different fields.

api/test/e2e/route_test.go
nic-chen: removed. @tokers

tokers (approved):

gxthrj (approved):

Issue #922, opened 2020-11-30
test: E2E test case unstable


membphis:
https://github.com/apache/apisix-dashboard/runs/1474065168#step:5:91

![image](https://user-images.githubusercontent.com/6814606/100618235-849cba80-3356-11eb-88a1-c1d15df15577.png)

juzhiyuan:
![image](https://user-images.githubusercontent.com/2106987/103138304-0f7d9480-470c-11eb-834e-c02813c86343.png)

FYI: @starsz is working on this issue.

starsz:
Yes. Please assign it to me.

PR #1139, opened 2020-12-28
fix: minish resync_delay for e2e test 

starsz:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

#922
___
### Bugfix
- Description

Because the resync_delay's default value is 5.
So apisix will sleep 5 or more seconds in sync data from etcd.
This will cause the e2e test to fail.

See more details: https://github.com/apache/apisix/pull/2977

- How to fix?

Set the resync_delay to 0.5 seconds.
___


api/test/docker/apisix_config.yaml
```diff
@@ -42,6 +42,7 @@ apisix:
 
 nginx_config:
   error_log_level: "debug"
+  worker_processes: "1"
```
membphis, line 45: need a comment, only for debug now

api/test/e2e/global_rule_test.go
```diff
@@ -79,7 +80,7 @@ func TestGlobalRule(t *testing.T) {
 			ExpectStatus:  http.StatusOK,
 			ExpectBody:    "hello world",
 			ExpectHeaders: map[string]string{"X-VERSION": "1.0"},
-			Sleep:         sleepTime,
+			Sleep:         longSleepTime,
```
membphis, line 83: need some comment too.

membphis: And please update other similar points

starsz: OK. I had added the comment at base.go already.

https://github.com/starsz/apisix-dashboard/blob/e836046319fe37920f0d65c5a48d19aaa96b7d54/api/test/e2e/base.go#L185-L186

api/test/docker/apisix_config.yaml
starsz: Only for debug? Maybe I will add it to master if it can make the e2e test passed.

membphis: we need to find the real bug. 

I think we should remove this field in E2E case.

membphis:
@starsz CI failed again. please take a look

starsz:
Hi, @membphis @nic-chen  Please have a look. I found the cause of the problem.

nic-chen (approved):

tokers (approved):

PR #1151, opened 2020-12-30
fix: search labels with the same key 

### LINKED ISSUE #1130: It doesn't return all records when seaching routes with labels
# Bug report

## Describe the bug

image

image

image

```sh
curl 'http://139.217.190.60/apisix/admin/routes?label=version%3Av1%2Cversion%3Av2&page=1&page_size=20'
```

Not sure if it's a bug when searching route by labels 🤔 It should return 2 records instead of 1 IMO.

cc @idbeta @nic-chen

starsz:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

fix #1130 
___
### Bugfix
- Description

If the label argument with the same key, the previous value will be covered by the latter value in genLabelMap.
So the response is not correct.

- How to fix?

Use the `key:value` as a key in genLabelMap.


membphis (commented):
LGTM, I only reviewed the E2E test cases

nic-chen (commented):
We use `OR` or `AND` when there are multiple conditions?

starsz:
> We use `OR` or `AND` when there are multiple conditions?

OR. It is used for filter routes by label.

nic-chen (approved):

api/internal/handler/label/label.go
```diff
@@ -78,15 +78,18 @@ type ListInput struct {
...
 
 	var res = make(map[string]string)
 	for k, v := range labels {
-		l, exist := reqLabels[k]
-		if exist && ((l == "") || v == l) {
+		if _, exist := reqLabels[k]; exist {
+			res[k] = v
+		}
+
+		if _, exist := reqLabels[k+":"+v]; exist {
```
juzhiyuan, line 92: `[k+":"+v]` formmated?

api/test/e2e/label_test.go
```diff
@@ -200,6 +200,16 @@ func TestLabel(t *testing.T) {
 			ExpectStatus: http.StatusOK,
 			ExpectBody:   "{\"build\":\"16\"},{\"build\":\"17\"}",
 		},
+		{
+			Desc:         "get labels with the same key (key = build)",
+			Object:       ManagerApiExpect(t),
+			Method:       http.MethodGet,
+			Headers:      map[string]string{"Authorization": token},
+			Query:        "label=build:16,build:17",
+			Path:         "/apisix/admin/labels/all",
```
juzhiyuan, line 209: 🤔 What will happen if calling `/labels` only?

api/test/e2e/route_with_management_fileds_test.go
```diff
@@ -333,6 +333,17 @@ func TestRoute_search_by_label(t *testing.T) {
...
 		},
+		{
+			Desc:         "search the route by label (combination)",
+			Object:       ManagerApiExpect(t),
+			Path:         "/apisix/admin/routes",
+			Query:        "label=build:16,build:17",
+			Method:       http.MethodGet,
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusOK,
+			ExpectBody:   "\"total_size\":2",
```
juzhiyuan, line 344: just to confirm, is this all contents in response?

api/internal/handler/label/label.go
starsz: Yes.It's the behavior of IDE.

tokers (approved):

api/test/e2e/label_test.go
starsz: It will return `{"error_msg":"404 Route Not Found"}`.

api/test/e2e/route_with_management_fileds_test.go
starsz: Yes.

PR #1184, opened 2021-01-03
fix(cli): support running manager-api cmd in non-default diretory

### LINKED ISSUE #1147: bug: manager-api -p doesn't work
# Bug report

## Describe the bug

manager-api `-p` parameter doesn't work, it will prompt  `panic: fail to read configuration: ./conf/conf.yaml`

1. move `output` to another dir
2. use `ln -s` create manager-api soft links to `/usr/bin`
3. use manager-api -p `/path/to/output`
4. it will prompt  `panic: fail to read configuration: ./conf/conf.yaml`

@nic-chen

imjoey:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

Fixes #1147 .
___
### Bugfix
- Description

Running the prebuilt `manager-api` under non-default directory may lead to failed to start, see #1147 for details. This is mainly caused by the `-p` flag of `manager-api` is not worked as expected.

- How to fix?

Introduce a new value `test` for `ENV` environment to indicates the program is running a unit testing. If `ENV` is set to `test`, the original `init()` functions are retained, otherwise we explicitly calling `InitConf()` and `InitLogger()` functions to  initialization, right after receiving the input `WorkDir` parameter via `-p` flag. This would ensure configs and logger could be initialized just at the right time.

### DIFF
--- a/api/internal/conf/conf.go
+++ b/api/internal/conf/conf.go
@@ -100,11 +100,6 @@ type Config struct {
 	Authentication Authentication
 }
 
-// TODO: it is just for integration tests, we should call "InitLog" explicitly when remove all handler's integration tests
-func init() {
-	InitConf()
-}
-
 func InitConf() {
 	//go test
 	if workDir := os.Getenv("APISIX_API_WORKDIR"); workDir != "" {
--- a/api/internal/conf/conf.go
+++ b/api/internal/conf/conf.go
@@ -34,6 +34,7 @@ const (
 	EnvBETA  = "beta"
 	EnvDEV   = "dev"
 	EnvLOCAL = "local"
+	EnvTEST  = "test"
 
 	WebDir = "html/"
 )
@@ -98,6 +99,13 @@ type Authentication struct {
 type Config struct {
 	Conf           Conf
 	Authentication Authentication
+}
+
+// ENV=test is for integration tests only, other ENV should call "InitConf" explicitly
+func init() {
+	if env := os.Getenv("ENV"); env == EnvTEST {
+		InitConf()
+	}
 }
 
 func InitConf() {
--- a/api/internal/conf/conf.go
+++ b/api/internal/conf/conf.go
@@ -101,6 +101,7 @@ type Config struct {
 	Authentication Authentication
 }
 
+// TODO: we should no longer use init() function after remove all handler's integration tests
 // ENV=test is for integration tests only, other ENV should call "InitConf" explicitly
 func init() {
 	if env := os.Getenv("ENV"); env == EnvTEST {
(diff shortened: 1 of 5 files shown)


juzhiyuan:
very quick...

juzhiyuan:
Just ignored this check, it failed due to we only have 1 commit in this PR, once we pushed another, it will success. Not required.

image

imjoey:
> Just ignored this check, it failed due to we only have 1 commit in this PR, once we pushed another, it will success. Not required.
> 
> image

@juzhiyuan Got you~ Thanks.

imjoey:
CI passed and it's ready for review. ping @nic-chen @tokers , much appreciated.

nic-chen:
Thanks. We should add test case for it.

api/internal/conf/conf.go
```diff
@@ -100,9 +101,11 @@ type Config struct {
 	Authentication Authentication
 }
 
-// TODO: it is just for integration tests, we should call "InitLog" explicitly when remove all handler's integration tests
+// ENV=test is for integration tests only, other ENV should call "InitConf" explicitly
```
nic-chen, line 104: TODO should be retained here. We need to finally remove `init()`.

imjoey: > TODO should be retained here. We need to finally remove `init()`.

@nic-chen Done, thanks.

starsz:
Hi @imjoey, can you add some tests for it?

imjoey:
> Hi @imjoey, can you add some tests for it?

@starsz sure, I will add soon. Thanks.

imjoey:
@nic-chen @starsz Test case added and ready for review. Any feedback is appreciated. Thanks.

api/test/shell/cli_test.sh
```diff
@@ -126,6 +126,28 @@ if [[ $res != "hi~" ]]; then
 fi
 clean_up
 
+# run with -p flag out of the default directory
+workDir=$(pwd)
```
nic-chen, line 130: Looks like it's still in default dir. could we mv it to another dir to test it ?

imjoey: @nic-chen From my point of view, the default directory only makes sense for `manager-api` binary. But it's indeed not intuitive. I would renew this test case. Thanks.

nic-chen: oh, yes, that make sense.

nic-chen (approved):

api/test/shell/cli_test.sh
imjoey: @nic-chen Hi, PR renewed with a more intuitive way. Thanks for reviewing. 😄

juzhiyuan (approved):

PR #1207, opened 2021-01-05
fix(be): search by status on the route list page is invalid

### LINKED ISSUE #1189: bug: search by status on the route list page is invalid
# Bug report

## Describe the bug
image
I think the query string is wrong.
```
Request URL: http://127.0.0.1:9000/apisix/admin/routes?label=&page=1&page_size=20
Request Method: GET
```

Jaycean:
Signed-off-by: Jacean <jinchen_jacean@163.com>

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fix #1189 
___
### Bugfix
- Description
The backend does not judge the status field

- How to fix?
Add status field


api/internal/handler/route/route.go
```diff
@@ -218,7 +220,11 @@ func (h *Handler) List(c droplet.Context) (interface{}, error) {
 			if input.Label != "" && !utils.LabelContains(obj.(*entity.Route).Labels, labelMap) {
 				return false
 			}
-
+			
+		        status, _ := strconv.Atoi(input.Status)
```
starsz, line 224: Hi, I think this can be done before the `List` function. 
And we need to check the error. Return `400` if err is not nil.

api/internal/handler/route/route.go
```diff
@@ -218,7 +220,11 @@ func (h *Handler) List(c droplet.Context) (interface{}, error) {
...
+			if input.Status != "" && uint8(obj.(*entity.Route).Status) != uint8(status) {
```
starsz, line 225: There is an easy way.Just compare `strconv.Itoa(int(obj.(*entity.Route).Status))` and status.

Jaycean: Yes, but my biggest problem now is that no matter whether I use uint8 or string for comparison, my test cases can't pass. There's no problem with the function itself. Can you help me take a look at my test cases?  Thks

liuxiran:
This pr can work fine with the https://github.com/apache/apisix-dashboard/pull/1205, but the Unit tests always fail, and logically, we could not find anything wrong.

Please take a look when you are free @nic-chen @tokers @starsz @imjoey , many thanks~~

api/internal/handler/route/route.go
imjoey: @Jaycean you can try to `fmt.Sprintf("%d", obj.(*entity.Route).Status)` to compare in string type.

starsz:
> This pr can work fine with the #1205, but the Unit tests always fail, and logically, we could not find anything wrong.
> 
> Please take a look when you are free @nic-chen @tokers @starsz @imjoey , many thanks~~

Hi, I think it was updated by the Update test.

See: 
https://github.com/apache/apisix-dashboard/blob/5a007c759e40854ec6a75ad64178eae9081356ba/api/internal/handler/route/route_test.go#L409-L422

nic-chen:
yes, it has been updated. you should set `status` to 1 in close-in test case.

Jaycean:
@starsz @nic-chen Thank you for your help.

api/internal/handler/route/route_test.go
```diff
@@ -925,6 +927,42 @@ func TestRoute(t *testing.T) {
...
+	err = json.Unmarshal([]byte(reqBody), listInput)
+	assert.Nil(t, err)
+	ctx.SetInput(listInput)
+	retPage, err = handler.List(ctx)
+	assert.Nil(t, err)
+	dataPage = retPage.(*store.ListOutput)
+	assert.Equal(t, len(dataPage.Rows), 1)
+
+        //sleep
+        time.Sleep(time.Duration(100) * time.Millisecond)
```
starsz, line 942: Style: redundant space.

api/internal/handler/route/route.go
```diff
@@ -219,6 +220,10 @@ func (h *Handler) List(c droplet.Context) (interface{}, error) {
 				return false
 			}
 
+			if input.Status != "" && fmt.Sprintf("%d", obj.(*entity.Route).Status) != input.Status {
```
starsz, line 223: I recommend using `strconv.Itoa` instead of `fmt.Sprintf`. Because the former is faster than the latter in the benchmark test.

Jaycean: Thank you for your suggestion. I'll revise it right away

api/internal/handler/route/route_test.go
starsz: Hi, @Jaycean don't forget to fix the style problem.

juzhiyuan:
image

CI pending?

Jaycean:
> image
> 
> CI pending?

It should be. I'll run again

api/internal/handler/route/route_test.go
Jaycean: done.

liuxiran (approved):
thanks for all helps, test cases run successfully

starsz (approved):

juzhiyuan (approved):

imjoey:
@Jaycean @juzhiyuan @liuxiran @nic-chen @starsz  Hi, thanks for the great work.

I'm writing here just for your information, I've committed a PR #1224 to backport this bug fix into v2.3 branch for the v2.3 release.

@juzhiyuan shall we also create two new labels?
- `need_backport` to mark a PR that needs backport to other branches
- `backport` to mark a PR is only for backport

Any feedback is welcome. Thanks.

PR #1226, opened 2021-01-07
feat(release): Enable CI trigger for v2.3 branch

imjoey:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance

___

As the preparation of v2.3 release, we need to continuously backport hot bugfixs to `v2.3` branch, so enable CI process at v2.3 branch would help us keep the release more stable. 

Since v2.3 will not be a LTS version, which means there will no activities in v2.3 branch after release v2.3, so this PR is going to use the single `v2.3` for all branch configurations.

### DIFF
--- a/.github/workflows/backend-cli-test.yml
+++ b/.github/workflows/backend-cli-test.yml
@@ -3,10 +3,10 @@ name: Backend CLI Test
 on:
   push:
     branches:
-      - master
+      - v2.3
   pull_request:
     branches:
-      - master
+      - v2.3
 
 jobs:
   run-test:
--- a/.github/workflows/backend-e2e-test.yml
+++ b/.github/workflows/backend-e2e-test.yml
@@ -3,10 +3,10 @@ name: Backend E2E Test
 on:
   push:
     branches:
-      - master
+      - v2.3
   pull_request:
     branches:
-      - master
+      - v2.3
 
 jobs:
   run-test:
--- a/.github/workflows/backend-unit-test.yml
+++ b/.github/workflows/backend-unit-test.yml
@@ -3,10 +3,10 @@ name: Backend Unit Test
 on:
   push:
     branches:
-      - master
+      - v2.3
   pull_request:
     branches:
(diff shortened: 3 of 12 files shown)


.github/workflows/backend-cli-test.yml
```diff
@@ -3,10 +3,10 @@ name: Backend CLI Test
 on:
   push:
     branches:
-      - master
+      - v2.3
```
juzhiyuan, line 6: 2.x is acceptable :)

juzhiyuan: both the master and v2.x need to be checked IMO

juzhiyuan (changes requested):

.github/workflows/backend-cli-test.yml
imjoey: Thanks for reviewing. IMHO, CI for `v2.3` will no longer exist after version `2.3` is released, cause it is not a LTS version. And this change will also not be backported into master branch.  For example, for the release version `2.4`, we can simply use `v2.4` here. Meanwhile, the `master` is the only branch name needed to listed here in the `master` branch.

Once we have LTS versions, then we can keep `master` and all LTS version branches in the `master` branch. So What do you think? 😄 Much appreciated.

juzhiyuan: got it, let's merge

juzhiyuan (approved):

.github/workflows/backend-cli-test.yml
imjoey: > both the master and v2.x need to be checked IMO

 The codes in master branch will also be checked if we commit a PR targeted to `master` branch. So we can place different branch names here, respectively. Please correct me if I was wrong. Thanks.

imjoey:
@juzhiyuan thank you very much for reviewing. Very glad that we can come to an agreement. Much appreciated for the openness. 👍

Issue #825, opened 2020-11-18
[Proposal]: Import & Export route from OpenAPI Specification3.0


liuxiran:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Feature or performance improvement

Refer to https://github.com/apache/apisix-dashboard/issues/611 
___
### Why

Sometimes user's apis have already defined based on the openapi specification and when they want to publish those apis on APISIX, they need to learn the concepts of our products,  such as how to do parameter validation, how to do authentication, how to do path matching and so on. Actually, many users don't care how to publish a route, so If we can support importing the configuration of openapi to generate the route, that would reduce the user's use cost. 

### How

- add buttons for import and export in route page
   import button for upload the profile of route based on the openapi 3.0, and the export button for download it.

- manage-api parse the file, generate an apisix-compliant data structure, and write to the etcd

LiteSun:
I think this feature will save a lot of time for users who need to import a lot of APIs. 

According to the above description, the workload on the front-end is relatively small, and the challenge is to convert the profile into an apisix-compliant data structure.

let's cc @nic-chen @gxthrj for more suggestions.

nic-chen:
@liuxiran @LiteSun 

I think we need a detailed example of the imported and exported data so that everyone can understand the details more easily

gxthrj:
OpenAPI is a description document used to provide quick reading and testing. The route in APISIX is also used to manage API, there are still some differences. For example, we need to consider how to convert the plugin logic in route.

We'd better limit the import / export responsibility boundary and provide a more detailed transformation example.

liuxiran:
> OpenAPI is a description document used to provide quick reading and testing. The route in APISIX is also used to manage API, there are still some differences. For example, we need to consider how to convert the plugin logic in route.

Yep, OpenApi is mainly  defines some of the contents of the api itself, including: `name`, `description`, `service-side address`, `method`, `path`, `request params`, `response params`, `authentication and authorization` , etc, example configurations can refer to [1].

In addition to the basic content of the api mentioned above, `Route` defined in APISIX also contains powerful extension properties, these need to be done by configuring different plugins. The configuration of these plugins are not currently supported by OpenApi except for Authentication plugins. 


> We'd better limit the import / export responsibility boundary and provide a more detailed transformation example.

The limit I think would be as follows:

[...]

PR #1245, opened 2021-01-08
feat(BE): Export route from OpenAPI Specification3.0

Jaycean:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#825 
___
- Export route from OpenAPI Specification3.0


api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,349 @@
...
+				//}
+
+				// analysis request-validation plugin
+				if key == "request-validation" {
+					for k, v := range value.(map[string]interface{}) {
+						if k == "header_schema" {
+							param.In = "header"
+							requestValidation := &entity.RequestValidation{}
+							reqBytes, _ := json.Marshal(&v)
+							err := json.Unmarshal(reqBytes, requestValidation)
```
imjoey, line 167: @Jaycean Additional check is also required here.

```suggestion
							err := json.Unmarshal(reqBytes, requestValidation)
							if err != nil {
							        // log error and return back to client
							}
```

imjoey (changes requested):

api/internal/handler/route_export/route_export.go
Jaycean: Yes, I just added it

Jaycean: Thks for checking.

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,357 @@
...
+								match := compile.Match([]byte(va.([]interface{})[0].(string)))
+								if match {
+									m[va.([]interface{})[2].(string)] = &openapi3.MediaType{Schema: &openapi3.SchemaRef{Value: schema}}
+									requestBody.Content = m
+								}
+							}
+						}
+					}
+					continue
+				}
```
imjoey, line 205: @Jaycean If only caring about the keys that equals to  `body_schema` and `header_schema`, IMHO, there is no need to iterate the entire map. Just use 

```go
if valueMap, ok := value.(map[string]interface{}); ok {
    if hsVal, ok := valueMap["header_schema"]; ok {
        // Deal with header schema
    }

    if bsVal, ok := valueMap["body_schema"]; ok {
        // Deal with body schema
    }
}

```

BTW: It would be better to check every type assertion in Golang.

Jaycean: Thks, I'll check and change the code right away.

starsz:
Hi, @Jaycean. Need unit test.

Jaycean:
> Hi, @Jaycean. Need unit test.

At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.

starsz:
> > Hi, @Jaycean. Need unit test.
> 
> At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.

Not agree with you.
First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
For example, in this case.
We need to test these situations at least:
1. route can't found
2. Uris contain "*"
3. plugins with `request-validation`
4. different route method
...

> and if the relevant route data is not created, the interface will not work normally use.

We can mock data in etcd.
See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.

Jaycean:
> > > Hi, @Jaycean. Need unit test.
> > 
> > 
> > At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.
> 
> Not agree with you.
> First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
> For example, in this case.
> We need to test these situations at least:
> 
> 1. route can't found
> 2. Uris contain "*"
> 3. plugins with `request-validation`
> 4. different route method
>    ...
> 
> > and if the relevant route data is not created, the interface will not work normally use.
> 
> We can mock data in etcd.
> See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.

Thank you for your advice. I also think unit testing is really necessary. I just thought that this interface test is almost the same as E2E, but unit testing should cover all situations as much as possible

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,358 @@
...
+	components := &openapi3.Components{}
+	secSchemas := openapi3.SecuritySchemes{}
+	for _, route := range routes {
+
+		extensions := make(map[string]interface{})
+		path.Summary = route.Desc
+		path.OperationID = route.Name
+		if route.URI != "" {
+			extensions["x-apisix-uri"] = route.URI
+		}
```
liuxiran, line 98: `route.URI` need to be parsed into `paths[uri]`

Jaycean: Thks for review.
Just now, a bug was detected, which did not parse the URI in the create route, but only parsed the URI. I will deal with this problem later. At the same time, x-apisix-uri really does not need to be parsed

api/test/e2e/route_export_test.go
```diff
@@ -0,0 +1,192 @@
...
+		},
```
starsz, line 151: This line is too long. 
I think we can you JSON.Marshal to generate the expectBody.

Like this

```
ecpectR1, _ := json.Marshal("....“）
expectR2, _:= json.Marshal("....")
...

ExpectBody: []string{string(expectR1), string(expectR2)},
```

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,365 @@
...
+	input := c.Input().(*ExportInput)
+	ids := strings.Split(input.IDs, ",")
+	routes := []*entity.Route{}
+	for _, id := range ids {
+		route, err := h.routeStore.Get(id)
+		if err != nil {
+			return nil, err
+		}
+		routes = append(routes, route.(*entity.Route))
+
```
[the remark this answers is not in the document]
Jaycean: done.

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,365 @@
...
+func routeToOpenApi3(routes []*entity.Route) *openapi3.Swagger {
+	paths := openapi3.Paths{}
+	pathItem := &openapi3.PathItem{}
+	path := &openapi3.Operation{}
+	paramsRefs := []*openapi3.ParameterRef{}
+	requestBody := &openapi3.RequestBody{}
+	components := &openapi3.Components{}
+	secSchemas := openapi3.SecuritySchemes{}
+	for _, route := range routes {
+
```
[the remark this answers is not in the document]
Jaycean: done.

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,365 @@
...
+		extensions["x-apisix-priority"] = route.Priority
+		extensions["x-apisix-status"] = route.Status
+		extensions["x-apisix-enableWebsocket"] = route.EnableWebsocket
+		path.Extensions = extensions
+		path.Parameters = paramsRefs
+		path.RequestBody = &openapi3.RequestBodyRef{Value: requestBody}
+		path.Responses = openapi3.NewResponses()
+		for i := range route.Methods {
+			switch strings.ToUpper(route.Methods[i]) {
```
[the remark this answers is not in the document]
Jaycean: The unified tab indentation used here may be that GitHub web pages look a lot indented

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,365 @@
...
+					Description:    path.Description,
+					OperationID:    path.OperationID + "Head",
+					Parameters:     path.Parameters,
+					RequestBody:    path.RequestBody,
+					Responses:      path.Responses,
+					Callbacks:      path.Callbacks,
+					Deprecated:     path.Deprecated,
+					Security:       path.Security,
+					Servers:        path.Servers,
+					ExternalDocs:   path.ExternalDocs,
```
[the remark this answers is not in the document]
Jaycean: done.

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	components.SecuritySchemes = secSchemas
+	swagger := openapi3.Swagger{
+		OpenAPI:    openApi,
+		Info:       &openapi3.Info{Title: title, Version: openApi},
+		Paths:      paths,
+		Components: *components,
+	}
+	return &swagger
+}
```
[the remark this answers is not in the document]
Jaycean: It's really long at the moment. I'll try to take it apart

api/test/e2e/route_export_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	for _, tc := range tests2 {
+		testCaseCheck(tc, t)
+	}
+
+	tests3 := []HttpTestCase{
+
```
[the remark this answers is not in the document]
Jaycean: done.

api/test/e2e/route_export_test.go
Jaycean: In this case, because it is to verify the exported data, in order to make the data cover as wide as possible, the resulting data is very large. I'll take a look at splitting them, so the readability is really better

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		if route.FilterFunc != "" {
+			extensions["x-apisix-filterFunc"] = route.FilterFunc
+		}
+		if route.ServiceID != nil {
+			extensions["x-apisix-serviceID"] = route.ServiceID
+		}
+		if route.UpstreamID != nil {
+			extensions["x-apisix-upstreamID"] = route.UpstreamID
+		}
```
liuxiran, line 120: when export a route with `upstreamId`, we have to  export the `uptream` which contained in the `upstreamid`
when export a route with `serviceid`, we have to export the `upstream` and the `plugins` which contained in the `serviceid`

reference: https://github.com/apache/apisix-dashboard/issues/825#issuecomment-748009952

Jaycean: There is a problem here. I'll modify the code and reprocess the data of upstream and plugin.

api/internal/handler/route_export/route_export.go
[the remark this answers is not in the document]
starsz: It's still too long.IMO, we can deal with `route.plugin` in another function.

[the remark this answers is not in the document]
Jaycean: Yes, new parsing content has been added today, and there is no time to separate functions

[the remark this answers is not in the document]
Jaycean: Tomorrow I will split it into several methods

api/internal/handler/route_export/route_export.go
```diff
@@ -0,0 +1,432 @@
...
+
+var (
+	openApi = "3.0.0"
+	title   = "RoutesExport"
+	service interface{}
+	err     error
+)
+
+func (h *Handler) routeToOpenApi3(routes []*entity.Route) (*openapi3.Swagger, error) {
+	paths := openapi3.Paths{}
```
tokers, line 98: It would be better to put these variable initializations inside a var block.

Jaycean: Basicauth, keyauth and jwtauth are three kinds of authentication methods. I don't think they should be changed in the program, so I use constants

api/test/e2e/route_export_test.go
```diff
@@ -0,0 +1,716 @@
...
+			Desc:         "hit route that not exist",
+			Object:       APISIXExpect(t),
+			Method:       http.MethodGet,
+			Path:         "/hello_",
+			Headers:      map[string]string{"Host": "foo.com"},
+			ExpectStatus: http.StatusNotFound,
+			ExpectBody:   "{\"error_msg\":\"404 Route Not Found\"}\n",
+		},
+		{
+			Desc:   "create route",
```
liuxiran, line 108: it would be better to change the test case desc
```suggestion
			Desc:   "create route with uris and hosts",
```

api/test/e2e/route_export_test.go
```diff
@@ -0,0 +1,716 @@
...
+			Desc:         "hit route2 that not exist",
+			Object:       APISIXExpect(t),
+			Method:       http.MethodGet,
+			Path:         "/hello2",
+			Headers:      map[string]string{"Host": "foo.com"},
+			ExpectStatus: http.StatusNotFound,
+			ExpectBody:   "{\"error_msg\":\"404 Route Not Found\"}\n",
+		},
+		{
+			Desc:   "create route2",
```
liuxiran, line 237:
```suggestion
			Desc:   "create route2 with uri and hosts to test whether the uri parsing is correct ",
```

api/test/e2e/route_export_test.go
```diff
@@ -0,0 +1,716 @@
...
+			Desc:       "get the service s2",
+			Object:     ManagerApiExpect(t),
+			Method:     http.MethodGet,
+			Path:       "/apisix/admin/services/s2",
+			Headers:    map[string]string{"Authorization": token},
+			ExpectCode: http.StatusOK,
+			ExpectBody: serviceStrS2,
+		},
+		{
+			Desc:   "create route4 using the service just created",
```
liuxiran, line 643: you'd better to update the desc too

liuxiran:
it would be better to add two more test cases:

- export a route created with upstream id
- export a route created with service which is created with upstream id

Jaycean:
> it would be better to add three more test cases:
> 
> * export a route created with upstream id
> * export a route created with service which is created with upstream id
> * export a route created with `basic-auth` plugin enabled



> it would be better to add three more test cases:
> 
> * export a route created with upstream id
> * export a route created with service which is created with upstream id
> * export a route created with `basic-auth` plugin enabled

I have added test cases for these three kinds of exports. Auth authentication has added the route export test cases opened by JWT plugin, and the following basic auth and API key will also be added.

api/test/e2e/route_export_test.go
Jaycean: done.

api/test/e2e/route_export_test.go
Jaycean: done.

api/test/e2e/route_export_test.go
Jaycean: done.

Jaycean:
> > > Hi, @Jaycean. Need unit test.
> > 
> > 
> > At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.
> 
> Not agree with you.
> First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
> For example, in this case.
> We need to test these situations at least:
> 
> 1. route can't found
> 2. Uris contain "*"
> 3. plugins with `request-validation`
> 4. different route method
>    ...
> 
> > and if the relevant route data is not created, the interface will not work normally use.
> 
> We can mock data in etcd.
> See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.

Now I have encountered a problem in unit testing, exporting the data that needs to create route, upstream and service. After reading the example you gave me, I still don't know how to write it. Can you give me some advice.

Jaycean:
> > > > Hi, @Jaycean. Need unit test.
> > > 
> > > 
> > > At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.
> > 
> > 
> > Not agree with you.
> > First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
> > For example, in this case.
> > We need to test these situations at least:
> > 
> > 1. route can't found
> > 2. Uris contain "*"
> > 3. plugins with `request-validation`
> > 4. different route method
> >    ...
> > 
> > > and if the relevant route data is not created, the interface will not work normally use.
> > 
> > 
> > We can mock data in etcd.
> > See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.
> 
> @starsz Now I have encountered a problem in unit testing, exporting the data that needs to create route, upstream and service. After reading the example you gave me, I still don't know how to write it. Can you give me some advice. Thks

starsz:
> > > > > Hi, @Jaycean. Need unit test.
> > > > 
> > > > 
> > > > At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.
> > > 
> > > 
> > > Not agree with you.
> > > First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
> > > For example, in this case.
> > > We need to test these situations at least:
> > > 
> > > 1. route can't found
> > > 2. Uris contain "*"
> > > 3. plugins with `request-validation`
> > > 4. different route method
> > >    ...
> > > 
> > > > and if the relevant route data is not created, the interface will not work normally use.
> > > 
> > > 
> > > We can mock data in etcd.
> > > See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.
> > 
> > 
> > @starsz Now I have encountered a problem in unit testing, exporting the data that needs to create route, upstream and service. After reading the example you gave me, I still don't know how to write it. Can you give me some advice. Thks

OK. Sorry, we don't have a doc to introduce the test contribute doc yet.

First, you should know the mock technology. 
You can reference the doc: https://github.com/stretchr/testify#mock-package.
Mock provides a way to check input and give the return value that we want.

In your code, we should mock those function:
```
routeStore.Get
routeStore.List
upstreamStore.Get
serviceStore.Get
```

Second, IMO, you should write two testcases
One test case for `ExportRoutes`, like:
`TestDataLoader_ExportRoutes` 
And another test case for `ExportAllRoutes`.like 
`TestDataLoader_ExportAllRoutes`

Third, in those test case function.We should mock the store like:

```
routeMockStore := &store.MockInterface{}
routeMockStore.On("Get", mock.Anything, mock.Anything).Run(func(args mock.Arguments) {
         id :=  args.Get(0).(string)
        // here we can do some check on the id.				
}).Return(tc.giveRet, tc.giveErr)
```
Mock.Anything stands for the args when you called routeMockStore.Get.
tc.giveRet and tc.giveErr means what you want to get when called this line: 
https://github.com/apache/apisix-dashboard/pull/1245/files#diff-195642dee38f4cdcc68921704b4b59dce5afc67d06074c1bc53004eb9b5dfc5eR70
The same as other mock functions.

Any feedback is welcome.

liuxiran:
checklist:

- [ ] export route with host
- [ ] export route with hosts
- [ ] export route with uri
- [ ] export route with uris
- [ ] export route with upstream
- [ ] export route with upstream_id
- [ ] export route with service_id(service create with upstream)
- [ ] export route with service_id(service create with upstream_id)
- [ ] export route with labels and service_id(service create with labels)
- [ ] export route without labels and service_id(service create with labels)
- [ ] export route with plugins and service_id(service create with plugins which is not the same as route's plugins)
- [ ] export route with basic-auth plugin
- [ ] export route with jwt-auth plugin
- [ ] export route with key-auth plugin
- [ ] export route with request-validation (body_schema and header_schema) plugin

liuxiran:
Test cases are good for me

Jaycean:
> > > > > > Hi, @Jaycean. Need unit test.
> > > > > 
> > > > > 
> > > > > At present, the function interface needs to export the existing route data and ensure the correctness of the exported data, so I think E2E should be written more perfectly. I think if the simple unit test only tests whether the interface can export data, it does not represent whether the function of the interface is correct, and if the relevant route data is not created, the interface will not work normally use.
> > > > 
> > > > 
> > > > Not agree with you.
> > > > First of all, unit testing is very important. It helps find bugs easily and quickly if someone changes the code.And unit testing can cover more code than e2e test.
> > > > For example, in this case.
> > > > We need to test these situations at least:
> > > > 
> > > > 1. route can't found
> > > > 2. Uris contain "*"
> > > > 3. plugins with `request-validation`
> > > > 4. different route method
> > > >    ...
> > > > 
> > > > > and if the relevant route data is not created, the interface will not work normally use.
> > > > 
> > > > 
> > > > We can mock data in etcd.
> > > > See https://github.com/apache/apisix-dashboard/blob/master/api/internal/handler/server_info/server_info_test.go for more details.
> > > 
> > > 
> > > @starsz Now I have encountered a problem in unit testing, exporting the data that needs to create route, upstream and service. After reading the example you gave me, I still don't know how to write it. Can you give me some advice. Thks
> 
> OK. Sorry, we don't have a doc to introduce the test contribute doc yet.
> 
> First, you should know the mock technology.
> You can reference the doc: https://github.com/stretchr/testify#mock-package.
> Mock provides a way to check input and give the return value that we want.
> 
> In your code, we should mock those function:
> 
> ```
> routeStore.Get
> routeStore.List
> upstreamStore.Get
> serviceStore.Get
> ```
> 
> Second, IMO, you should write two testcases
> One test case for `ExportRoutes`, like:
> `TestDataLoader_ExportRoutes`
> And another test case for `ExportAllRoutes`.like
> `TestDataLoader_ExportAllRoutes`
> 
> Third, in those test case function.We should mock the store like:
> 
> ```
> routeMockStore := &store.MockInterface{}
> routeMockStore.On("Get", mock.Anything, mock.Anything).Run(func(args mock.Arguments) {
>          id :=  args.Get(0).(string)
>         // here we can do some check on the id.				
> }).Return(tc.giveRet, tc.giveErr)
> ```
> 
> Mock.Anything stands for the args when you called routeMockStore.Get.
> tc.giveRet and tc.giveErr means what you want to get when called this line:
> https://github.com/apache/apisix-dashboard/pull/1245/files#diff-195642dee38f4cdcc68921704b4b59dce5afc67d06074c1bc53004eb9b5dfc5eR70
> The same as other mock functions.
> 
> Any feedback is welcome.

Thank you very much. I'll try to write it.

membphis:
@Jaycean please merge the master branch, FE e2e test cases have been fixed

Jaycean:
@liuxiran @nic-chen
The unit test and test cases have been supplemented. In the process of testing, I found a problem：
- At present, the path data structure of OpenAPI 3 is used to store URIs, and the data structure is map [string]. If the URI is repeated, the data will be covered, if the repeated URIs are directly covered, only one URI can be exported, which will lead to the loss of user data.
- My idea is to add a uniform suffix after the same URI when exporting, so that the import can recognize the URI
For example：
```
{
	"components": {},
	"info": {
		"title": "RoutesExport",
		"version": "3.0.0"
	},
	"openapi": "3.0.0",
	"paths": {
		"/hello": {
			"get": {}
		}
		"/hello-repeaturi2": {
			"get": {}
		}
		"/hello-repeaturi2": {
			"get": {}
		}
	}
}
```
- Do you have any better suggestions? We can discuss them. Thks.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+	IDs string `auto_read:"ids,path"`
+}
+
+func (h *Handler) ExportRoutes(c droplet.Context) (interface{}, error) {
+	input := c.Input().(*ExportInput)
+	ids := strings.Split(input.IDs, ",")
+	routes := []*entity.Route{}
+
+	for _, id := range ids {
+		route, err := h.routeStore.Get(context.Background(), id)
```
starsz, line 72: Hi, I think we need to avoid use context.Background here. We should use c.Context().

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+	for _, route := range routes {
+		extensions := make(map[string]interface{})
+		pathItem := &openapi3.PathItem{}
+		path := openapi3.Operation{}
+		path.Summary = route.Desc
+		path.OperationID = route.Name
+
+		if route.ServiceID != nil {
+			serviceID := utils.InterfaceToString(route.ServiceID)
+			service, err = h.serviceStore.Get(context.Background(), serviceID)
```
starsz, line 141: Ditto.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+			_service := service.(*entity.Service)
+			servicePlugins = _service.Plugins
+			serviceLabels = _service.Labels
+		}
+
+		if route.Upstream != nil {
+			extensions["x-apisix-upstream"] = route.Upstream
+		} else if route.UpstreamID != nil && route.Upstream == nil {
+			upstreamID := utils.InterfaceToString(route.UpstreamID)
+			upstream, err := h.upstreamStore.Get(context.Background(), upstreamID)
```
starsz, line 158: Ditto.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+				return nil, err
+			}
+			extensions["x-apisix-upstream"] = upstream
+		} else if route.UpstreamID == nil && route.Upstream == nil && route.ServiceID != nil {
+			_service := service.(*entity.Service)
+			if _service.Upstream != nil {
+				extensions["x-apisix-upstream"] = _service.Upstream
+			} else if _service.Upstream == nil && _service.UpstreamID != nil {
+				upstreamID := utils.InterfaceToString(_service.UpstreamID)
+				upstream, err := h.upstreamStore.Get(context.Background(), upstreamID)
```
starsz, line 172: Ditto.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+	swagger := openapi3.Swagger{
+		OpenAPI:    openApi,
+		Info:       &openapi3.Info{Title: title, Version: openApi},
+		Paths:      paths,
+		Components: *components,
+	}
+	return &swagger, nil
+}
+
+func parseLabels(route *entity.Route, serviceLabels map[string]string, labels map[string]string) (map[string]string, error) {
```
starsz, line 303: What's the label args use for?

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+			extensions["x-apisix-filter_func"] = route.FilterFunc
+		}
+
+		if route.Script != nil {
+			extensions["x-apisix-script"] = route.Script
+		}
+
+		if route.ServiceProtocol != "" {
+			extensions["x-apisix-service_protocol"] = route.ServiceProtocol
+		}
```
starsz, line 220: style:  It's better to have a empty line here. Just keep the same style.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,441 @@
...
+					Value: &openapi3.Parameter{
+						In:          "path",
+						Name:        "params",
+						Required:    true,
+						Description: "params in path",
+						Schema:      &openapi3.SchemaRef{Value: &openapi3.Schema{Type: "string"}}}})
+			} else {
+				paths[uri] = pathItem
+			}
+		}
```
starsz, line 249: Can we split `analysis  route.URIs` to another function?

starsz (changes requested):

liuxiran:
> @liuxiran @nic-chen
> The unit test and test cases have been supplemented. In the process of testing, I found a problem：
> 
> * At present, the path data structure of OpenAPI 3 is used to store URIs, and the data structure is map [string]. If the URI is repeated, the data will be covered, if the repeated URIs are directly covered, only one URI can be exported, which will lead to the loss of user data.
> * My idea is to add a uniform suffix after the same URI when exporting, so that the import can recognize the URI
>   For example：
> 
> ```
> {
> 	"components": {},
> 	"info": {
> 		"title": "RoutesExport",
> 		"version": "3.0.0"
> 	},
> 	"openapi": "3.0.0",
> 	"paths": {
> 		"/hello": {
> 			"get": {}
> 		}
> 		"/hello-repeaturi1": {
> 			"get": {}
> 		}
> 		"/hello-repeaturi2": {
> 			"get": {}
> 		}
> 	}
> }
> ```
> 
> * My method also has the risk of repeating the user named URI
> * Do you have any better suggestions? We can discuss them. Thks.

>OpenAPI defines a unique operation as a combination of a path and an HTTP method. This means that two GET or two POST methods for the same path are not allowed – even if they have different parameters (parameters have no effect on uniqueness)
refer from https://swagger.io/docs/specification/paths-and-operations/

 After communicating with @Jaycean and @nic-chen there are three alternative solutions:

- A: add a uniform suffix after the same URI when exporting, and add a description to the user guide.
This solution can meet the platform-wide data is not lost, the disadvantage is that when users need to use the swagger file to generate documents or sdk, they  need to modified manually.

- B: return error when exporting routes with same URI
In this way exported OAS3.0 file  can be used directly, with the disadvantage that there may be data loss for APISIX. The remedy is that we provide additional import and export of platform-wide data, in accordance with the apisix data format.


which one do you perfer?  any suggestions are welcome, thanks a lot

@Jaycean @nic-chen @membphis @starsz @juzhiyuan @imjoey

api/internal/handler/data_loader/route_export.go
Jaycean: no problem.

api/internal/handler/data_loader/route_export.go
Jaycean: When service and route have labels at the same time, use route's label. When route has no label, service sometimes uses service's label. This function is used to process this logic

api/internal/handler/data_loader/route_export.go
Jaycean: No problem, but there are some problems that need to be solved by voting. I'll deal with them after the results come out

juzhiyuan:
because it's using `URI` as paths's `Key`, then data gets covered would happen. I would prefer adding a global prefix to avoid this happen.

nic-chen:
I vote for option A, at least it can be imported and exported normally on APISIX platform.

juzhiyuan:
not sure how other platforms would resolve those issues, we may have a search.

Jaycean:
In my opinion, from the perspective of apimix platform, I choose scheme a to satisfy the users of apimix platform. The function that URI can be renamed is more open and friendly for users.

liuxiran:
> not sure how other platforms would resolve those issues, we should have a search.

I tried Aliyun API Gateway and eolinker( a API management platform):

- In Aliyun API Gateway, when we export  multiple APIs, it will split each API into a single file(It's also a good choice)
- In eolinker, it does not support export OpenAPI.

As most people choose option A, let's solve the problem with A first, after this feature megerd and released, let's wait for more users feedback ^_^

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
starsz: No, I didn't see the lables args was used.

Jaycean: I don't quite understand your question. Do you mean you don't see users or business scenarios using this field? I don't know if my understanding is correct, because more coverage possibilities are considered everywhere to ensure that the data is not lost. After discussing with nic-chen, this processing function is added.

api/internal/handler/data_loader/route_export.go
liuxiran: it seems that this case has been fixed, right @Jaycean ?

Jaycean: done

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+		upstreamStore: store.GetStore(store.HubKeyUpstream),
+		serviceStore:  store.GetStore(store.HubKeyService),
+		consumerStore: store.GetStore(store.HubKeyConsumer),
+	}, nil
+}
+
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.POST("/apisix/admin/routes/export/:ids", wgin.Wraps(h.ExportRoutes,
+		wrapper.InputType(reflect.TypeOf(ExportInput{}))))
+	r.GET("/apisix/admin/exportall/routes", wgin.Wraps(h.ExportAllRoutes))
```
starsz, line 60: Question:
1. Why the first API's method is `POST`
2. "/apisix/admin/exportall/routes" is not very good. cc @liuxiran

liuxiran: > Question:
> 
> 1. Why the first API's method is `POST`
> 2. "/apisix/admin/exportall/routes" is not very good. cc @liuxiran

1. `GET` would be better
2. In fact, we didn't find a very satisfactory name either, do you have any suggestions? @starsz

starsz: GET `/apisix/admin/export/routes/:ids`
GET `/apisix/admin/export/routes`

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+//ExportRoutes Export data by passing route ID, such as "R1" or multiple route parameters, such as "R1,R2"
+func (h *Handler) ExportRoutes(c droplet.Context) (interface{}, error) {
+	input := c.Input().(*ExportInput)
+	ids := strings.Split(input.IDs, ",")
+	routes := []*entity.Route{}
+
+	for _, id := range ids {
+		route, err := h.routeStore.Get(c.Context(), id)
+		if err != nil {
+			return nil, err
```
starsz, line 76: Need to return special info if err == data.ErrNotFound.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+func (h *Handler) RouteToOpenAPI3(c droplet.Context, routes []*entity.Route) (*openapi3.Swagger, error) {
+	paths := openapi3.Paths{}
+	paramsRefs := []*openapi3.ParameterRef{}
+	requestBody := &openapi3.RequestBody{}
+	components := &openapi3.Components{}
+	secSchemas := openapi3.SecuritySchemes{}
+	servicePlugins := make(map[string]interface{})
+	plugins := make(map[string]interface{})
+	serviceLabels := make(map[string]string)
+	labels := make(map[string]string)
```
starsz, line 134: Those variables can put in the cycle.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+		Info:       &openapi3.Info{Title: title, Version: openApi},
+		Paths:      paths,
+		Components: *components,
+	}
+	return &swagger, nil
+}
+
+//ParseLabels When service and route have labels at the same time, use route's label.
+//When route has no label, service sometimes uses service's label. This function is used to process this logic
+func ParseLabels(route *entity.Route, serviceLabels map[string]string, labels map[string]string) (map[string]string, error) {
```
starsz, line 264: labels args is not used.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+					if bsVal, ok := valueMap["body_schema"]; ok {
+						m := map[string]*openapi3.MediaType{}
+						reqBytes, _ := json.Marshal(&bsVal)
+						schema := &openapi3.Schema{}
+						err := json.Unmarshal(reqBytes, schema)
+						if err != nil {
+							log.Errorf("json marshal failed: %s", err)
+						}
+
+						m["*/*"] = &openapi3.MediaType{Schema: &openapi3.SchemaRef{Value: schema}}
```
starsz, line 344: Need some comments.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+	if route.Plugins != nil {
+		param := &openapi3.Parameter{}
+		secReq := &openapi3.SecurityRequirements{}
+
+		// analysis plugins
+		for key, value := range route.Plugins {
+			// analysis proxy-rewrite plugin
+			//if key == "proxy-rewrite" {
+			//	continue
+			//}
```
starsz, line 308: Those lines can be deleted.

api/internal/handler/data_loader/route_export.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	for _, uri := range routeURIs {
+		_time := time.Now().UnixNano() / 1e6
+		if strings.Contains(uri, "*") {
+			if _, ok := paths[strings.Split(uri, "*")[0]+"{params}"]; !ok {
+				paths[strings.Split(uri, "*")[0]+"{params}"] = pathItem
+			} else {
+				paths[strings.Split(uri, "*")[0]+"{params}"+"-APISIX-REPEAT-URI-"+strconv.FormatInt(_time, 10)] = pathItem
+			}
```
starsz, line 423: This can be done in closure.

api/internal/handler/data_loader/route_export_test.go
```diff
@@ -0,0 +1,1864 @@
...
+
+	h := Handler{routeStore: mStore}
+	ctx := droplet.NewContext()
+	ctx.SetInput(input)
+
+	ret, err := h.ExportAllRoutes(ctx)
+	assert.Nil(t, err)
+	ret1, err := json.Marshal(ret)
+	if err != nil {
+	}
```
starsz, line 1851: `assert.Nil(t, err)` or `_ `

starsz (changes requested):

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done.

api/internal/handler/data_loader/route_export.go
Jaycean: done

api/internal/handler/data_loader/route_export.go
Jaycean: done

api/internal/handler/data_loader/route_export.go
Jaycean: done

api/internal/handler/data_loader/route_export.go
Jaycean: Here I add a global variable, which is initialized to 0 every time I call the interface. In the loop, if it meets the conditions, it will automatically increase to 1. I don't know if this is OK

api/internal/handler/data_loader/route_export_test.go
Jaycean: done

api/internal/handler/data_loader/route_export.go
Jaycean: done

Jaycean:
@starsz @nic-chen   PTAL. Thks.

starsz (approved):

liuxiran (approved):
this pr has covers all the requirements according to #825

imjoey (approved):

membphis:
wow, Congratulations this PR was merged

PR #1258, opened 2021-01-11
docs: added back docs for release 2.3

juzhiyuan:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

sync master with v2.3

PR #1259, opened 2021-01-11
manger api with front web 

Yangzcy423911:
The back-end and front-end services are deployed separately. There should be more detailed documents to explain. The user experience of the front-end document is better, and the back-end go should strengthen the description of the relevant documents, so as to facilitate users to use better, so as to reduce the service coupling and improve the availability.


juzhiyuan:
emmm no, duplicate with https://github.com/apache/apisix-dashboard/pull/1258

Issue #1038, opened 2020-12-12
Client-side URL redirect


juzhiyuan:
# Bug report

## Describe the bug

https://github.com/apache/apisix-dashboard/security/code-scanning/3?query=ref%3Arefs%2Fheads%2Fmaster

![image](https://user-images.githubusercontent.com/2106987/101979572-0ebf1a00-3c99-11eb-841f-d739b0e00752.png)

Redirecting to a URL that is constructed from parts of the DOM that may be controlled by an attacker can facilitate phishing attacks. In these attacks, unsuspecting users can be redirected to a malicious site that looks very similar to the real site they intend to visit, but which is controlled by the attacker.

juzhiyuan:
We may limit the target redirect URL to have the same host as the current URL.

ThanoshanMV:
Hi @juzhiyuan, I would like to work on this issue. Can you please guide me here?

juzhiyuan:
@ThanoshanMV sure! please give me some time to prepare this issue.

BTW, here[1] is our Slack channel, welcome to join in :)

[1] http://s.apache.org/slack-invite

juzhiyuan:
![image](https://user-images.githubusercontent.com/2106987/104311758-a3bf6800-5510-11eb-8291-45765ac335d3.png)


Hi @ThanoshanMV, here is the explanation from GitHub, we should limit the `redirect` URL, it MUST do the redirect action with the current host.

ThanoshanMV:
Am I okay here: `decodeURIComponent(redirect)`'s host should be the current host. If it is, then we'll perform the redirect action and if it doesn't, we don't perform any redirect action.

juzhiyuan:
@ThanoshanMV yep :)

PR #1284, opened 2021-01-13
fix: Client-side URL redirect

ThanoshanMV:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#1038
___
### Bugfix
- Description
Untrusted URL redirection due to user-provided value.

- How to fix?
We'll get the `redirectUrl` from `decodeURIComponent(redirect)`. Extract hostname from the `redirectUrl` and check if the current host and  `redirectUrl`'s host are same. If yes, we'll perform the redirect action otherwise we don't perform any redirect action. As `window.location.href` considered risk, we'll use `history.push(path)`.  `path` will be extracted from `redirectUrl`


web/src/pages/Setting/Setting.tsx
```diff
@@ -62,7 +62,15 @@ const Setting: React.FC = () => {
         });
         setTimeout(() => {
           const redirect = getUrlQuery('redirect');
-          window.location.href = redirect ? decodeURIComponent(redirect) : '/';
+          const currentHost = window.location.host;
+          if (redirect) {
+            const redirectUrl = decodeURIComponent(redirect);
+            const redirectHost = redirectUrl.split('/')[2];
+            if (currentHost === redirectHost) {
+              window.location.href = redirectUrl;
```
liuxiran, line 70: > To guard against untrusted URL redirection, it is advisable to avoid putting user input directly into a redirect URL. Instead, maintain a list of authorized redirects on the server; then choose from that list based on the user input provided.

After read the recommended note from codeql, it seems that all urls derived from user input are considered risky, so only limit the same host is not enough(I seem the risky is still exit). 

Since the redirectUrl could not come from server(the recommended way), we may try to use `history.push` instead of `window.location.href` to avoid this sec risky, which can also complete the page redirect action.

membphis:
@ThanoshanMV CI failed

web/src/pages/Setting/Setting.tsx
ThanoshanMV: I studied the source code and understood that `history.push()` is done via importing history from umi. I referred the [umi documentation on `history.push()`](https://umijs.org/api#history). 

So, in line 70, can we try to replace the code like below:
`history.push(redirectUrl);`

liuxiran: > I studied the source code and understood that `history.push()` is done via importing history from umi. I referred the [umi documentation on `history.push()`](https://umijs.org/api#history).
> 
> So, in line 70, can we try to replace the code like below:
> `history.push(redirectUrl);`

you would better only pass the path in the `redirecrUrl` to the `history.push()`
so you have to：
1. extract the `path` form `redirecrUrl`
2. replace the code in line 70 to `history.push(path);`

just try it ^_^  @ThanoshanMV

liuxiran:
another unstable fe e2e teat case: https://github.com/apache/apisix-dashboard/pull/1284/checks?check_run_id=1693627365
may be we need a issue to trace all the unstable test cases.

cc @LiteSun

juzhiyuan:
image

Too curious, but @LiteSun has been working on this.

LiteSun:
> another unstable fe e2e test case: https://github.com/apache/apisix-dashboard/pull/1284/checks?check_run_id=1693627365
> may be we need a issue to trace all the unstable test cases.
> 
> cc @LiteSun

This test case error is not caused by an unstable e2e test. I ran the test locally and found that the error was caused by the page not automatically jumping after a successful configuration.

juzhiyuan:
Please fix it ASAP.

web/src/pages/Setting/Setting.tsx
```diff
@@ -62,7 +62,20 @@ const Setting: React.FC = () => {
...
+            const redirectHost = pathArray[2];
+            if (currentHost === redirectHost) {
+              let path = '';
+              for (let i = 3; i < pathArray.length; i++) {
+                path += '/';
+                path += pathArray[i];
+              }
+              history.push(path);
+            }
+          }
```
liuxiran, line 78: it still needs a case to cover when redirect is ''
```suggestion
          } else {
          history.push('/');
          }
```

liuxiran:
> > another unstable fe e2e test case: https://github.com/apache/apisix-dashboard/pull/1284/checks?check_run_id=1693627365
> > may be we need a issue to trace all the unstable test cases.
> > cc @LiteSun
> 
> This test case error is not caused by an unstable e2e test. I ran the test locally and found that the error was caused by the page not automatically jumping after a successful configuration.

I'm sorry for misjudgment, I found that there is a bug in the updated code, just wait for updating again.

cc @LiteSun @juzhiyuan

web/src/pages/Setting/Setting.tsx
ThanoshanMV: Okay, I'll commit sugesstion

web/src/pages/Setting/Setting.tsx
```diff
@@ -75,6 +75,8 @@ const Setting: React.FC = () => {
               }
               history.push(path);
             }
+          } else {
+          history.push('/');
```
juzhiyuan, line 79: you could run `yarn prettier --write .` under the `web` directory to format codes.

LiteSun: you can also click https://github.com/apache/apisix-dashboard/pull/1284/checks?check_run_id=1696195316 to find out what lint error it is.

ThanoshanMV: yarn-command
@juzhiyuan, I could not able to execute the command inside the `web` directory

juzhiyuan: Try running yarn install then have a try? It seems that the prettier Module is missing :)

ThanoshanMV: Sure, I'll try it and let you know

liuxiran (approved):

juzhiyuan (approved):

Issue #1161, opened 2020-12-30
Feat: remove the etcd dependency in the unit test


starsz:
# Feature request

## Please describe your feature

Now, `route`, `ssl` `services`  and `upstream` module are use etcd to store info.
I think we should remove the etcd dependency in the unit test.

## Describe the solution you'd like

Use store.MockInterface{} instead of using the etcd.

- [x] route
- [x] ssl
- [x] services
- [x] upstream

tokers:
Agree +1.

juzhiyuan:
If we need to resolve this issue in 2.3 or 2.4?

starsz:
> If we need to resolve this issue in 2.3 or 2.4?

I think it's a feature. So we may resolve it in 2.4.

starsz:
This is a big project, I will finish part of them in M2.4

nic-chen:
finish test cases of  route, service

Jaycean:
@starsz 
Hi,I have finished writing the SSL unit test, and if you do not start writing the upstream unit test, I can do so.

starsz:
> @starsz
> Hi,I have finished writing the SSL unit test, and if you do not start writing the upstream unit test, I can do so.

Oh, good.Thank you very much.

starsz:
Because the test cases of the route, service had finished, so I think we can change the millstone to M2.5.

starsz:
This problem was solved. Thank you for every contributor.

PR #859, opened 2020-11-23
feat: add a unit test for consumer and remove implicit init

ShiningRush:
- add a real unit test for consumer `Get`
- remove implicit init function, go is not recommended.

PR #1318, opened 2021-01-17
feat: add store mock

starsz:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

#1161
___

Add store mock so that we can use this instead of the etcd dependency in unit test.

### DIFF
--- a/api/internal/core/store/mock_interface.go
+++ b/api/internal/core/store/mock_interface.go
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package store
+
+import (
+	"context"
+
+	"github.com/stretchr/testify/mock"
+)
+
+type MockInterface struct {
+	mock.Mock
+}
+
+func (m *MockInterface) Get(key string) (interface{}, error) {
+	ret := m.Called(key)
+
+	if rf, ok := ret.Get(0).(func(key string) (interface{}, error)); ok {
+		return rf(key)
+	}
+
+	var r0 interface{}
+	if rf, ok := ret.Get(0).(func(string) interface{}); ok {
+		r0 = rf(key)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(interface{})
+		}
+	}
+
+	var r1 error
+	if rf, ok := ret.Get(1).(func(string) error); ok {
+		r1 = rf(key)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
+func (m *MockInterface) List(input ListInput) (*ListOutput, error) {
+	ret := m.Called(input)
+
+	if rf, ok := ret.Get(0).(func(ListInput) (*ListOutput, error)); ok {
+		return rf(input)
+	}
+
+	var r0 *ListOutput
+	if rf, ok := ret.Get(0).(func(ListInput) *ListOutput); ok {
+		r0 = rf(input)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(*ListOutput)
+		}
+	}
+
+	var r1 error
+	if rf, ok := ret.Get(1).(func(ListInput) error); ok {
+		r1 = rf(input)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
+func (m *MockInterface) Create(ctx context.Context, obj interface{}) (interface{}, error) {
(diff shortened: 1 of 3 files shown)


api/internal/core/store/mock_interface.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+type MockInterface struct {
+	mock.Mock
+}
+
+func (m *MockInterface) Get(key string) (interface{}, error) {
+	ret := m.Called(key)
+
+	if rf, ok := ret.Get(0).(func(key string) (interface{}, error)); ok {
```
tokers, line 32: Could we just use a type assert switch statement here? The type assertions are all around.

api/internal/core/store/store_mock.go
```diff
@@ -18,48 +18,184 @@ package store
...
 
+	"github.com/apisix/manager-api/internal/core/entity"
+	"github.com/shiningrush/droplet/data"
 	"github.com/stretchr/testify/mock"
 )
 
-type MockInterface struct {
-	mock.Mock
+type Mock struct {
+	MockInterface
```
tokers, line 31: It would be better to keep an empty line after the embed object.

api/internal/core/store/store_mock.go
```diff
@@ -18,48 +18,184 @@ package store
...
+
+	if m.validator != nil {
+		if err := m.validator.Validate(obj); err != nil {
+			return nil, err
+		}
+	}
+
+	key := m.keyFunc(obj)
+	if key == "" {
+		return nil, fmt.Errorf("key is required")
```
tokers, line 66: If the error construction is just a plain string, just use `errors.New`.

api/internal/core/store/store_mock.go
```diff
@@ -18,48 +18,184 @@ package store
...
+	rf := func(ctx context.Context, obj interface{}, createIfNotExist bool) error {
+		if m.validator != nil {
+			if err := m.validator.Validate(obj); err != nil {
+				return err
+			}
+		}
+
+		key := m.keyFunc(obj)
+		if key == "" {
+			return fmt.Errorf("key is required")
```
tokers, line 158: Ditto, just use `errors.New` is enough.

api/internal/core/store/store_mock_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"testing"
+
+	"github.com/apisix/manager-api/internal/core/entity"
+	"github.com/apisix/manager-api/internal/utils"
+	"github.com/shiningrush/droplet/data"
+	"github.com/stretchr/testify/assert"
+)
+
+type testObj struct {
+	entity.BaseInfo
```
tokers, line 32: Keep an empty line after the embed object.

api/internal/core/store/store_mock_test.go
```diff
@@ -0,0 +1,200 @@
...
+	m := NewMock(keyFunc, nil)
+
+	// create success
+	ret, err := m.Create(ctx, &testObj{Name: "test"})
+	assert.Nil(t, err)
+	assert.Equal(t, "test", ret.(*testObj).Name)
+
+	// create failed, "key is required"
+	ret, err = m.Create(ctx, &testObj{Name: ""})
+	assert.Equal(t, fmt.Errorf("key is required"), err)
```
tokers, line 52: 1. Maybe `assert.Error` method is more proper.
2. Use `errors.New`.

api/internal/core/store/mock_interface.go
starsz: Not very agree. Because this file is generated by mockery, and I do little modification.
So maybe we should keep the original structure.

ShiningRush:
Why we need this PR?I think the original mock store is enough, please have a look at #859

api/internal/core/store/store_mock_test.go
starsz: Updated.

api/internal/core/store/store_mock.go
starsz: Updated.Thank you very much.

api/internal/core/store/store_mock.go
starsz: Updated.

api/internal/core/store/store_mock.go
starsz: Updated

starsz:
> Why we need this PR? I think the original mock store is enough, please have a look at #859

Original mock doesn't have the store ability, so we need to provide mock data in every unit test.
IMO, I think this is more like a fake ETCD and can help we write unit test more convenient.

api/internal/core/store/mock_interface.go
tokers: Better to add some comments to mark that the file is generated automically, just like kubebuilder. Otherwise we cannot distinguish the code source.

ShiningRush:
> Original mock doesn't have the store ability, so we need to provide mock data in every unit test.

Yes, this is the job that stub module should do, if you want to reuse stub data, you should just extract those data instead of the logic.
Put the store logic into mock object will cause `Shotgun Surgery` that when we change the logic of store, we need also change the mock object's loigc.

nic-chen:
> Yes, this is the job that stub module should do, if you want to reuse stub data, you should just extract those data instead of the logic.
Put the store logic into mock object will cause Shotgun Surgery that when we change the logic of store, we need also change the mock object's logic.

that's what I want to say

juzhiyuan:
ping @starsz Please resolve those conflicts.

starsz:
@ShiningRush @nic-chen Thanks for your advice.

Sorry, I don't think too much about it.
Maybe we can discuss it.

@juzhiyuan This PR is WIP now.

juzhiyuan:
is this urgent?

membphis:
ping @starsz

starsz:
I will close this PR.
When we need `Integration Testing` then we can reopen it.

PR #1277, opened 2021-01-12
feat: return back the data just created via POST method in manager API

imjoey:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

Related to https://github.com/apache/apisix-dashboard/issues/1154 .

This PR is only going to support HTTP POST method for returning data. In order to completely fix #1154 , we also need to fix that for HTTP PUT methods. I will commit another PR for that soon.

___
### New feature or improvement
- Describe the details and related test reports.

As mentioned in #1154 , when creating an object via `POST` method in manager-api, there is no way to refer to that new object for API caller. This PR is going to additionally return all the object data stored in etcd back. See the following example:

Request sent to manager-api:

```
$ curl http://127.0.0.1:9000/apisix/admin/routes -H 'X-API-KEY: edd1c9f034335f136f87ad84b625c8f1' -H 'Authorization:auth-token' -X POST -d '{"uri": "/get","methods": ["GET"],"upstream": {"type": "roundrobin","nodes": {"httpbin.org:443": 1}},"plugins": {"proxy-rewrite": {"uri": "\/get","scheme": "https"},"key-auth": {},"response-rewrite": {"status_code": 200,"body": "{\"code\": 200, \"msg\": \"success\"}"}}}'
```

Response back to API caller:
```
{"code":0,"message":"","data":{"id":"337109164457198464","create_time":1610462319,"update_time":1610462319,"uri":"/get","methods":["GET"],"plugins":{"key-auth":{},"proxy-rewrite":{"scheme":"https","uri":"/get"},"response-rewrite":{"body":"{\"code\": 200, \"msg\": \"success\"}","status_code":200}},"upstream":{"nodes":{"httpbin.org:443":1},"type":"roundrobin"},"status":1},"request_id":"f9520603-528a-4494-9934-5620788f2ec2"
```

Issue #1154, opened 2020-12-30
Resource create request should return the ultimate data


tokers:
# Feature request

## Please describe your feature

When create resources like Route without specifying id, the ultimate data should be returned, otherwise people cannot reference the just created resource by id.

![image](https://user-images.githubusercontent.com/10428333/103332608-8f598500-4aa5-11eb-856e-c2a23fd97740.png)

## Describe the solution you'd like

Output the data in the response.

starsz:
Agree with it.

IMO, we need to keep the same response as apisix.

membphis:
agree with this too

imjoey:
@tokers @starsz @membphis Users could create a resource via `POST` or `PUT` methods. No way to refer the new object only occurs at `POST` method. Shall we also return the ultimate data when using `PUT` method?  FYI, the answer is yes for APISIX, need to make manager-api the same as Admin API?

Looking forward to your insights. Thanks.

membphis:
> Shall we also return the ultimate data when using PUT method? FYI, the answer is yes for APISIX, need to make manager-api the same as Admin API?

the Admin API and Manager API should work in the same way. 
So I think the answer is yes, they should return the final data at the same time, whether it is PUT or POST method.

imjoey:
> > Shall we also return the ultimate data when using PUT method? FYI, the answer is yes for APISIX, need to make manager-api the same as Admin API?
> 
> the Admin API and Manager API should work in the same way.
> So I think the answer is yes, they should return the final data at the same time, whether it is PUT or POST method.

@membphis copy. Thanks for the directions.

juzhiyuan:
> we also need to fix that for HTTP PUT methods

This is needed.

imjoey:
> > we also need to fix that for HTTP PUT methods
> 
> This is needed.

@juzhiyuan yep, I will commit an PR for this.

imjoey:
@tokers  @juzhiyuan  since both #1277 and #1322 had been merged, I could close this issue. Please feel free to reopen it if anything unexpected. Thanks.

PR #1322, opened 2021-01-18
feat: Add returning value for HTTP PUT and PATCH methods

imjoey:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

Related to https://github.com/apache/apisix-dashboard/issues/1154 .

___
### Bugfix
- Description

As described in #1154 , when manipulating an object via PUT/PATCH method in manager-api, there is no data returned.

- How to fix?

 In order to get along with APISIX Admin API, this PR is going to additionally return all the object data back for  PUT or PATCH methods. 

> This PR is also a counterpart of https://github.com/apache/apisix-dashboard/pull/1277 , will completely revolve #1154 .

### DIFF
--- a/api/internal/handler/consumer/consumer.go
+++ b/api/internal/handler/consumer/consumer.go
@@ -147,11 +147,12 @@ func (h *Handler) Set(c droplet.Context) (interface{}, error) {
 	input.Consumer.ID = input.Consumer.Username
 	ensurePluginsDefValue(input.Plugins)
 
-	if err := h.consumerStore.Update(c.Context(), &input.Consumer, true); err != nil {
+	ret, err := h.consumerStore.Update(c.Context(), &input.Consumer, true)
+	if err != nil {
 		return handler.SpecCodeResponse(err), err
 	}
 
-	return nil, nil
+	return ret, nil
 }
 
 func ensurePluginsDefValue(plugins map[string]interface{}) {
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -284,8 +284,7 @@ func (s *GenericStore) Update(ctx context.Context, obj interface{}, createIfNotE
 	storedObj, ok := s.cache.Load(key)
 	if !ok {
 		if createIfNotExist {
-			_, err := s.Create(ctx, obj)
-			return nil, err
+			return s.Create(ctx, obj)
 		}
 		log.Warnf("key: %s is not found", key)
 		return nil, fmt.Errorf("key: %s is not found", key)
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -38,7 +38,7 @@ type Interface interface {
 	Get(key string) (interface{}, error)
 	List(input ListInput) (*ListOutput, error)
 	Create(ctx context.Context, obj interface{}) (interface{}, error)
-	Update(ctx context.Context, obj interface{}, createIfNotExist bool) error
+	Update(ctx context.Context, obj interface{}, createIfNotExist bool) (interface{}, error)
 	BatchDelete(ctx context.Context, keys []string) error
 }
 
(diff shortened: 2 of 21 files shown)


imjoey:
@nic-chen @ShiningRush   CI passed and ready for review. Thanks

nic-chen:
LGTM,  add some E2E test should be better. thanks @imjoey

tokers (approved):

api/internal/handler/consumer/consumer.go
```diff
@@ -147,11 +147,12 @@ func (h *Handler) Set(c droplet.Context) (interface{}, error) {
...
 	ensurePluginsDefValue(input.Plugins)
 
-	if err := h.consumerStore.Update(c.Context(), &input.Consumer, true); err != nil {
+	ret, err := h.consumerStore.Update(c.Context(), &input.Consumer, true)
+	if err != nil {
 		return handler.SpecCodeResponse(err), err
 	}
 
-	return nil, nil
+	return ret, nil
```
ShiningRush, line 155: We can directly return `&input.Consumer`, because store is just return the pointer of it.
So that we can avoid to return object in `Store`, keep one way to get changes would be better.

nic-chen (approved):

api/internal/handler/consumer/consumer.go
juzhiyuan: ping @imjoey

imjoey:
> LGTM, add some E2E test should be better. thanks @imjoey

@nic-chen agree with you and thanks for the advice. Do you mean lacking for test cases on `PATCH` methods? 

AFAIK, we indeed need much more backend tests cases for both E2E and Unit test. Cause almost all test case only use `PUT` for creating and updating, which leads to extremely low testing coverage for creating by `POST` and updating by `PATCH`.

The test cases in this PR will only try to cover the changes within this PR. So how about we commit other PRs for improving the testing coverage? I would be happy to commit an issue for discussion.

Looking forward to your feedback. Much appreciated.

api/internal/handler/consumer/consumer.go
imjoey: @ShiningRush  Thanks for your reviewing. IMHO, directly returning the `&input.Consumer` would mix up the DTO and the DAO, which may be confusing to developer. Please allow me use the following case as an example:

```go
type SetInput struct {
	entity.Consumer
	Username string `auto_read:"username,path"`
}

func (h *Handler) Set(c droplet.Context) (interface{}, error) {
	input := c.Input().(*SetInput)
	if input.ID != nil && utils.InterfaceToString(input.ID) != input.Username {
		return &data.SpecCodeResponse{StatusCode: http.StatusBadRequest},
			fmt.Errorf("consumer's id and username must be a same value")
	}
	if input.Username != "" {
		input.Consumer.Username = input.Username
	}
	input.Consumer.ID = input.Consumer.Username
	ensurePluginsDefValue(input.Plugins)

	err := h.consumerStore.Update(c.Context(), &input.Consumer, true)
	if err != nil {
		return handler.SpecCodeResponse(err), err
	}

        // Notice: Desensitize the response data  =======
        input.Consumer.Desc = "*******"

	return &input.Consumer, nil
}

```

As the comment mentioned above, in order to do customization before sending back to client, we need to change input value for output, which seems some sort of werid.

Looking forward to your insights. Thanks.

ShiningRush: > directly returning the &input.Consumer would mix up the DTO and the DAO

In this case, input is not only `DTO` also a resource.
I support to return resource in `Patch` because it is indeed different with input.But `Update` method required a complete resource, so DTO also is a resource.

> which may be confusing to developer

On the contrary, the situation that confuses developers is that you return an object and make them think that you did not perform any operations on the original object, but this is not the truth.

BTW: here is a better way to implement you example that move logic to entity,so code could be
```go
....

input.Consumer.Desensitize()
return &input.Consumer, nil
```
`rich domain model` would be more meaningful.

imjoey: @ShiningRush 

Thanks for getting back. Let's wait for others opinions. 😄 

> BTW: here is a better way to implement you example that move logic to entity,so code could be
Thanks. The code you provided is awesome. While IMHO, it is still not able to resolve the mixup of concepts of request and response.

nic-chen:
> The test cases in this PR will only try to cover the changes within this PR. So how about we commit other PRs for improving the testing coverage? I would be happy to commit an issue for discussion.

that should be great. thanks.

imjoey:
@ShiningRush  I think you are right. Simply considering the `&input.Consumer` as a Resource will make the question be easier. I will renew the PR according to your suggestions. Thanks.

BTW, in https://github.com/apache/apisix-dashboard/pull/1277 , I also added the returned data for `store.Create`. Shall I revert that change to keep same as this?

juzhiyuan:
ok, we may create an issue to track this if needed :)

imjoey:
> ok, we may create an issue to track this if needed :)

@juzhiyuan  @ShiningRush  Thanks. Then I will commit an issue for tracking the `Create`/`Update` methods of `Storage`. 😄

PR #1324, opened 2021-01-18
feat: pass context for `Get` and `List`

nic-chen:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

___

pass context for function `Get` and `List`, which is convenient to pass custom parameters or other data in custom development

### DIFF
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -136,9 +136,8 @@ func (s *GenericStore) Init() error {
 	return nil
 }
 
-func (s *GenericStore) Get(ctx context.Context, key string) (interface{}, error) {
+func (s *GenericStore) Get(_ context.Context, key string) (interface{}, error) {
 	ret, ok := s.cache.Load(key)
-	_ = ctx
 	if !ok {
 		log.Warnf("data not found by key: %s", key)
 		return nil, data.ErrNotFound
@@ -174,9 +173,8 @@ var defLessFunc = func(i, j interface{}) bool {
 	return iID < jID
 }
 
-func (s *GenericStore) List(ctx context.Context, input ListInput) (*ListOutput, error) {
+func (s *GenericStore) List(_ context.Context, input ListInput) (*ListOutput, error) {
 	var ret []interface{}
-	_ = ctx
 	s.cache.Range(func(key, value interface{}) bool {
 		if input.Predicate != nil && !input.Predicate(value) {
 			return true
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -35,8 +35,8 @@ import (
 )
 
 type Interface interface {
-	Get(key string) (interface{}, error)
-	List(input ListInput) (*ListOutput, error)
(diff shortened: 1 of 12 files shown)


api/internal/core/store/store.go
```diff
@@ -35,8 +35,8 @@ import (
 )
 
 type Interface interface {
-	Get(key string) (interface{}, error)
-	List(input ListInput) (*ListOutput, error)
+	Get(ctx context.Context, key string) (interface{}, error)
+	List(ctx context.Context, input ListInput) (*ListOutput, error)
```
ShiningRush, line 39: `Get` and `List` is work in memory, how to use the context?

api/internal/core/store/store.go
```diff
@@ -136,8 +136,9 @@ func (s *GenericStore) Init() error {
 	return nil
 }
 
-func (s *GenericStore) Get(key string) (interface{}, error) {
+func (s *GenericStore) Get(ctx context.Context, key string) (interface{}, error) {
 	ret, ok := s.cache.Load(key)
+	_ = ctx
```
tokers, line 141: If `ctx` is not in use, just use dummy variable when define the function argument.

api/internal/core/store/store.go
```diff
@@ -173,8 +174,9 @@ var defLessFunc = func(i, j interface{}) bool {
 	return iID < jID
 }
 
-func (s *GenericStore) List(input ListInput) (*ListOutput, error) {
+func (s *GenericStore) List(ctx context.Context, input ListInput) (*ListOutput, error) {
 	var ret []interface{}
+	_ = ctx
```
tokers, line 179: Ditto.

api/internal/core/store/store.go
nic-chen: we don't use it currently, but it's is convenient to pass custom parameters or other data in custom development

api/internal/core/store/store.go
nic-chen: fixed.

api/internal/core/store/store.go
nic-chen: fixed.

tokers (approved):

api/internal/core/store/store.go
ShiningRush: Keep parameter enough would be better, is here any concret cases which must need context? It will need context only when do some io operations(disk io or network io) as far as I know, store is responsible to query data in memory, so it looks like will never do IO operations.

starsz (approved):

api/internal/core/store/store.go
membphis: please @nic-chen confirm

nic-chen: For example, there may be some business parameters that need to be passed. If we pass these custom parameters one by one, the handler code needs to be modified, and the code will be conflicts  with upstream. If use `ctx` to deliver, user only need to pay attention to the business part.

juzhiyuan (approved):

Issue #1325, opened 2021-01-18
allow - character in the OnlineDebug module


juzhiyuan:
# Feature request

## Please describe your feature

We need to improve Regexp to allow the `-` character in the OnlineDebug module's URI input box.

## Describe the solution you'd like

Improve Regexp.

## Additional context

![image](https://user-images.githubusercontent.com/2106987/104902058-2e90de80-59b9-11eb-9efa-f32e5c3a3b22.png)

PR #1308, opened 2021-01-15
test: Create Upstream in Route.

guoqqqi:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
resolve #1125  
___
E2E: create Upstream in Route.
1. Create Upstream in Route
2. Select Upstream in Route then Input manually

PR #1326, opened 2021-01-18
fix(fe): update url regexp in online debug

liuxiran:

- [ ] Bugfix

- Related issues
fix: #1325 
___
### Bugfix
- Description

- How to fix?
update url Regexp to allow `-` in url path, at the same time allow `%` `+` `-` in query params

___

already added.

### DIFF
--- a/web/config/defaultSettings.ts
+++ b/web/config/defaultSettings.ts
@@ -31,6 +31,14 @@ export default {
   title: 'APISIX Dashboard',
   pwa: false,
   iconfontUrl: '',
+  serveUrlMap: {
+    dev: 'http://139.217.190.60',
+    test: 'http://localhost:9000',
+  }
 } as LayoutSettings & {
   pwa: boolean;
+  serveUrlMap: {
+    dev: string;
+    test: string;
+  }
 };
--- a/web/config/defaultSettings.ts
+++ b/web/config/defaultSettings.ts
@@ -34,11 +34,11 @@ export default {
   serveUrlMap: {
     dev: 'http://139.217.190.60',
     test: 'http://localhost:9000',
-  }
+  },
 } as LayoutSettings & {
   pwa: boolean;
   serveUrlMap: {
     dev: string;
     test: string;
-  }
+  };
 };
--- a/web/cypress/integration/route/online-debug.spec.js
+++ b/web/cypress/integration/route/online-debug.spec.js
@@ -16,7 +16,7 @@
  */
 /* eslint-disable no-undef */
 import menuLocaleUS from '../../../src/locales/en-US/menu';
-import RouteLocaleUS from '../../../src/pages/Route/locales/en-US';
+import routeLocaleUS from '../../../src/pages/Route/locales/en-US';
 import defaultSettings from '../../../config/defaultSettings';
 
 context('Online debug', () => {
@@ -46,15 +46,15 @@ context('Online debug', () => {
     cy.contains(menuLocaleUS['menu.routes']).click();
 
     // show online debug draw
-    cy.contains(RouteLocaleUS['page.route.onlineDebug']).click();
+    cy.contains(routeLocaleUS['page.route.onlineDebug']).click();
 
     // input uri with specified special characters
     urisWithSpecialChars.forEach((uri) => {
       cy.get(domSelector.uriInput).clear();
       cy.get(domSelector.uriInput).type(`${defaultSettings.serveUrlMap[SERVE_ENV].split('//').pop()}/${uri}`);
-      cy.contains(RouteLocaleUS['page.route.button.send']).click();
+      cy.contains(routeLocaleUS['page.route.button.send']).click();
       // should not show the notification
-      cy.contains(RouteLocaleUS['page.route.input.placeholder.requestUrl']).should('not.exist');
+      cy.contains(routeLocaleUS['page.route.input.placeholder.requestUrl']).should('not.exist');
     });
   });
 });
--- a/web/cypress/integration/route/online-debug.spec.js
+++ b/web/cypress/integration/route/online-debug.spec.js
@@ -20,18 +20,24 @@ import routeLocaleUS from '../../../src/pages/Route/locales/en-US';
 import defaultSettings from '../../../config/defaultSettings';
 
 context('Online debug', () => {
-  const urisWithSpecialChars = [
-    'localhost:9000/get',
-    '127.0.0.1:9000/get',
-    'baidu.com/get',
-    'localhost:9000/deps-get',
-    'localhost:9000/deps.get',
-    'localhost:9000/deps_get',
-    'localhost:9000/get?search=1',
-    'localhost:9000/get?search=%U%E',
-    'localhost:9000/get?search=v1&number=-1',
-    'localhost:9000/get?search=1+1',
-  ];
+  const data = {
+    uris: [
+      'localhost:9000/get',
+      '127.0.0.1:9000/get',
+      'www.baidu.com/get',
+      'baidu.com/get',
+      'test-host.com/get',
+      'test-host.com/aBCdsf/defsdf456eweWQE',
+      'localhost:9000/deps-get',
+      'localhost:9000/deps.get',
+      'localhost:9000/deps_get',
+      'localhost:9000/get?search=1',
+      'localhost:9000/get?search=v1,sd&number=-1',
+      'localhost:9000/get?search=1+1',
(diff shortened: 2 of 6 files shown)


juzhiyuan:
@guoqqqi Please take a look at this PR, if anything you are not clear, you could leave comments.

web/cypress/integration/route/online-debug.spec.js
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* eslint-disable no-undef */
+
+context('Online debug', () => {
+  const ServerHost = '139.217.190.60';
```
LiteSun, line 20: why not using `http://localhost:9000`?

liuxiran: > why not using `http://localhost:9000`?

Actually it would be better to use APISIX's address here, and I'm not sure this if this address is also for APISIX(It shouldn't be, is it)

BTW, will github CI work well if  I use `http://localhost:9000`? @LiteSun

web/cypress/integration/route/online-debug.spec.js
```diff
@@ -42,6 +42,7 @@ context('Online debug', () => {
 
     // input uri with specified special characters
     urisWithSpecialChars.forEach((uri) => {
+      cy.get('#debugUri').clear();
```
juzhiyuan, line 45: https://github.com/apache/apisix-dashboard/pull/1308/files#diff-f7a862c3853a71e325c09d7d25c05ccd96ba223ad611a71302d3547e8163275cR20-R28

Please take a look at this example, we may move all those mappers to a specified file in the future, but for now, we could use this way.

web/cypress/integration/route/online-debug.spec.js
LiteSun: Not sure if APISIX is started on the backend.  Need @nic-chen to confirm.

What I am trying to say is: we should not rely on outside services for testing 😊

web/cypress/integration/route/online-debug.spec.js
liuxiran: It's a good idea to unified define `domSelectors`, fixed.

web/cypress/integration/route/online-debug.spec.js
nic-chen: > Not sure if APISIX is started on the backend. Need @nic-chen to confirm.
> 
> What I am trying to say is: we should not rely on outside services for testing 😊

agree with you, we should not rely on outside services for testing. if needed, we could run it in CI.

web/config/defaultSettings.ts
```diff
@@ -31,6 +31,14 @@ export default {
   title: 'APISIX Dashboard',
   pwa: false,
   iconfontUrl: '',
+  serveUrlMap: {
+    dev: 'http://139.217.190.60',
+    test: 'http://localhost:9000',
+  },
```
liuxiran, line 37: I found that both fe dev and cy test need this config, I moved it to the `defaultSettings`

web/cypress/integration/route/online-debug.spec.js
```diff
@@ -0,0 +1,60 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* eslint-disable no-undef */
+import menuLocaleUS from '../../../src/locales/en-US/menu';
+import routeLocaleUS from '../../../src/pages/Route/locales/en-US';
+import defaultSettings from '../../../config/defaultSettings';
```
liuxiran, line 20: in order to reuse the definition, I imported the locales files instead of using the text of the dom. 

cc @juzhiyuan @LiteSun

web/cypress/integration/route/online-debug.spec.js
```diff
@@ -54,6 +54,9 @@ context('Online debug', () => {
       cy.get(domSelector.uriInput).type(`${defaultSettings.serveUrlMap[SERVE_ENV].split('//').pop()}${uri}`);
       cy.contains(routeLocaleUS['page.route.button.send']).click();
       // should not show the notification
+      
+      // test ci failed 
+      cy.wait(500)
```
juzhiyuan, line 59: Could we use `interceptor` to check if that request is finished?

liuxiran: > Could we use `interceptor` to check if that request is finished?

In fact, this test case does not have to wait for the response of the request, because It only needs to validate the url
and it passed in my local env. 

However, it is possible that multiple requests are made frequently and errors are reported because of blocking. PLS let me try to find out the exact cause of the problem first, then I'll try `interceptor` ^_^

liuxiran: The regular expression does not take into account the case of `localhost`, that is why CI failed🤣, I will update soon, thanks all your help @LiteSun @juzhiyuan

nic-chen:
image

we should need some more test cases for it..

juzhiyuan:
> image
> 
> we should need some more test cases for it..

https://github.com/apache/apisix-dashboard/pull/1326/files#diff-20cb223ab12b7a91431dc130015be393007e4b28b1fbb8b0a91d315c1e2f3986R23-R32

web/cypress/integration/route/online-debug.spec.js
```diff
@@ -21,19 +21,20 @@ import defaultSettings from '../../../config/defaultSettings';
...
-    '/deps-get',
-    '/deps.get',
-    '/deps_get',
-    '/get?search=1',
-    '/get?search=%U%E',
-    '/get?search=v1&number=-1',
-    '/get?search=1+1',
+    'localhost:9000/get',
+    '127.0.0.1:9000/get',
+    'baidu.com/get',
```
juzhiyuan, line 26: `baidu` may not exist here, could we use a library to do this check?

liuxiran: > `baidu` may not exist here, could we use a library to do this check?

It will take a little time to research, let me have a try

liuxiran:
> image
> 
> we should need some more test cases for it..

this pr will solve your problem @nic-chen , at the same time I will check the regexp again, and add more test cases, thanks for your advice.

web/cypress/integration/route/online-debug.spec.js
liuxiran: for data: `baidu.com/get` just a case to test the host which does not contain `www`, and network does not affect the execution of test case, so I keep `baidu.com/get` in the data ^_^ @juzhiyuan 

and for the regexp,  I just tried a simpler one to cover complex situations.

juzhiyuan:
@liuxiran  just for reference: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url

liuxiran:
> @liuxiran just for reference: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url

thanks a lot, I tried it, which covers all data except `localhost:9000`

web/src/pages/Route/components/DebugViews/DebugDrawView.tsx
```diff
@@ -143,7 +143,7 @@ const DebugDrawView: React.FC<RouteModule.DebugDrawProps> = (props) => {
 
   const handleDebug = (url: string) => {
     /* eslint-disable no-useless-escape */
-    const urlReg = /^(?=^.{3,255}$)(www\.)?[a-zA-Z0-9][-a-zA-Z0-9]{0,62}(\.[a-zA-Z0-9][-a-zA-Z0-9]{0,62})+(:\d+)*(\/\w*)*(\.\w+)*([\?&]\w+=\w*)*$/;
+    const urlReg = /^[0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])*(:(0-9)*)*(\/?)([a-zA-Z0-9\-\=\.\"\?\,\'\/\\\+&amp;%$#_]*)?$/;
```
juzhiyuan, line 146: image

juzhiyuan (approved):

LiteSun (approved):

imjoey (approved):
/lgtm

PR #1344, opened 2021-01-21
fix: ignore the init_dir event when watching etcd events

### LINKED ISSUE #1329: bug: non-json data causes error logs
# Bug report

## Describe the bug

json parse failed log in error.log

## How to Reproduce

1. start Manager API (new installation)
2. start APISIX (new installation)

## Expected behavior

no error log

## System information

- OS: [e.g. macOS, Windows] macOS
- Browser (if applies) [e.g. Chrome, Safari, Edge]
- Version: [e.g. 2.0-rc3] master

## Additional context

first time installation and start APISIX after the Manager API started.

imjoey:
Signed-off-by: imjoey <majunjie@apache.org>

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

Fixes #1329 .

___
### Bugfix
- Description

Current implementation only ignores the `init_dir` etcd event at `init` stage, not at `watch` stage. This would lead to JSON unmarshaling error when re-initialization the etcd data via `apisix init_etcd` command.

- How to fix?

This PR will make an additional check for each event at `watch` stage, skipping the `init_dir` ones.

IMHO, each kind of storage and even each kind of data in single storage, will have their own rules to check if the data is skippable. So this PR will also move the `init_dir` judgement from store to storage.

This PR also introduce the `Keypair` into `Event` as the embedded field, to simply the definitions.

### DIFF
--- a/api/internal/core/storage/etcd.go
+++ b/api/internal/core/storage/etcd.go
@@ -28,6 +28,12 @@ import (
 	"github.com/apisix/manager-api/internal/utils"
 )
 
+const (
+	// SkippedValueInitDir indicates the init_dir etcd
+	// event will be skipped
+	SkippedValueInitDir = "init_dir"
+)
+
 var (
 	Client *clientv3.Client
 )
@@ -120,6 +126,8 @@ func (s *EtcdV3Storage) List(ctx context.Context, key string) ([]Keypair, error)
 		data := Keypair{
 			Key:   string(resp.Kvs[i].Key),
 			Value: string(resp.Kvs[i].Value),
+			// Mark the init_dir etcd event as skippable if value is init_dir
+			Skipped: string(resp.Kvs[i].Value) == SkippedValueInitDir,
 		}
 		ret = append(ret, data)
 	}
@@ -171,8 +179,13 @@ func (s *EtcdV3Storage) Watch(ctx context.Context, key string) <-chan WatchRespo
 
 			for i := range event.Events {
 				e := Event{
-					Key:   string(event.Events[i].Kv.Key),
-					Value: string(event.Events[i].Kv.Value),
+					Keypair: Keypair{
+						Key:   string(event.Events[i].Kv.Key),
+						Value: string(event.Events[i].Kv.Value),
+						// Mark the init_dir etcd event as skippable
+						// if it's a init_dir event
+						Skipped: string(event.Events[i].Kv.Value) == SkippedValueInitDir,
+					},
 				}
 				switch event.Events[i].Type {
 				case clientv3.EventTypePut:
--- a/api/internal/core/storage/etcd.go
+++ b/api/internal/core/storage/etcd.go
@@ -29,9 +29,9 @@ import (
 )
 
 const (
-	// SkippedValueInitDir indicates the init_dir etcd
-	// event will be skipped
-	SkippedValueInitDir = "init_dir"
+	// SkippedValueEtcdInitDir indicates the init_dir
+	// etcd event will be skipped
+	SkippedValueEtcdInitDir = "init_dir"
 )
 
 var (
@@ -127,7 +127,7 @@ func (s *EtcdV3Storage) List(ctx context.Context, key string) ([]Keypair, error)
 			Key:   string(resp.Kvs[i].Key),
 			Value: string(resp.Kvs[i].Value),
 			// Mark the init_dir etcd event as skippable if value is init_dir
-			Skipped: string(resp.Kvs[i].Value) == SkippedValueInitDir,
+			Skipped: string(resp.Kvs[i].Value) == SkippedValueEtcdInitDir,
 		}
 		ret = append(ret, data)
 	}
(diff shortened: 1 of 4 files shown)


tokers:
could we add a test case to verify that these events are really ignored?

imjoey:
> could we add a test case to verify that these events are really ignored?

@tokers  sure, Thanks.

imjoey:
> > could we add a test case to verify that these events are really ignored?
> 
> @tokers sure, Thanks.

@tokers Since the skipped field is a built-in field in Keypair or Event, I added two skippable data in the existing test case. Eventually, the test will pass if the `wantCache` does not contain the skippable data defined in `giveListRet` and `giveResp`.
Any feedback is welcome. Thanks.

imjoey:
ping @ShiningRush @nic-chen @starsz for reviewing at your convenience. Many thanks.

api/internal/core/storage/etcd.go
```diff
@@ -120,6 +126,8 @@ func (s *EtcdV3Storage) List(ctx context.Context, key string) ([]Keypair, error)
 		data := Keypair{
 			Key:   string(resp.Kvs[i].Key),
 			Value: string(resp.Kvs[i].Value),
+			// Mark the init_dir etcd event as skippable if value is init_dir
+			Skipped: string(resp.Kvs[i].Value) == SkippedValueEtcdInitDir,
```
nic-chen, line 130: I think empty string or `{}` should be skipped too.

imjoey: @nic-chen  yep, there definitely exists various kinds of values that should be skippable for each storage type. This PR is going to explicitly skip the `init_dir` values to fix #1329 . How about adding other more condition checks when we need to deal with them?

nic-chen: APISIX may change it to `{}`, so I think we should deal with them in this PR if you have time for it. Thanks.

imjoey: > APISIX may change it to `{}`, so I think we should deal with them in this PR if you have time for it. Thanks.

@nic-chen thanks for the directions. I will set `{}` as skippable soon

tokers (approved):

nic-chen (approved):

imjoey:
@tokers @nic-chen PR renewed with marking `{}` also skippable. While honestly, for now `{}` is invalid, but may be considered to be valid in the future. So I also added relevant important comments to remind developers.

api/internal/core/storage/etcd.go
imjoey: @nic-chen  PR renewed and it's done. Thanks.

api/internal/core/storage/storage.go
```diff
@@ -36,12 +36,13 @@ type WatchResponse struct {
 type Keypair struct {
 	Key   string
 	Value string
+	// If true, this keypair will NOT be processed in store.
+	Skipped bool
```
ShiningRush, line 40: Why we need to skip it in store not storage?

imjoey: @ShiningRush

IMHO, no data processing logic in storage layer would keep it single responsibility, which is only to provide data to store. Store will handle more logic to determine if really going to skip the data during processing. Looking forward to your insights.

imjoey: @ShiningRush  From my point of view, the difference is about the design.

1. Storage provides ALL data to Store for further processing
>  This would lead to skip data in Store, leaving Storage out of data processing
2. Storage provides ONLY VALID data to Store for further processing
> This would guide us to eliminate invalid data before sending to Store, as you mentioned.

Looking forward to your insights.

imjoey: @ShiningRush 

Cause the original skip check is done in Store, I guess your opinion is the design 1. If not, we could also use design 2 instead. It's not a heavy workload. 😄

ShiningRush: In fact, I prefer option 2, because invalid data should be cleaned from the source. If the skip logic of storage requires store participation, then this part of the content can be provided to the store as options instead of a `Skipped` flag, which is just a notification to the store and does not have much meaning.

BTW: I did not participate in the review of the previous skip logic, so I don't know it before.
Sometimes I’m busy, so I can’t participate in every pr, sorry : (

nic-chen: @ShiningRush 
Just reminder, the placeholder could not be cleared, because APISIX needs to simulate a directory like `/apisix/route/`, otherwise there will be problems with `watch`.

ShiningRush: Got it

imjoey: @ShiningRush

> If the skip logic of storage requires store participation, then this part of the content can be provided to the store as options instead of a `Skipped` flag, which is just a notification to the store and does not have much meaning.

Thanks for your opinion. I would use design 2 to refactor this functionality.

> BTW: I did not participate in the review of the previous skip logic, so I don't know it before.
> Sometimes I’m busy, so I can’t participate in every pr, sorry : (

That's OK. Just try enjoy the open source. 😄

PR #1302, opened 2021-01-14
feat: user can skip upstream when select service_id

LiteSun:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

- Related issues
resolve #1213, resolve #1278 
___
### New feature or improvement
- Describe the details and related test reports.
user can skip upstream when select service_id.
___
### Please add the corresponding test cases if necessary.
- [ ] Todo
___

PR #1361, opened 2021-01-22
test: skip upstream after service is seleted

guoqqqi:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
resolve #1302


web/cypress/integration/route/create-route-can-skip-upstream.spec.js
```diff
@@ -0,0 +1,141 @@
+/*
```
juzhiyuan, line 1: bad title

juzhiyuan: test what

juzhiyuan (approved):

web/cypress/integration/route/can-skip-upstream-when-select-service-id.spec.js
```diff
@@ -18,19 +18,25 @@
 
 context('Can select service_id skip upstream in route', () => {
   const data = {
-    upstreamName: 'test_upstream',
-    serviceName: 'test_service',
-    routeName: 'test_route',
+    upstream_name: 'test_upstream',
```
LiteSun, line 21: Why use UnderScoreCase？

juzhiyuan: it's called snake_case and the API will return such a case, to reduce transformation, we allow to use that case.

LiteSun (approved):

Issue #1383, opened 2021-01-27
chore: When exporting route, whether the parameter is empty or not


Jaycean:
# Bug report

## Describe the bug

- When exporting route data, the current logic will export empty swagger data when the route data is null or there is no route data. Whether to add judgment will directly return an error.

```
{
  "components": {},
  "info": {
    "title": "RoutesExport",
    "version": "3.0.0"
  },
  "openapi": "3.0.0",
  "paths": {}
}
```

## Expected behavior

- When null characters are passed in, routeid is not passed in, or all routeid data is exported, there is no data in the database, and an error should be reported directly.

Jaycean:
cc @liuxiran @nic-chen @starsz

starsz:
Agree with you. Thank you for your PR.

nic-chen:
we could close it now since the pr merged.

PR #1384, opened 2021-01-27
 fix: When exporting route, whether the parameter is empty or not

Jaycean:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#1383 
___
### Bugfix
- When exporting route data, the current logic will export empty swagger data when the route data is null or there is no route data. Whether to add judgment will directly return an error.

- How to fix?
When null characters are passed in, routeid is not passed in, or all routeid data is exported, there is no data in the database, and an error should be reported directly.

### DIFF
--- a/api/internal/handler/data_loader/route_export_test.go
+++ b/api/internal/handler/data_loader/route_export_test.go
@@ -19,6 +19,7 @@ package data_loader
 
 import (
 	"encoding/json"
+	"errors"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


api/internal/handler/data_loader/route_export_test.go
```diff
@@ -1903,10 +1904,66 @@ func TestExportRoutesSameURI(t *testing.T) {
...
+	mStore.On("Get", mock.Anything).Run(func(args mock.Arguments) {
+	}).Return(route, nil)
+
+	h := Handler{routeStore: mStore}
+	exportInput := &ExportInput{}
+	exportInput.IDs = ""
+	ctx := droplet.NewContext()
+	ctx.SetInput(exportInput)
+	_, err1 := h.ExportRoutes(ctx)
+	assert.Equal(t, errors.New("Parameter IDs cannot be empty"), err1)
```
tokers, line 1936: What about pre defining error object and reusing them here, if the description changed, we have to change it here.

```go
var ErrEmptyID = errors.New("empty id")
```

Jaycean: I personally think that this return error is used by this test alone and only once. I don't think it's necessary to declare a new variable

api/test/e2e/route_export_test.go
```diff
@@ -26,6 +26,24 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestExport_Route_Data_Empty(t *testing.T) {
+	tests := []HttpTestCase{
```
liuxiran, line 30: is it necessary to confirm whether the route list is empty?

Jaycean: I think this test returns an error when the route list is empty. Before this test, make sure whether the route list is empty will be repeated.

api/internal/handler/data_loader/route_export_test.go
```diff
@@ -1903,10 +1904,66 @@ func TestExportRoutesSameURI(t *testing.T) {
...
+			},
+			"type": "roundrobin"
+		}
+}`
+	var route *entity.Route
+	err := json.Unmarshal([]byte(r1), &route)
+	assert.Nil(t, err)
+	mStore := &store.MockInterface{}
+	mStore.On("Get", mock.Anything).Run(func(args mock.Arguments) {
+	}).Return(route, nil)
```
starsz, line 1928: IMO, mStore.Get will never be called if the input.IDs is "".
So we can remove the mock here.

Jaycean: done. Thks for review

starsz (approved):

liuxiran (approved):

PR #1408, opened 2021-02-01
feat: support return the manager api's git hash and version

### LINKED ISSUE #1404: We can get the backend version and git hash at the front end.
# Feature request

Hello, there is no way to get the backend version and the git hash at the front end.
But I think it's in need.

Backend adds an API like `/apisix/admin/version" to return the version and git hash.

starsz:

- [] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

fix #1404

Add `info` API to support get the backend info.


membphis (commented):
please use a better title, it confused me

api/internal/handler/tool/tool.go
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package tool
+
+import (
+	"github.com/apisix/manager-api/internal/handler"
```
nic-chen, line 20: style

api/internal/handler/tool/tool.go
```diff
@@ -0,0 +1,49 @@
...
+	Hash    string `json:"commit_hash"`
+	Version string `json:"version"`
+}
+
+func NewHandler() (handler.RouteRegister, error) {
+	return &Handler{}, nil
+}
+
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.GET("/info", wgin.Wraps(h.Info))
```
nic-chen, line 40: `/info` doesn't feel reasonable…

tokers: What about `/version`?

api/internal/utils/utils.go
```diff
@@ -206,3 +211,14 @@ func ValueEqual(a interface{}, b interface{}) bool {
 	}
 	return bytes.Equal(aBytes, bBytes)
 }
+
+// set the hash and version
+func SetHashAndVersion(hash, version string) {
```
tokers, line 216: Can't understand why this function is required.

If you want to assign these two values from the go linker, just referencing them in the `-X` option.

starsz: Yes. We can.

api/internal/handler/tool/tool.go
starsz: Got it.

api/internal/handler/tool/tool.go
starsz: We can add more fields in the future if we use `/info`.
If we only want to add `hash` and `version` fields then we could use `/version`.

juzhiyuan: This endpoint is used to know about more information for the manager-api (Dashboard), API_VERSION, commit hash, etc.

@imjoey what's your opinion?

api/internal/handler/tool/tool.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (h *Handler) Info(c droplet.Context) (interface{}, error) {
+	hash, version := utils.GetHashAndVersion()
+	return &InfoOutput{
+		Hash:    hash,
+		Version: version,
```
juzhiyuan, line 48: `Version` is a little confusing, it's manager-api's version actually 🤔

api/test/e2e/info_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestInfo(t *testing.T) {
+	tests := []HttpTestCase{
+		{
+			Desc:         "get info",
+			Object:       ManagerApiExpect(t),
+			Method:       http.MethodGet,
+			Path:         "/info",
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusOK,
+			ExpectBody:   []string{"commit_hash", "\"version\":\"master\""},
```
juzhiyuan, line 33: where is `commit_hash`'s value?

api/internal/handler/tool/tool.go
starsz: Maybe the described work should be done in the FE.
In fact, it's the version of manager-api. There is no problem.

api/test/e2e/info_test.go
starsz: Because the commit_hash is always changing. So we only guarantee the key exists.

api/internal/utils/version.go
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package utils
+
+var (
+	gitHash string
```
tokers, line 20: Why not just exposing these two variables 😂

api/internal/handler/tool/tool.go
juzhiyuan: because the manager-api and web are under the same repo `apisix-dashboard`, when users call it, we will think it's dashboard's version instead 🤔 

cc @liuxiran @ShiningRush what do you think?

api/test/e2e/info_test.go
juzhiyuan: ok

api/internal/handler/tool/tool.go
imjoey: I would prefer `/version` for now.
Just FYI, `version` indicates the details about the distribution of manager-api binary, such as `OS`, `Arch`, `Go version`, `API version` and etc. `info` indicates the status and metrics of the running manager-api binary, such as `connections`, `memory usage` and etc. 

Please see the output `docker version` and `docker info` for examples. Maybe we could follow that. 😄 

```
➜  apisix-dashboard git:(master) docker version
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:12:42 2020
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8891c58
  Built:            Mon Dec 28 16:15:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```
➜  apisix-dashboard git:(master) docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.5.0)

Server:
 Containers: 9
  Running: 8
  Paused: 0
  Stopped: 1
 Images: 36
 Server Version: 20.10.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc version: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 init version: de40ad0
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.121-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 1.943GiB
 Name: docker-desktop
 ID: E2J3:HRSM:RBMG:UQ4R:ASEK:GBXM:G5FG:KYI4:WZPM:SDQI:L7GF:EYFJ
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 HTTP Proxy: gateway.docker.internal:3128
 HTTPS Proxy: gateway.docker.internal:3129
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

api/internal/handler/tool/tool.go
imjoey: AFAIK, manager-api and web have the same version, which is also the version of apisix-dashboard. Will the versions of manager-api and web be shown to users, respectively?  Thanks.

api/internal/utils/version.go
starsz: IMO, It's not safe enough.

api/internal/handler/tool/tool.go
starsz: Got it. Now I think `/version` is better.

api/internal/utils/version.go
imjoey: Agree with @starsz , for me, immutable variables here would be safer.

api/internal/handler/tool/tool.go
juzhiyuan: Agree with you!

api/internal/handler/tool/tool.go
juzhiyuan: OK

juzhiyuan:
@starsz try to sync with the latest codes from the master branch?

starsz:
> @starsz try to sync with the latest codes from the master branch?

It's already the latest. Please see the CI on the master branch.

juzhiyuan:
fine, then please request changes from other members, once the FE test is fixed, I will let you know.

nic-chen (approved):

tokers (approved):

juzhiyuan:
FE is working on why test failed, once test fixed, we will merge.

api/internal/handler/tool/tool.go
```diff
@@ -0,0 +1,50 @@
...
+	Hash    string `json:"commit_hash"`
+	Version string `json:"version"`
+}
+
+func NewHandler() (handler.RouteRegister, error) {
+	return &Handler{}, nil
+}
+
+func (h *Handler) ApplyRoute(r *gin.Engine) {
+	r.GET("/version", wgin.Wraps(h.Version))
```
juzhiyuan, line 41: @starsz @nic-chen 

This will lead to 2 kinds of APIs:
1. `/apisix/admin/xxx`
2. `/version`

When we deprecate `admin-api` in V3, we will have to omit that prefix `apisix/admin` either, to have ManagerAPIs.

Now FE meets one issue: because HTTP call is handled by UmiJS, which only supports one prefix, `/apisix/admin` or `/`, then FE has to make hack codes to be compatible with those 2 kinds of APIs.

Could we add a prefix for this API? with some description about this case.

starsz: OK.Got it.
I will add the prefix and ignore this API's auth checking in the next PR.

nic-chen: I'm not sure if we need to omit the prefix, but unification is appropriate.

juzhiyuan: Thanks!!!

PR #1435, opened 2021-02-05
feat: show warning notification when dashboard version not matching apisix

### LINKED ISSUE #1165: tip: display a warning message on the dashboard if the user is not using the recommended APISIX + Dashboard version
this is the recommended way.

image

If the user does not use it in this way, we can display a warning message on the dashboard.

It might be a good way to display it on the status page.

related issue: https://github.com/apache/apisix-dashboard/issues/1148

LiteSun:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
resolves #1165 

image


juzhiyuan (approved):
`Notification` will occur when the version gets mismatched and page refresh.

juzhiyuan:
`Warning Notification` is enough currently, we could support showing more information or recommendation.

imjoey:
> `Warning Notification` is enough currently, we could support showing more information or recommendation if needed.

+1 agreed. The recommendation of matched version is recommended. We could improve this later in other PR.

imjoey (approved):
/lgtm

Thanks @LiteSun .

liuxiran:
fe e2d test failed, tried to solve in #1442

liuxiran (approved):
LGTM, it will very helpful for users, thanks

liuxiran:
sorry I found that In addition to the test case run failed, there are some problems with lint, maybe it needs to solve in this pr

web/src/components/RightContent/index.tsx
```diff
@@ -49,6 +51,9 @@ const GlobalHeaderRight: React.FC = () => {
   if ((navTheme === 'dark' && layout === 'top') || layout === 'mix') {
     className = `${styles.right}  ${styles.dark}`;
   }
+  useEffect(() => {
```
liuxiran, line 54: > Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function. You can follow the documentation here. (from https://stackoverflow.com/questions/57620799/react-hook-useeffect-is-called-conditionally)

just move this `useEffect` before line 44

LiteSun:
> sorry I found that In addition to the test case run failed, there are some problems with lint, maybe it needs to solve in this pr

Thanks for the review, I will fix it.

PR #1445, opened 2021-02-06
chore: sync JSON Schema from APISIX version 2.3

nic-chen:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

 sync JSON Schema from APISIX version 2.3

PR #1451, opened 2021-02-07
test: make create route with upstream test cases stable

liuxiran:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description
1. https://github.com/apache/apisix-dashboard/runs/1843264349
I reproduced this failure locally, the error occurred due to It has not changed to step two after click `Next` button in step one  
2021-02-07 15-12-51屏幕截图
Solution:
add `cy.get(this.domSelector.name).should('value', data.route_name);` to ensure edit page has ready loaded
add `{force: true}` to `Next` button click to ensure the button clicked

2. https://github.com/apache/apisix-dashboard/pull/1445/checks?check_run_id=1847652754
I reproduced this failure locally, the error occurred due to It has stop on login page after `should create an upstream` run off, I tried to merge `should enter the Route creator` `should disable Upstream input boxes after selecting an existing upstream` and `should submit custom Upstream properties successfully` into one case `should create route with upstream just created`, because it will do login before every test case runs

### DIFF
--- a/web/cypress/integration/route/create-route-with-upstream.spec.js
+++ b/web/cypress/integration/route/create-route-with-upstream.spec.js
@@ -26,6 +26,7 @@ context('Create Route with Upstream', () => {
     ip2: '127.0.0.2',
     delete_route_success: 'Delete Route Successfully',
     delete_upstream_success: 'Delete Upstream Successfully',
+    submitSuccess: 'Submit Successfully',
   };
 
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


web/cypress/integration/route/create-route-with-upstream.spec.js
```diff
@@ -46,32 +47,27 @@ context('Create Route with Upstream', () => {
     cy.contains('Submit').click();
   });
 
-  it('should enter the Route creator', function () {
+  it('should create route with upstream just created', function () {
```
juzhiyuan, line 50: put those into smaller test cases because they are clearer, as for login, could we try to put Line33~Line35 under `beforeAll`?

juzhiyuan:
cc @guoqqqi to review.

juzhiyuan:
ping @guoqqqi

web/cypress/integration/route/create-route-with-upstream.spec.js
liuxiran: > put those into smaller test cases because they are clearer, as for login, could we try to put Line33~Line35 under beforeAll?

Let me have a try~

liuxiran: While learning how to use `beforeAll`, I saw the official recommendation best practices of cypress[1], 
> Anti-Pattern: Coupling multiple tests together. 
Best Practice: Tests should always be able to be run independently from one another and still pass.

> Anti-Pattern: Acting like you’re writing unit tests.
Best Practice: Add multiple assertions and don’t worry about it

As There is an interdependence between those small test cases, it would be better to use more `assertions` in a merged test case.

Reference: 
[1] https://docs.cypress.io/guides/references/best-practices.html#Dangling-state-is-your-friend

juzhiyuan: cc @LiteSun @guoqqqi to know

guoqqqi (approved):

guoqqqi:
I agree, it seems that we need to use assertions more often to make the tests more stable. thanks a lot, that reminds me.

juzhiyuan (approved):

LiteSun (approved):

Issue #1217, opened 2021-01-05
support debug file transfer in online debug


liuxiran:
# Feature request

## Please describe your feature

In the process of solving #1196, I found that the current online debug does not support debugging file transfer. It would be better to support this feature since APISIX support it.

## Describe the solution you'd like

![image](https://user-images.githubusercontent.com/2561857/103676650-4e3d0580-4fbc-11eb-9d88-3d7685c00221.png)

I would like to implement this feature in form of request forwarding the `Content-type: multipart/form-data`, just refer to postman.

liuxiran:
a TODO feature, reopen now

Jaycean:
@liuxiran @juzhiyuan I can help with the BE, please assign to me ,Thks.

PR #1465, opened 2021-02-09
feat(be): Online debug function adds support for transfer files

Jaycean:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#1217 
#1500
___
- Online debug function adds support for transfer files


juzhiyuan:
Hi, please sync your codes with the master branch to fix the CI failure.

Jaycean:
> Hi, please sync your codes with the master branch to fix the CI failure.

Thank you. I haven't added the BE E2E test for the corresponding code. I will add it later.

api/internal/handler/route_online_debug/route_online_debug.go
```diff
@@ -63,66 +66,87 @@ type Result struct {
...
 	resp, err := client.Do(req)
 	if err != nil {
 		return &data.SpecCodeResponse{StatusCode: http.StatusInternalServerError}, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if resp != nil {
+			resp.Body.Close()
+		}
+	}()
```
starsz, line 133: Oh. What's this? Why we should wrap the function?

starsz (changes requested):

api/internal/handler/route_online_debug/route_online_debug.go
Jaycean: It can be used directly `resp.Body.Close ()` is OK. Correct reading will also close the transaction.

nic-chen:
hi @Jaycean 

we could add e2e test cases in https://github.com/apache/apisix-dashboard/tree/master/api/test/e2enew

Jaycean:
> hi @Jaycean
> 
> we could add e2e test cases in https://github.com/apache/apisix-dashboard/tree/master/api/test/e2enew

OK, Thks, all E2E tests are added to the new folder? Or need to add both？

nic-chen:
> > hi @Jaycean
> > we could add e2e test cases in https://github.com/apache/apisix-dashboard/tree/master/api/test/e2enew
> 
> OK, Thks, all E2E tests are added to the new folder? Or need to add both？

added to the new folder is ok, thanks.

api/internal/handler/route_online_debug/route_online_debug.go
Jaycean: done.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+		bytesData, _ := json.Marshal(reqParams)
+		return bytes.NewReader(bytesData), contentType
+	}
+	if files != nil {
+		body := &bytes.Buffer{}
+		writer := multipart.NewWriter(body)
+		for _, uploadFile := range files {
+			file, err := os.Open(uploadFile.Filepath)
+			if err != nil {
+				panic(err)
```
starsz, line 119: `panic` here is not very good.
`return err` would be better.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+			}
+			part, err := writer.CreateFormFile(uploadFile.Name, filepath.Base(uploadFile.Filepath))
+			if err != nil {
+				panic(err)
```
starsz, line 123: Ditto.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+			}
+			_, err = io.Copy(part, file)
+			file.Close()
```
starsz, line 126: We can use defer.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+		}
+		for k, v := range reqParams {
+			if err := writer.WriteField(k, v); err != nil {
+				panic(err)
```
starsz, line 130: Ditto.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+			}
+		}
+		if err := writer.Close(); err != nil {
+			panic(err)
```
starsz, line 134: Ditto.

starsz (changes requested):

api/test/e2enew/base/http.go
Jaycean: done.

api/test/e2enew/base/http.go
Jaycean: done.

api/test/e2enew/base/http.go
Jaycean: done.

api/test/e2enew/base/http.go
Jaycean: done.

api/test/e2enew/base/http.go
Jaycean: done.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+			file, err := os.Open(uploadFile.Filepath)
+			if err != nil {
+				return nil, "", err
+			}
+			part, err := writer.CreateFormFile(uploadFile.Name, filepath.Base(uploadFile.Filepath))
+			if err != nil {
+				return nil, "", err
+			}
+			_, err = io.Copy(part, file)
+			defer file.Close()
```
starsz, line 126: `defer file.Close()` should be put when the file is opened. At line 121.
Else if the writer.CreateFormFile's err is not nil, the file wouldn't be closed. This will cause `fd leak`.

api/test/e2enew/base/http.go
```diff
@@ -92,3 +104,44 @@ func BatchTestServerPort(times int) map[string]int {
...
+	}
+
+	urlValues := url.Values{}
+	for key, val := range reqParams {
+		urlValues.Set(key, val)
+	}
+
+	reqBody := urlValues.Encode()
+
+	return strings.NewReader(reqBody), contentType, nil
```
starsz, line 146: Put the `param` to the body? This confused me.

Jaycean: Yes, because the sending file is added, the data type in the body is bytes. If it is put in the body, it is not convenient to read. So I communicated with @liuxiran at that time, so it can only put other parameters in the header. Do you have a better suggestion?

Jaycean: http.go In the getReader function, It from the old E2E/ http.go.

My idea is that E2E will be rewritten with the new ginkgo, so I migrate it directly here without modifying the code.

My understanding here is to deal with a variety of data formats, such as form-data, JSON, files.

The incoming params may be in a variety of situations. Here is the third one url.Values set reqParams map[string]string

api/test/e2enew/base/http.go
Jaycean: Yes, this is my negligence.
Done.
Thks.

api/test/e2enew/base/http.go
starsz: @nic-chen Do you know about this?

nic-chen: I feel that there is a problem with this. The content of the file should not be read in the test case, but the content of the uploaded file should be read in the main program.

Jaycean: @liuxiran 

Because the front-end has cross-domain problems, it needs to conduct a transparent transmission through the manage API. 

Here, the front-end places the file bytes in the body and transfers them to the back-end, and then directly transmits them through the back-end. Therefore, the E2E test needs to read the file here.

starsz: Sorry, my question is that why we should put the `reqParams` to `reqBody`?

liuxiran: Request params include url, prototal, method, header params and body params, they used to form the new request rewarding by golang.

Before this pr, we put all request params to the request body, and we use `json` to transfor data
due to this new feature, the origin request support not only `json`, `form-data` is accepted. Inorder not to disturb the origin formdata, we put others params including: url, prototal, method and header params to the request header, and the reqest body keep the origin body, so that the forwarding request can use the body directly.

starsz: @liuxiran Thanks for the explanation. 😄

starsz (approved):

api/test/e2enew/route_online_debug/route_online_debug_test.go
```diff
@@ -0,0 +1,785 @@
...
+			Body: `{
+				"uri": "/hello",
+				"methods": ["GET"],
+				"plugins": {
+					"key-auth": {}
+				},
+				"upstream": {
+					"type": "roundrobin",
+					"nodes": [{
+						"host": "172.16.238.20",
```
nic-chen, line 406: replace this IP in this file too, @Jaycean, thanks.

nic-chen (approved):

api/test/e2enew/route_online_debug/route_online_debug_test.go
Jaycean: done. Thks.

PR #1479, opened 2021-02-18
feat: added tip on the Metrics page

juzhiyuan:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

 added tip on the Metrics page, to tell users that the Metrics page should be used with Grafana Share URL, and it's stored in local only.

image

NOTE: this is only a tip, no need to add a testcase.

### DIFF
--- a/web/cypress/integration/metrics/metrics-smoketest.spec.js
+++ b/web/cypress/integration/metrics/metrics-smoketest.spec.js
@@ -31,6 +31,6 @@ context('metrics page smoke test', () => {
       .children()
       .should('contain', 'Metrics')
       .and('contain', 'You have not configured Grafana')
-      .and('contain', 'Configure Now');
+      .and('contain', 'Configure');
   });
 });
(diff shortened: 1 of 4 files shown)


liuxiran (approved):
LGTM.

BTW we stored the monitor url locally, how about turn the url as a global config setting, so that when we configed grafana start dashboard, all the user could visit the monitor page without setting the Grafana URL in setting page?

juzhiyuan:
> LGTM.
> 
> BTW we stored the monitor url locally, how about turn the url as a global config setting, so that when we configed grafana start dashboard, all the user could visit the monitor page without setting the Grafana URL in setting page?

Just to make sure, does this mean store a local grafana default URL?

liuxiran:
> > LGTM.
> > BTW we stored the monitor url locally, how about turn the url as a global config setting, so that when we configed grafana start dashboard, all the user could visit the monitor page without setting the Grafana URL in setting page?
> 
> Just to make sure, does this mean store a local grafana default URL?

emm, I mean store the url in the config file(this may need manager-api's help) instead of localstorage ^_^

juzhiyuan:
Got it, and if users want a more powerful dashboard, they could use `Grafana` directly than a single page.

LiteSun (approved):

Issue #1500, opened 2021-02-23
Feat: Rewrite all E2E tests using the Ginkgo


Jaycean:
# Feature request

Rewrite all E2E tests using the Ginkgo.

There are a lot of E2E tests that need to be rewritten, so we record the rewriting of E2E in this issue.

## E2E test rewrite list (by file name)
- [ ] skywalking-route_with_trace_plugin_test
- [ ] authentication_test  - @nic-chen 
- [ ] balancer_test  - @nic-chen 
- [ ] consumer_test  - @nic-chen 
- [ ] global_rule_test  - @nic-chen 
- [ ] id_compatible_test  - @nic-chen 
- [ ] json_schema_validate_test  - @nic-chen 
- [ ] label_test  - @nic-chen 
- [ ] route_export_test  - @Jaycean 
- [ ] route_import_test - @bisakhmondal 
- [x] route_service_upstream_test  - @Jaycean 
- [ ] route_remote_addr_test  - @Jaycean 
- [x] route_test   - @nic-chen 
- [x] route_with_management_fileds_test - @bisakhmondal 
- [x] route_with_methods_test - @bisakhmondal 
- [x] route_with_plugin_cors_test  - @bisakhmondal 
- [x] route_with_plugin_http_logger_test
- [ ] route_with_plugin_jwt_test - @bisakhmondal 
- [x] route_with_plugin_limit_count_test  - @Jaycean 
- [x] route_with_plugin_orchestration_test  - @Jaycean 
- [x] route_with_plugin_prometheus_test  - @Jaycean 
- [x] route_with_plugin_proxy_rewrite_test  - @Jaycean 
- [x] route_with_priority_test  - @Jaycean 
- [x] route_with_script_luacode_test  - @Jaycean 
- [x] route_with_uri_uris_test  - @Jaycean 
- [x] route_with_valid_remote_addr_test  - @Jaycean 
- [x] route_with_vars_test  - @Jaycean 
- [x] server_info_test  - @Jaycean 
- [x] service_test  - @Jaycean 
- [x] ssl_test  - @Jaycean 
- [x] upstream_chash_hash_on_test  - @Jaycean 
- [x] upstream_chash_query_string_arg_xxx_test  - @Jaycean 
- [x] upstream_test  - @Jaycean 
- [x] version_test  - @Jaycean 
- [x] route_online_debug_test  - @Jaycean

Jaycean:
The PR and issue of the completed E2E can be associated and recorded.

[...]

PR #1560, opened 2021-03-05
feat: rewrite e2e test(route-with-valid-remote-addr-test) with ginkgo

Jaycean:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#1500

___
-  Rewrite e2e test(route-with-valid-remote) with ginkgo


nic-chen (commented):
Using `table.Entry` is better for us, and we could clearly see the comparison before and after the modification.

starsz (approved):

Jaycean:
> Using `table.Entry` is better for us, and we could clearly see the comparison before and after the modification.

Here also need to create route and replace upstream_ id，`table.entry` There is no way to create a new variable.

nic-chen:
> > Using `table.Entry` is better for us, and we could clearly see the comparison before and after the modification.
> 
> Here also need to create route and replace upstream_ id，`table.entry` There is no way to create a new variable.

the old test cases are all table driven

Jaycean:
> > > Using `table.Entry` is better for us, and we could clearly see the comparison before and after the modification.
> > 
> > 
> > Here also need to create route and replace upstream_ id，`table.entry` There is no way to create a new variable.
> 
> the old test cases are all table driven

It's my problem. I'll revise it

nic-chen (approved):

nic-chen:
merged. thanks a lot. @Jaycean

PR #1569, opened 2021-03-09
feat: supports changing number of executing cpu cores for manager api

### LINKED ISSUE #682: feature: support custom CPU quantity used for `manager api`
- [ ] support custom cup quantity used for `manager api` 

- [ ] support config custom cup quantity in conf file

bisakhmondal:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
resolves #682 
___

Supports changing the number of os threads attached to the goroutines.

For performance reason, if in the configuration file, the defined `max_cpu` is greater than the number of available cpu cores (including hyperthreading), the `manager api` will stick to the max available cores. As it has a certain limit to boost the performance since the context switching of goroutines from the same os thread is a lot faster than goroutines from different threads.

### DIFF
--- a/api/internal/conf/conf.go
+++ b/api/internal/conf/conf.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/tidwall/gjson"
 	"gopkg.in/yaml.v2"
@@ -96,6 +97,7 @@ type Conf struct {
 	Listen    Listen
 	Log       Log
 	AllowList []string `yaml:"allow_list"`
+	MaxCpu    int `yaml:"max_cpu"`
 }
 
 type User struct {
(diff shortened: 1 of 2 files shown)


api/internal/conf/conf.go
```diff
@@ -249,3 +254,16 @@ func initEtcdConfig(conf Etcd) {
...
+func initParallelism(choiceCores int) {
+	if choiceCores < 0 {
+		return
+	}
+	maxSupportedCores := runtime.NumCPU()
+
+	runtime.GOMAXPROCS(maxSupportedCores)
+	if choiceCores < maxSupportedCores {
+		runtime.GOMAXPROCS(choiceCores)
+	}
```
starsz, line 268: > GOMAXPROCS sets the maximum number of CPUs that can be executing simultaneously and returns the previous setting. It defaults to the value of runtime.NumCPU. If n < 1, it does not change the current setting. This call will go away when the scheduler improves.

As the doc said, I think this will be more clear.

```golang
maxSupportedCores := runtime.NumCPU()
if choiceCores > maxSupportedCores {
      choiceCores = maxSupportedCores
}

runtime.GOMAXPROCS(choiceCores)
```

starsz (commented):
Pretty good.Thank you for your contribution.

api/internal/conf/conf.go
bisakhmondal: yeah, sure. Thanks for the review.

api/internal/conf/conf.go
```diff
@@ -249,3 +254,16 @@ func initEtcdConfig(conf Etcd) {
 		Prefix:    prefix,
 	}
 }
+
+// initialize parallelism settings
+func initParallelism(choiceCores int) {
+	if choiceCores < 0 {
```
nic-chen, line 260: how about
```suggestion
	if choiceCores < 1 {
```

because 0 means no limit.

bisakhmondal: True. It makes sense.

nic-chen (approved):
LGTM

starsz (approved):

nic-chen:
@bisakhmondal  merged, thanks a lot.

bisakhmondal:
You're welcome. Thanks for the review guys.

PR #1571, opened 2021-03-10
chore: improve ssl e2e test coverage

starsz:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Because some function in SSL e2e test is not covered.
This PR is to cover those functions.
image

### DIFF
--- a/api/test/e2enew/ssl/ssl_suite_test.go
+++ b/api/test/e2enew/ssl/ssl_suite_test.go
@@ -25,12 +25,11 @@ import (
 	"e2enew/base"
 )
 
-func TestRoute(t *testing.T) {
+func TestSSL(t *testing.T) {
 	ginkgo.RunSpecs(t, "ssl suite")
 }
 
 var _ = ginkgo.AfterSuite(func() {
 	base.CleanResource("ssl")
-	base.CleanResource("routes")
 	time.Sleep(base.SleepTime)
 })
--- a/api/test/e2enew/ssl/ssl_test.go
+++ b/api/test/e2enew/ssl/ssl_test.go
@@ -100,7 +100,7 @@ var _ = ginkgo.Describe("SSL Basic", func() {
 
 		_, err := http.Get("https://www.test2.com:9443")
 		assert.NotNil(t, err)
-		assert.EqualError(t, err, "Get \"https://www.test2.com:9443\": remote error: tls: internal error")
+		assert.EqualError(t, err, "Get https://www.test2.com:9443: remote error: tls: internal error")
 	})
 
 	table.DescribeTable("test ssl basic", func(testCase base.HttpTestCase) {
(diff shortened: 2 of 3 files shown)


starsz:
cc @Jaycean Can you review my code when you are free?

Jaycean:
> cc @Jaycean Can you review my code when you are free?

https://github.com/apache/apisix-dashboard/pull/1571/checks?check_run_id=2074615023

ci failed. I have also encountered this problem. Need to modify the matching characters.

starsz:
> > cc @Jaycean Can you review my code when you are free?
> 
> https://github.com/apache/apisix-dashboard/pull/1571/checks?check_run_id=2074615023
> 
> ci failed. I have also encountered this problem. Need to modify the matching characters.

Let me found the root cause.

api/test/e2enew/ssl/ssl_test.go
```diff
@@ -26,38 +26,80 @@ import (
...
-		body, err := json.Marshal(map[string]interface{}{
-			"id":   "1",
-			"cert": string(testCert),
-			"key":  string(testKey),
-			"labels": map[string]string{
-				"build":   "16",
-				"env":     "production",
-				"version": "v3",
+	t := ginkgo.GinkgoT()
+	var (
```
tokers, line 37: A better pattern is to put all these initialization logic to `ginkgo.BeforeEach()`, anyway, it's just a suggestion, you can optimize them in the future.

starsz: Yes. I had used `ginkgo.BeforeEach()` before, but we use those variables as the parameter later.So the `BeforeEach` can't modify those variables.

nic-chen (approved):

api/test/e2enew/ssl/ssl_suite_test.go
```diff
@@ -25,12 +25,11 @@ import (
...
 )
 
-func TestRoute(t *testing.T) {
+func TestSSL(t *testing.T) {
 	ginkgo.RunSpecs(t, "ssl suite")
 }
 
 var _ = ginkgo.AfterSuite(func() {
 	base.CleanResource("ssl")
-	base.CleanResource("routes")
```
Jaycean, line 34: @starsz @nic-chen 
I have a question here. If we create a route in the test case and delete the route, and verify that the route is indeed deleted, do we still need to executeAfterSuite(base.CleanResource("resource") to ensure the deletion of resources?

starsz: I think we don't need to do `execute AfterSuite(base.CleanResource("resource") `

starsz: By the way, I remove the response check. cc @nic-chen 

https://github.com/apache/apisix-dashboard/pull/1571/files#diff-a15775a07088498daed7a85df9b271a05806905c320e38029391c37bef35fb37L277

nic-chen: ok

Jaycean: OK

Jaycean (approved):

PR #1591, opened 2021-03-14
fix: fix CYPRESS cross platform env error

stu01509:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description
When we run the `yarn cypress:open-dev ` on Windows, it will occur the 'CYPRESS_SERVE_ENV' is not recognized as an internal or external command error.
image

- How to fix?
Add the `cross-env` in the npm script.
image

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -29,9 +29,9 @@
     "start:test": "cross-env REACT_APP_ENV=test MOCK=none umi dev",
     "test:component": "umi test ./src/components",
     "tsc": "tsc",
-    "cypress:open": "CYPRESS_SERVE_ENV=test cypress open",
-    "cypress:open-dev": "CYPRESS_SERVE_ENV=dev cypress open",
-    "cypress:run-ci": "CYPRESS_SERVE_ENV=test cypress run"
+    "cypress:open": "cross-env CYPRESS_SERVE_ENV=test cypress open",
+    "cypress:open-dev": "cross-env CYPRESS_SERVE_ENV=dev cypress open",
+    "cypress:run-ci": "cross-env CYPRESS_SERVE_ENV=test cypress run"
   },
   "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


web/package.json
```diff
@@ -29,9 +29,9 @@
     "start:test": "cross-env REACT_APP_ENV=test MOCK=none umi dev",
     "test:component": "umi test ./src/components",
     "tsc": "tsc",
-    "cypress:open": "CYPRESS_SERVE_ENV=test cypress open",
-    "cypress:open-dev": "CYPRESS_SERVE_ENV=dev cypress open",
-    "cypress:run-ci": "CYPRESS_SERVE_ENV=test cypress run"
+    "cypress:open": "cross-env CYPRESS_SERVE_ENV=test cypress open",
```
juzhiyuan, line 32: `cross-env` is used in Windows IMO 🤔 we can't set it here directly.

juzhiyuan: Maybe a FAQ or some tips would be better?

stu01509: But If we don't use the `cross-env` to set the env variables, the cypress definitely will fail to start.

juzhiyuan: fail on windows, right? does this work on Linux?

stu01509: Fail on Windows, working in Linux successfully

juzhiyuan: ok, cc @LiteSun to take a look., I remember he had faced this case.

LiteSun: there are many ways to set env, see: https://docs.cypress.io/guides/guides/environment-variables.html#Option-4-env , and  `cross-env` is ok IMO🤔

LiteSun (approved):

juzhiyuan (approved):

PR #1598, opened 2021-03-16
feat: check if the service is used by route when deleting

starsz:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Now, when deleting a service, we don't check whether there is a route using it.
This may lead to the DP error if the user deletes the service by mistake.

So, we should check if the service is using by route when deleting.
If the service is using, we return 400 to guarantee the DP works well.

___

Added.
___

### DIFF
--- a/api/internal/handler/service/service.go
+++ b/api/internal/handler/service/service.go
@@ -247,7 +247,7 @@ func (h *Handler) BatchDelete(c droplet.Context) (interface{}, error) {
 			fmt.Errorf("route: %s is using this service", ret.Rows[0].(*entity.Route).Name)
 	}
 
-	if err := h.serviceStore.BatchDelete(c.Context(), strings.Split(input.IDs, ",")); err != nil {
+	if err := h.serviceStore.BatchDelete(c.Context(), ids); err != nil {
 		return handler.SpecCodeResponse(err), err
 	}
 
--- a/api/internal/handler/service/service.go
+++ b/api/internal/handler/service/service.go
@@ -38,12 +38,14 @@ import (
 type Handler struct {
 	serviceStore  store.Interface
 	upstreamStore store.Interface
+	routeStore    store.Interface
 }
 
 func NewHandler() (handler.RouteRegister, error) {
 	return &Handler{
 		serviceStore:  store.GetStore(store.HubKeyService),
 		upstreamStore: store.GetStore(store.HubKeyUpstream),
+		routeStore:    store.GetStore(store.HubKeyRoute),
 	}, nil
 }
 
@@ -218,6 +220,32 @@ type BatchDelete struct {
 
 func (h *Handler) BatchDelete(c droplet.Context) (interface{}, error) {
 	input := c.Input().(*BatchDelete)
+	ids := strings.Split(input.IDs, ",")
+	mp := make(map[string]struct{})
+	for _, id := range ids {
+		mp[id] = struct{}{}
+	}
+
+	ret, err := h.routeStore.List(c.Context(), store.ListInput{
+		Predicate: func(obj interface{}) bool {
+			route := obj.(*entity.Route)
+			if _, exist := mp[utils.InterfaceToString(route.ServiceID)]; exist {
+				return true
+			}
+
+			return false
+		},
+		PageSize:   0,
+		PageNumber: 0,
+	})
+	if err != nil {
+		return handler.SpecCodeResponse(err), err
+	}
+
+	if ret.TotalSize > 0 {
+		return &data.SpecCodeResponse{StatusCode: http.StatusBadRequest},
+			fmt.Errorf("route: %s is using this service", ret.Rows[0].(*entity.Route).Name)
+	}
 
 	if err := h.serviceStore.BatchDelete(c.Context(), strings.Split(input.IDs, ",")); err != nil {
 		return handler.SpecCodeResponse(err), err
--- a/api/test/e2enew/service/service_test.go
+++ b/api/test/e2enew/service/service_test.go
@@ -83,9 +83,9 @@ var _ = ginkgo.Describe("create service without plugin", func() {
 			Method: http.MethodPut,
 			Path:   "/apisix/admin/routes/r1",
 			Body: `{
- 				"uri": "/server_port",
- 				"service_id": "s1"
- 			}`,
+				"uri": "/server_port",
+				"service_id": "s1"
+			}`,
 			Headers:      map[string]string{"Authorization": base.GetToken()},
 			ExpectStatus: http.StatusOK,
 			Sleep:        base.SleepTime,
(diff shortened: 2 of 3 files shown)


nic-chen (approved):

api/internal/handler/service/service.go
```diff
@@ -218,8 +220,34 @@ type BatchDelete struct {
...
+		Predicate: func(obj interface{}) bool {
+			route := obj.(*entity.Route)
+			if _, exist := mp[utils.InterfaceToString(route.ServiceID)]; exist {
+				return true
+			}
+
+			return false
+		},
+		PageSize:   0,
+		PageNumber: 0,
```
gxthrj, line 239: why both `pagesize` and `pagenumber` are 0 ?

starsz: We need to list all routes.

api/test/e2enew/service/service_test.go
```diff
@@ -462,3 +463,83 @@ var _ = ginkgo.Describe("service update use patch method", func() {
...
+					"port":   1980,
+					"weight": 1,
+				},
+			},
+		},
+	}
+	_createServiceBody, err := json.Marshal(createServiceBody)
+	assert.Nil(t, err)
+
+	table.DescribeTable("test service delete",
```
moonming, line 485: is there any test for DP？

moonming (changes requested):

api/test/e2enew/service/service_test.go
```diff
@@ -82,9 +83,9 @@ var _ = ginkgo.Describe("create service without plugin", func() {
 			Method: http.MethodPut,
 			Path:   "/apisix/admin/routes/r1",
 			Body: `{
-				"uri": "/server_port",
-				"service_id": "s1"
-			}`,
+ 				"uri": "/server_port",
```
tokers, line 86: Seems that you added some empty spaces?

api/test/e2enew/service/service_test.go
starsz: Yes.The DP test are here, you can see:

https://github.com/apache/apisix-dashboard/blob/01c952803d256c53ac411ca52d1c90a2263879fc/api/test/e2e/route_service_upstream_test.go#L93

api/test/e2enew/service/service_test.go
starsz: I can't found it out.

api/test/e2enew/service/service_test.go
moonming: but not in this file

starsz: And here, there are in this file: 

https://github.com/apache/apisix-dashboard/blob/01c952803d256c53ac411ca52d1c90a2263879fc/api/test/e2enew/service/service_test.go#L93

https://github.com/apache/apisix-dashboard/blob/01c952803d256c53ac411ca52d1c90a2263879fc/api/test/e2enew/service/service_test.go#L193

https://github.com/apache/apisix-dashboard/blob/01c952803d256c53ac411ca52d1c90a2263879fc/api/test/e2enew/service/service_test.go#L310

api/test/e2enew/service/service_test.go
starsz: Let me check the difference.

api/test/e2enew/service/service_test.go
moonming: not test DP for this new test case

api/test/e2enew/service/service_test.go
```diff
@@ -462,3 +463,83 @@ var _ = ginkgo.Describe("service update use patch method", func() {
...
+			ExpectBody:   "route: route1 is using this service",
+		}),
+		table.Entry("delete route first", base.HttpTestCase{
+			Desc:         "delete route first",
+			Object:       base.ManagerApiExpect(),
+			Method:       http.MethodDelete,
+			Path:         "/apisix/admin/routes/r1",
+			Headers:      map[string]string{"Authorization": base.GetToken()},
+			ExpectStatus: http.StatusOK,
+		}),
```
Jaycean, line 536: IMHO, we should add test cases that verify the data is deleted after deleting route data and services data

moonming: big +1

api/test/e2enew/service/service_test.go
starsz: IMO, different test cases test different features.
The former case tests the feature of creating a service and test it in DP.
This test case is testing for the relation of service and upstream (we can't delete service when the service is used by routes.

If this case tests the DP, I think the former test doesn't have mean to exist.

api/test/e2enew/service/service_test.go
starsz: Good capture. Let me add it.

api/test/e2enew/service/service_test.go
starsz: Change the space to tab.

starsz:
@LiteSun Please have a look at Frontend CI.

LiteSun:
> @LiteSun Please have a look at Frontend CI.

ok

nic-chen:
@starsz 
I think you should sync codes from branch master to see if there is conflict in e2e test cases.

starsz:
> @starsz
> I think you should sync codes from branch master to see if there is conflict in e2e test cases.

I think @LiteSun had done this before.

api/test/e2enew/service/service_test.go
starsz: Add test DP for this case.

tokers (approved):

nic-chen (approved):

gxthrj (approved):

moonming (approved):

api/internal/handler/service/service.go
juzhiyuan: `0` means all?

juzhiyuan: In restful APIs, when we omit `size` or `number`, it means to fetch all resources. If we pass `0`, there should return an empty array IMO.

cc @imjoey @tokers to confirm

juzhiyuan: `offset = abs(page - 1) * pageSize`

juzhiyuan (commented):
Here has one issue:

```
ret, err := h.routeStore.List(c.Context(), store.ListInput{
		Predicate: func(obj interface{}) bool {
			route := obj.(*entity.Route)
			if _, exist := mp[utils.InterfaceToString(route.ServiceID)]; exist {
				return true
			}

			return false
		},
		PageSize:   0,
		PageNumber: 0,
```

api/internal/handler/service/service.go
starsz: You can see here:

https://github.com/apache/apisix-dashboard/blob/38cf0ba32264d5a7bb4f22bd40ce037e2f66a939/api/internal/core/store/store.go#L210-L225

juzhiyuan:
cc @guoqqqi to check this failed CI

PR #1599, opened 2021-03-16
feat: check if the upstream is using by route when deleting

starsz:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Now, when deleting an upstream, we don't check whether there is a route using it.
This may lead to the DP error if the user deletes the upstream by mistake.

So, we should check if the upstream is using by route when deleting.
If the upstream is using, we return 400 to guarantee the DP works well.

___

Added.
___


nic-chen:
ci failed @starsz

membphis (approved):

nic-chen (commented):
I think we also need to check wether used by service

starsz:
> I think we also need to check wether used by service

Added.

api/test/e2enew/upstream/upstream_test.go
```diff
@@ -723,3 +724,126 @@ var _ = ginkgo.Describe("Upstream update use patch method", func() {
...
+			ExpectStatus: http.StatusOK,
+		}),
+		table.Entry("delete upstream success", base.HttpTestCase{
+			Desc:         "delete upstream success",
+			Object:       base.ManagerApiExpect(),
+			Method:       http.MethodDelete,
+			Path:         "/apisix/admin/upstreams/u1",
+			Headers:      map[string]string{"Authorization": base.GetToken()},
+			ExpectStatus: http.StatusOK,
+		}))
```
Jaycean, line 848: Do we also need to confirm whether the upstream and service data are deleted？

starsz: Yes, added.

api/test/e2enew/upstream/upstream_test.go
```diff
@@ -826,3 +827,158 @@ var _ = ginkgo.Describe("Upstream update use patch method", func() {
 		})
 	})
 })
+
+var _ = ginkgo.Describe("test upstream delete (route is using)", func() {
```
tokers, line 831: The description is not so good, `route is using` could be `route is in use`.

starsz: Good advice. Let me fix it.

starsz: Fixed.

tokers (approved):

nic-chen (approved):

gxthrj (approved):

imjoey (approved):
/lgtm

api/test/e2enew/upstream/upstream_test.go
```diff
@@ -826,3 +827,158 @@ var _ = ginkgo.Describe("Upstream update use patch method", func() {
...
+			base.RunTestCase(tc)
+		},
+		table.Entry("create upstream without plugin", base.HttpTestCase{
+			Desc:    "create upstream without plugin",
+			Object:  base.ManagerApiExpect(),
+			Method:  http.MethodPut,
+			Path:    "/apisix/admin/upstreams/u1",
+			Headers: map[string]string{"Authorization": base.GetToken()},
+			Body: `{
+					"nodes": {
```
RedemptionC, line 843: why isn't here a name field?

PR #1603, opened 2021-03-16
chore: improve user friendly on Upstream

juzhiyuan:

- [x] Improve user friendly

Improve i18n for Upstream's pass_host field & selector.

image

image


web/src/components/Upstream/UpstreamForm.tsx
```diff
@@ -728,8 +727,14 @@ const UpstreamForm: React.FC<Props> = forwardRef(
                     <Form.Item
                       label={formatMessage({ id: 'page.upstream.step.pass-host.upstream_host' })}
                       name="upstream_host"
+                      rules={[
+                        {
+                          required: true,
+                          message: "",
```
LiteSun, line 733: why the message is an empty string? 🤔

imjoey (approved):
/lgtm

The new messages are much better. Thanks.

web/src/components/Upstream/UpstreamForm.tsx
juzhiyuan: because when users leave this field empty, the input box will have a red border to alert them 🤔

juzhiyuan:
ping ~

web/src/components/Upstream/UpstreamForm.tsx
liuxiran: it seems that when the required field is empty,  the required message will appear behind the input box, so if we leave the message an empty string, there will be an empty line appear. anyway,  since the `host rewrite` become a disable option, this input box will never show in the page.

liuxiran (approved):
LGTM

web/src/components/Upstream/UpstreamForm.tsx
juzhiyuan: ya

LiteSun (approved):

LiteSun:
LGTM

moonming:
The result of the discussion in the mailing list is that only one switch option

juzhiyuan:
`[DISCUSS] about the upstream request’s Host header

It seems that the discussion is still in progress 🤔 why not review & take action on this improvement PR first?

nic-chen (approved):

PR #1605, opened 2021-03-17
fix: fix the uptime translation error

stu01509:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues 
Fix #1596

___
### Bugfix
 Fix the uptime translation error in English mode.

- How to fix?
Add moment locale.

English Mode:
image

Chinese Mode:
image

PR #1608, opened 2021-03-17
fix: unstable front-end tests

guoqqqi:

- [x] Bugfix

**Related issues**
related #1605 
___
fixed unstable front-end tests.

### DIFF
--- a/web/cypress/integration/route/import_export_route.spec.js
+++ b/web/cypress/integration/route/import_export_route.spec.js
@@ -127,10 +127,11 @@ context('import and export routes', () => {
     cy.get(this.domSelector.refresh).click();
 
     for (let i = 0; i < 2; i += 1) {
-      cy.contains('Delete').should('be.visible').click();
+      cy.contains(data[`route_name_${i}`]).siblings().contains('Delete').click();
       cy.contains('button', 'Confirm').click();
       cy.get(this.domSelector.notification).should('contain', this.data.deleteRouteSuccess);
       cy.get(this.domSelector.notificationCloseIcon).click().should('not.exist');
+      cy.reload();
     }
   });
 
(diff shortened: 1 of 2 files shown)


web/cypress/integration/route/import_export_route.spec.js
```diff
@@ -124,14 +124,13 @@ context('import and export routes', () => {
...
     cy.visit('/routes/list');
+    cy.get(this.domSelector.refresh).click();
+
     for (let i = 0; i < 2; i += 1) {
-      cy.contains(data[`route_name_${i}`])
-        .siblings()
-        .contains(componentLocaleUS['component.global.delete'])
-        .click();
-      cy.contains('button', componentLocaleUS['component.global.confirm']).click();
+      cy.contains('Delete').should('be.visible').click();
```
LiteSun, line 130: There will be two delete buttons here, and you should locate the specific delete button IMO.

guoqqqi: I don't think there is a need to find a specific delete button. Contains will only find the first element that meets the conditions.
reference: https://docs.cypress.io/api/commands/contains.html#Single-Element

LiteSun: I know the rule(only find the first element), but it's too tricky to write this way and not very friendly to new contributors IMO🤔

juzhiyuan: Would better to add more information about this Button.

guoqqqi: All right.

LiteSun (approved):

juzhiyuan (approved):
LGTM

bzp2010 (approved):

Jaycean (approved):
LGTM

Issue #1482, opened 2021-02-19
how to delete the historical version of the APISIX node.


Firstsawyou:
## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. Chrome, Safari, Edge]
- Version: [e.g. 2.0-rc3]: dashboard 2.4

How to delete the extra historical version in the APISIX node? It prompts mismatch when deleting.

![image](https://user-images.githubusercontent.com/52862365/108456747-f4fb0e00-72ab-11eb-9fa5-2e991a177921.png)

nic-chen:
Plugin `server-info` supports the expiration time of the specified information([refer](https://github.com/apache/apisix/blob/master/doc/plugins/server-info.md#how-to-customize-the-server-info-report-configurations)).


We could also support manual deletion of node data, what do you think?  @starsz @juzhiyuan

starsz:
> Plugin `server-info` supports the expiration time of the specified information([refer](https://github.com/apache/apisix/blob/master/doc/plugins/server-info.md#how-to-customize-the-server-info-report-configurations)).
> 
> We could also support manual deletion of node data, what do you think? @starsz @juzhiyuan

Agree +1.

nic-chen:
@bisakhmondal hi, could you help to fix this issue if you have time ?

bisakhmondal:
Yeah definitely.

nic-chen:
@bisakhmondal  
Thanks !
I think you could try to submit a proposal for this change  : )

nic-chen:
@bisakhmondal 
Please deal with this issue first, it is a task in milestone 2.5, thanks.

bisakhmondal:
Oh! I see. Then, I am going to look into it.
>We could also support manual deletion of node data, what do you think?

So this is the idea, right?

nic-chen:
> Oh! I see. Then, I am going to look into it.
> 
> > We could also support manual deletion of node data, what do you think?
> 
> So this is the idea, right?

sure. that's it.

LiteSun:
cc @guoqqqi to take a look.

juzhiyuan:
Hi, after discuss with @nic-chen @starsz @LiteSun, we have a new way to support this feature.

How about only adding one extra message that tells users: `APISIX Node List will get updated in 2 hours by default.`, then we don't have necessary to remove nodes by hand & calling APIs.

juzhiyuan:
[...]

PR #1583, opened 2021-03-12
fest: rewrite e2e test(route_import_test) with ginkgo

bisakhmondal:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#1500 

___
### New feature or improvement
Rewrite e2e test(route_import_test) with ginkgo.

PR #1615, opened 2021-03-18
feat: manual deletion of apisix node data

bisakhmondal:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
 #1482 
___
Manual deletion of apisix node data. Exposing a route for deleting node data.

List nodes
Screenshot from 2021-03-18 11-57-25
Delete
image
List nodes for checking
Screenshot from 2021-03-18 11-57-51

___
- [x] ~To be added~

### DIFF
--- a/api/test/e2e/server_info_delete_test.go
+++ b/api/test/e2e/server_info_delete_test.go
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package e2e
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestServerInfo_delete_node_data(t *testing.T) {
+	tests := []HttpTestCase{
+		{
+			Desc:          "delete node data for apisix-server1",
+			Object:        ManagerApiExpect(t),
+			Method:        http.MethodDelete,
+			Path:          "/apisix/admin/server_info/apisix-server1",
+			Headers:       map[string]string{"Authorization": token},
+			ExpectStatus:  http.StatusOK,
+			ExpectCode:    0,
+			ExpectMessage: "Successful",
+		},
+		{
+			Desc:         "Re requesting deleting node data for apisix-server1",
+			Object:       ManagerApiExpect(t),
+			Method:       http.MethodDelete,
+			Path:         "/apisix/admin/server_info/apisix-server1",
+			Headers:      map[string]string{"Authorization": token},
+			ExpectStatus: http.StatusNotFound,
+		},
+		{
+			Desc:         "deleting arbitrary random node data",
+			Object:       ManagerApiExpect(t),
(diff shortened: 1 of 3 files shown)


nic-chen (commented):
need unit test cases and e2e test cases

bisakhmondal:
Amazing! fresh netlify integrations :)

nic-chen (approved):

tokers (approved):

api/test/e2e/server_info_delete_test.go
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package e2e
+
+import (
+	"net/http"
+	"testing"
+)
+
```
Jaycean, line 23: We are using ginkgo to rewrite the E2E test. If you have time, you can use ginkgo to write the E2E test directly in the e2enew folder. Thks.

bisakhmondal: Hii @Jaycean, thanks for the suggestion. I intentionally didn't go with ginkgo as it runs **independent** tests concurrently/parallelly. Going with ginkgo introduces a racing condition between tests if one of the tests alters data in the environment on which the success of other test cases largely depends (which is the case here, removing a node data before other tests on server_info runs causes a failure). It breaks the entire test suite. I have faced the same scenario, you could check [here](https://github.com/apache/apisix-dashboard/pull/1583#issuecomment-799639410).
Still, I don't know any good hack to get around the issue. An alternative, not so good approach could be creating another suite for this type of scenarios (as we are running cleanups after each test suite) [ref](https://github.com/onsi/ginkgo/issues/294#issuecomment-305999281).

Just for the reconfirmation. 
image
Broke suite, haha.
Thank you.

Jaycean: Thank you for your patient explanation. I think we should discuss this problem. Do you have a better idea about the problem caused by the inability of ginkgo to run serially？ @starsz @nic-chen @tokers

juzhiyuan:
Hi, @bisakhmondal, what do you think of this solution? https://github.com/apache/apisix-dashboard/issues/1482#issuecomment-802523508

api/test/e2e/server_info_delete_test.go
starsz: IMO, I think in this situation, you can test delete the server info after the list.
Can this solve your problem?

nic-chen: In fact, our test cases are not executed parallelly by default. 
You could refer to other test cases that have been written.

starsz: Thank you for your patient answer.
I had tested this file in my local environment.

```golang
var _ = ginkgo.Describe("test", func() {
	ginkgo.It("1", func() {
		fmt.Printf("I am 1\n")
	})
	ginkgo.It("2", func() {
		fmt.Printf("I am 2\n")
	})

	ginkgo.It("3", func() {
		fmt.Printf("I am 3\n")
	})

	ginkgo.It("4", func() {
		fmt.Printf("I am 4\n")
	})
})
```
And I use `ginkgo .` to start the test.

The result is 

```
I am 1
I am 2
I am 3
I am 4
```

Do I miss anything? Very looking forward to your answer.Thank you very much.

bisakhmondal: Thanks for the details. I think, it totally depends upon the scheduler to decide the ordering, modern schedulers such as go's makes a lot of fancy optimisation before executing something. For example, without going over the overhead of context switching between multiple goroutines for such small instruction, it may wait for others to finish and then start itself up.

You can read the full issue in ginkgo mentioned in this [comment](https://github.com/apache/apisix-dashboard/pull/1615#discussion_r597392430).

Now coming back to what you suggested, I had tried that, also even I had tried to comment out intermediate describe definitions & describeTable to make the test suite containing a single describe & describeTable container with a sleep timer of 2 whopping seconds in between all the tests. 2/5 times the suit fails. The code is available [here](https://pastebin.com/vbGmh9W0)
TBH, I also get confused about seeing this erratic nature in ginkgo and after googling & reading the issue it's the only logical explanation that I came up with. 
Thanks again.

Issue #1632, opened 2021-03-20
improve some minor issues


juzhiyuan:
Here are some minor issues on the Login page we could resolve.

## TODO
- [ ] Left icon replacement. @LiteSun 
- [ ] Extracting Table as a separate component for reuse in various modules.
- [ ] APISIX plugin type change.
- [ ] GitHub hyperlink (icon) added to Header section.

iamayushdas:
Would like to work on this Assign this to me @juzhiyuan

juzhiyuan:
as for the first one, @LiteSun will do that.

iamayushdas:
> Here are some minor issues on the Login page we could resolve.
> 
> ## TODO
> * [ ]  Left icon replacement. @LiteSun
> * [ ]  Extracting Table as a separate component for reuse in various modules.
> * [ ]  APISIX plugin type change.
> * [ ]  GitHub hyperlink (icon) added to Header section.

could please brief 3rd todo?

juzhiyuan:
oh yes, we have 40+ plugins here[1], and take `basic-auth`[2] as an example, you could see it's belonged to `auth` type[3].

We have other plugins that without `type` filed or just use the `default` type field, like this one[4], we need to update them. You could find all `type` here[5]:

![image](https://user-images.githubusercontent.com/2106987/111891027-1de8fd00-8a2a-11eb-83f4-81c7f307ad7f.png)

Please use `lowercase` as `type`.

[1] https://github.com/apache/apisix/tree/master/apisix/plugins
[2] https://github.com/apache/apisix/blob/master/apisix/plugins/basic-auth.lua
[3] https://github.com/apache/apisix/blob/master/apisix/plugins/basic-auth.lua#L53
[4] https://github.com/apache/apisix/blob/master/apisix/plugins/kafka-logger.lua#L60
[5] http://apisix.apache.org/docs/apisix/plugins/batch-requests

iamayushdas:
Here
> [5] http://apisix.apache.org/docs/apisix/plugins/batch-requests
i am not able to find `type` for plugin that i have to put in local _M={}
can you please tell me more on what type will be used in transformation plugins.,i am lil confused

spacewander:
@juzhiyuan 
Your description of the `type` field is wrong!
It is nothing about the category in the documentation.

The `auth` type in the APISIX has a special meaning (although this name is quite confusing) - to mark a plugin should be interacted with the consumer.

I have explained the same thing many times:
[...]

PR #1644, opened 2021-03-22
feat: added github link to header #1632

iamayushdas:
#1632 

Github hyperlink(icon) added to header of login page


web/src/pages/User/Login.less
```diff
@@ -30,14 +30,13 @@
...
-  vertical-align: middle;
-  cursor: pointer;
+a{
+  color: black;
+  
+  &:hover {
+    color: @primary-color;
+  }
+
 }
```
stu01509, line 40: Hi @iamayushdas 

Thanks for your contribution, I think the `a` to should add the space.

web/src/pages/User/Login.tsx
```diff
@@ -74,13 +75,19 @@ const Page: React.FC = () => {
...
     history.replace('/');
     return null;
   }
   return (
     <div className={styles.container}>
       <div className={styles.lang}>
+
+        <div className={styles.github} >
+          <a className="githubIcon" href="https://apisix.apache.org"><GithubOutlined /></a> 
+        </div>
```
stu01509, line 89: Also, the link maybe can add `target='_blank'` to let the user can open the page in the new tab.

PR #1744, opened 2021-04-11
fix: use image of APISIX on docker hub instead of build from dockerfile

nic-chen:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

Due to the unavailability of the luarocks service, the image from the dockerfile build APISIX failed, which affected the back-end e2e test. So instead use the image on the docker hub to run the e2e test environment.

PR #1756, opened 2021-04-12
chore: update setup script

bisakhmondal:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Removed the part of the script responsible for downloading the apisix dockerfile.

**Related issues**
Combined with #1744 

**Checklist:**

### DIFF
--- a/api/test/docker/setup.sh
+++ b/api/test/docker/setup.sh
@@ -84,15 +84,6 @@ BACKUP_FILE="${CONF_PATH}.backup.yaml"
 
 up() {
   set -e
-  if [ ! -f "Dockerfile-apisix" ]; then
-    echo "Downloading Dockerfile"
-    if ! curl --version &>/dev/null; then
-      give_up "The script depends on curl. Please proceed after the installation."
-    else
-      curl -o Dockerfile-apisix https://raw.githubusercontent.com/apache/apisix-docker/master/alpine/Dockerfile
-    fi
-  fi
-
   #creating backup of current config
   if [ ! -f "$BACKUP_FILE" ]; then
     cp "$YAML_FILE" "$BACKUP_FILE"


nic-chen:
@bisakhmondal  we should  keep this script, we nee to download the Dockerfile when we need to test with APISIX master branch.

bisakhmondal:
> @bisakhmondal we should keep this script, we nee to download the Dockerfile when we need to test with APISIX master branch.

Hiii. Okay, sounds good. One thing I'd like to address i.e. the compose file has been changed, so if we want to build the apisix image from the downloaded dockerfile, we have to manually update the compose file (just for the testing and again put it back to the original).

As a hack, I think we can do something like `./setup.sh up --afresh` (--afresh an extra flag) which will 
1. Download the dockerfile
2. retag the already present `apache/apisix:2.4-alpine:latest` to `apache/apisix:2.4-alpine:stable`
3. Build an image from the docker file with tag `apache/apisix:2.4-alpine:latest`

A kind of abstraction that doesn't require altering the compose file. When we run `./setup.sh down`, it will revert the steps. 

How does it sound to you :)

bisakhmondal:
Any thoughts here 😄🤡? @nic-chen

nic-chen:
> > @bisakhmondal we should keep this script, we nee to download the Dockerfile when we need to test with APISIX master branch.
> 
> Hiii. Okay, sounds good. One thing I'd like to address i.e. the compose file has been changed, so if we want to build the apisix image from the downloaded dockerfile, we have to manually update the compose file (just for the testing and again put it back to the original).
> 
> As a hack, I think we can do something like `./setup.sh up --afresh` (--afresh an extra flag) which will
> 
> 1. Download the dockerfile
> 2. retag the already present `apache/apisix:2.4-alpine:latest` to `apache/apisix:2.4-alpine:stable`
> 3. Build an image from the docker file with tag `apache/apisix:2.4-alpine:latest` (just to match the name, under the hood, it may use any version of APISIX)
> 
> A kind of abstraction that doesn't require altering the compose file. When we run `./setup.sh down`, it will revert the steps.
> 
> How does it sound to you :)


hi  @bisakhmondal 
Your idea is great. 
But I still think it is enough to keep the current state. 
The current way is relatively clear.
We don’t need to invest too much time in writing this script and maintaining it. 
The benefits are not much, what do you think?

bisakhmondal:
>We don’t need to invest too much time in writing this script and maintaining it.
>The benefits are not much.

+1
Yeah, very true :)
So I think we can close this PR now.

gxthrj (approved):
I think we can merge it anyway.

bzp2010 (approved):

juzhiyuan (approved):

PR #1767, opened 2021-04-14
change: remove listen.host from api/conf/conf.yaml

### LINKED ISSUE #1743: change: remove `listen.host` from api/conf/conf.yaml
# Feature request

The configuration item `listen.host` in conf.yaml often confuses users, because when the configuration dashboard can be accessed non-locally, two places need to be modified (the other place is `allow_list`).
We can remove the `listen.host` configuration item from conf.yaml, but retain the feature, and users can add it when they need it.

1. remove `listen.host` from api/conf/conf.yaml
2. set `ServerHost` in [the code](https://github.com/apache/apisix-dashboard/blob/master/api/internal/conf/conf.go#L47
) to an empty string

nic-chen:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

The configuration item `listen.host` in conf.yaml often confuses users, because when the configuration dashboard can be accessed non-locally, two places need to be modified (the other place is allow_list).
remove the `listen.host` configuration item from conf.yaml, but retain the feature, and users can add it when they need it.

remove listen.host from api/conf/conf.yaml
set ServerHost in the code to an empty string

**Related issues**

fix/resolve #1743

**Checklist:**


juzhiyuan:
bugfix or feature?

juzhiyuan:
may sync master codes to fix CI

juzhiyuan:
Please sync the master branch

starsz (commented):
Why not set “0.0.0.0” as default？

.github/workflows/deploy-with-docker.yml
```diff
@@ -33,8 +33,9 @@ jobs:
       - name: Modify Config
         run: |
           sed -i 's/127.0.0.1:2379/172.16.238.10:2379/' api/conf/conf.yaml
-          sed -i 's/host: 127.0.0.1/host: 0.0.0.0/' api/conf/conf.yaml
```
tokers, line 36: Should add some notes about the default listen IP address(es).

nic-chen:
> Why not set “0.0.0.0” as default？

Because in our configuration, `listen` will only have  `port`.

So it would be strange to display `0.0.0.0:9000` after the `manager-api` started, `:9000` should be better. 

What do you think?  @starsz

.github/workflows/deploy-with-docker.yml
nic-chen: added.

tokers (approved):

api/conf/conf.yaml
```diff
@@ -17,10 +17,15 @@
 
 conf:
   listen:
-    host: 127.0.0.1     # `manager api` listening ip or host name
```
starsz, line 20: We still support configuring the host. So can we comment on this instead of removing them?

membphis: @starsz your way is better

api/conf/conf.yaml
```diff
@@ -17,10 +17,15 @@
...
     port: 9000          # `manager api` listening port
   allow_list:           # If we don't set any IP list, then any IP access is allowed by default.
     - 127.0.0.0/24
+    - 10.0.0.0/8
```
membphis, line 23: only one `127.0.0.0/24` is enough.

we can add more comment to show how to set it

gxthrj (approved):

api/conf/conf.yaml
nic-chen: updated.

api/conf/conf.yaml
nic-chen: updated.

starsz (approved):

starsz:
Emmm.Need to fix the link error. @nic-chen

nic-chen:
> Emmm.Need to fix the link error. @nic-chen

sure

PR #1788, opened 2021-04-19
feat: embed assets in api binary

### LINKED ISSUE #1391: feat: embed all generated web assets into api binary
# Feature request

Embed all generated web assets into api binary, so it would be convenient to run or deploy the dashboard.

Just like grafana.

bisakhmondal:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

[Comment #1391 ](https://github.com/apache/apisix-dashboard/issues/1391#issuecomment-821702445)

**Related issues**

resolve #1391 

**Checklist:**


bisakhmondal:
As `go:embed` and related package `io/fs` has been introduced in stable go version 1.16, we need to update the CI in the master branch. We can cherry-pick the CI update commit from this PR. 
cc @nic-chen

bisakhmondal:
go1.16 has been added to the environment
https://github.com/apache/apisix-dashboard/pull/1788/checks?check_run_id=2380595572#step:4:10

but still `api/test/shell/cli_test.sh` is using 1.15
https://github.com/apache/apisix-dashboard/pull/1788/checks?check_run_id=2380595572#step:6:79

any idea why🤔

nic-chen:
hi @bisakhmondal 
I think we should also embed `dag-to-lua`, see:
https://github.com/apache/apisix-dashboard/blob/master/api/build.sh#L38-L42

bisakhmondal:
> hi @bisakhmondal
> I think we should also embed `dag-to-lua`, see:
> https://github.com/apache/apisix-dashboard/blob/master/api/build.sh#L38-L42

Okay. Let me have a look. I'll update it accordingly. Thanks :)

bisakhmondal:
Hii, what about `conf.yaml`, do we also have to embed this?

nic-chen:
> Hii, what about `conf.yaml`, do we also have to embed this?

should not embed it. people need to change configuration in it.

Makefile
```diff
@@ -50,10 +59,17 @@ endif
...
 endif
-
+	@if [ "${GO_MAJOR}" -lt "${REQUIRED_MIN_GO_MAJOR}" ] ; then \
+		echo "ERROR: Need to install golang 1.16+ first"; \
+		exit 1; \
+	fi
+	@if [ "${GO_MINOR}" -lt "${REQUIRED_MIN_GO_MINOR}" ] ; then \
+		echo "ERROR: Need to install golang 1.16+ first"; \
+		exit 1; \
+	fi
```
starsz, line 72: The logic will make a mistake if we use the go version like 2.0 in the future.
: )

api/internal/route.go
```diff
@@ -54,10 +55,13 @@ func SetUpRouter() *gin.Engine {
 	r := gin.New()
 	logger := log.GetLogger(log.AccessLog)
 	r.Use(filter.CORS(), filter.RequestId(), filter.IPFilter(), filter.RequestLogHandler(logger), filter.SchemaCheck(), filter.RecoverHandler())
-	r.Use(static.Serve("/", static.LocalFile(filepath.Join(conf.WorkDir, conf.WebDir), false)))
-	r.NoRoute(func(c *gin.Context) {
-		c.File(fmt.Sprintf("%s/index.html", filepath.Join(conf.WorkDir, conf.WebDir)))
-	})
+	filesystem := fs.FS(StaticFiles)
+	subtree, err := fs.Sub(filesystem, "html")
+
```
starsz, line 60: Redundant line.

api/internal/route.go
```diff
@@ -54,10 +55,13 @@ func SetUpRouter() *gin.Engine {
...
+	if err != nil {
+		log.Errorf("%s\n", err)
```
starsz, line 62: I think we should panic here if we can't embed assets in API binary.

api/test/shell/cli_test.sh
```diff
@@ -118,45 +118,45 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
-$workDir/manager-api stop
-sleep 6
-cd -
-rm -rf html
-
-if [[ $res != "hi~" ]]; then
-    echo "failed: manager-api cant run on a different path"
-    exit 1
-fi
+# workDir=$(pwd)
```
starsz, line 121: I think we can remove them if we don't use them.

starsz:
As we talked about the issue. https://github.com/apache/apisix-dashboard/issues/1391#issuecomment-824492523

> provide a flag to use the embed or the external assets in startup?

We need a flag to use the embed or the external assets in a startup.

Makefile
bisakhmondal: Why I wrote such horrible logic🤣🤣
Changing ASAP

bisakhmondal: Why I wrote such horrible logic🤣🤣
Changing ASAP

api/internal/route.go
bisakhmondal: Okay!! I thought users could opt to operate through CLI even without the UI, so skipped intentionally.
changing..

api/test/shell/cli_test.sh
bisakhmondal: Cool

bisakhmondal:
> As we talked about the issue. [#1391 (comment)](https://github.com/apache/apisix-dashboard/issues/1391#issuecomment-824492523)
> 
> > provide a flag to use the embed or the external assets in startup?
> 
> We need a flag to use the embed or the external assets in a startup.

You mean, a flag on manager-api itself (cobra commands), right?

starsz:
> > As we talked about the issue. [#1391 (comment)](https://github.com/apache/apisix-dashboard/issues/1391#issuecomment-824492523)
> > > provide a flag to use the embed or the external assets in startup?
> > 
> > 
> > We need a flag to use the embed or the external assets in a startup.
> 
> You mean, a flag on manager-api itself (cobra commands), right?

No. I think a flag on `Makefile` would be better.

nic-chen:
hi @bisakhmondal  please sync code form branch master

juzhiyuan:
and the conflicts too :)

Makefile
bisakhmondal: fixed

api/internal/route.go
bisakhmondal: fixed

bisakhmondal:
Hi guys,
regarding embedding the static lua codes into the binary,

this is the method that handles the conversion of lua code, but I am not getting how this is done!!
https://github.com/apache/apisix-dashboard/blob/a59e8eff7ecdf38c197b4678c34d2fd15f364f72/api/internal/handler/route/route.go#L271-L306

The documentation in `gopher-lua` is not that good.
Could anyone give me some details?

It seems pointing to the directory is enough, but after the embedding, it's going to be an **in-memory filesystem**.
I haven't found such apis in gopher-lua, could you please guide me?
Thanks.

nic-chen:
> Hi guys,
> regarding embedding the static lua codes into the binary,
> 
> this is the method that handles the conversion of lua code, but I am not getting how this is done!!
> https://github.com/apache/apisix-dashboard/blob/a59e8eff7ecdf38c197b4678c34d2fd15f364f72/api/internal/handler/route/route.go#L271-L306
> 
> The documentation in `gopher-lua` is not that good.
> Could anyone give me some details?
> 
> It seems pointing to the directory is enough, but after the embedding, it's going to be an **in-memory filesystem**.
> I haven't found such apis in gopher-lua, could you please guide me?
> Thanks.

Are there any related error logs?

bisakhmondal:
>Are there any related error logs?

Hi @nic-chen, sorry I didn't get it.

The snippet I mentioned earlier was the existing code that we are using for Lua to go compiler VM. Till now, I couldn't figure out the workaround as after embedding the dag-to-dua package will be an in-memory filesystem (no way I can make `os.Chdir` and `local dag_to_lua = require 'dag-to-lua'` to import the package  ). 

All I can do is, read the content of the individual script from the directory stored in memory fs as per the current set of apis exposed through the go official `io/fs` package [link](https://golang.org/pkg/io/fs/)

Another way we can tackle this by reading and joining the individual scripts into a monolithic script. what do you think?

I don't know Lua (wishing to give it a try real soon 😄), so it would be great to have some heads up on this topic :)

bisakhmondal:
@nic-chen Something just come up into my mind. Let me know what you think about it.
Since our main goal is to create a single binary, we can tackle the issue (I've described earlier) by this way,

step-1
`during build : `  it involves downloading the lua scripts and integrate it into a single binary.
[now the lua package is in memory whenever the manager api will start]

step-2 
`running manager-api : ` whenever the  binary of manager-api will be loaded into memory, along with jobs including writing pid file, ..., it will have another job that is writing the lua package into temporary directory (`os.TempDir()`).

step-3 
`cleanup during stop: ` Optional.

Now, this approach won't affect the current way we are translating the Lua script.

starsz:
Hi, @bisakhmondal 

>  it will have another job that is writing the lua package into temporary directory (`os.TempDir()`).
I don't get the point of this line, can you give more details about it ?

bisakhmondal:
> Hi, @bisakhmondal
> 
> > it will have another job that is writing the lua package into temporary directory (`os.TempDir()`).
> > I don't get the point of this line, can you give more details about it ?

Sure. Actually because of this
https://github.com/apache/apisix-dashboard/blob/a59e8eff7ecdf38c197b4678c34d2fd15f364f72/api/internal/handler/route/route.go#L280-L296

Line 281: navigating to lua package directory
Line 296: Importing the package from the same directory.

But the issue is with import when the package is in a filesystem that is inside memory/RAM (due to embedding).
So, as a workaround, we can  do something like
1. embed the package. {provides the flexibility of a single binary}
2. when manager-api starts, it will dump the memory content of the package into a temp directory (maybe say /tmp). Now it'll work fine.
What do you think?

nic-chen:
> @nic-chen Something just come up into my mind. Let me know what you think about it.

hi, @bisakhmondal , LGTM. But I think it’s better to put it in the `[WorkDir](https://github.com/apache/apisix-dashboard/blob/master/api/cmd/managerapi.go#L91)`, not in a temporary directory.

bisakhmondal:
Aye Aye, Captain😄

api/internal/route.go
```diff
@@ -91,3 +94,55 @@ func SetUpRouter() *gin.Engine {
...
+
+func serve(urlPrefix string, fss fs.FS) gin.HandlerFunc {
+	fileserver := http.FileServer(&spaFileSystem{http.FS(fss)})
+	if urlPrefix != "" {
+		fileserver = http.StripPrefix(urlPrefix, fileserver)
+	}
+	return func(c *gin.Context) {
+
+		if c.Request.Method == "GET" &&
+			(exists(urlPrefix, c.Request.URL.Path, &fss) || !isBlacklisted(c.Request.URL.Path)) {
```
bisakhmondal, line 124: the `exists` method and `isBlacklisted` are doing the same work in different ways. Please let me know which one should be perfect.

bisakhmondal:
Hi, the tasks are done.
The PR is ready to be reviewed.

api/build.sh
```diff
@@ -32,13 +35,13 @@ set -x
 export ENV=local
 pwd=`pwd`
 
-rm -rf output && mkdir -p output/conf && mkdir -p output/dag-to-lua
+rm -rf output && mkdir -p output/conf && mkdir -p ./api/cmd/dag-to-lua
```
nic-chen, line 38: oh, why change `output` to `api/cmd` ?

api/cmd/html/index.html
```diff
@@ -0,0 +1,36 @@
+<!--
```
nic-chen, line 1: we should remove this file. an build it form source when building a binary.

api/internal/route.go
nic-chen: I don't quite understand why these two functions are the same. .

api/test/shell/cli_test.sh
```diff
@@ -117,47 +117,8 @@ if [[ `grep -c "INFO" ./error.log` -eq '0' ]]; then
...
-sleep 6
-rm -fr $distDir
-
-if [[ $res != "hi~" ]]; then
-    echo "failed: manager-api can't run with -p flag out of the default directory"
-    exit 1
-fi
-cd $workDir && git checkout conf/conf.yaml
-
 # test start info
```
nic-chen, line 122: we should add test cases for changes.

api/internal/utils/pid.go
```diff
@@ -30,16 +30,14 @@ func WritePID(filepath string) error {
 		return fmt.Errorf("instance of Manager API already running: a pid file exists in %s", filepath)
 	}
 	pid := os.Getpid()
-	f, err := os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC|os.O_CREATE, 0600)
+	f, err := os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
starsz, line 33: Just want to know why we should change it.

bisakhmondal: Hi, actually the `O_CREATE` flag had multiple occurrences :)

api/build.sh
```diff
@@ -22,6 +22,9 @@ GITHASH=$(cat ./.githash 2> /dev/null || HASH="ref: HEAD"; while [[ $HASH == ref
 
 GOLDFLAGS="-X github.com/apisix/manager-api/internal/utils.version=${VERSION} -X github.com/apisix/manager-api/internal/utils.gitHash=${GITHASH}"
 
+# Ensuring clean state
+# rm -rf api/cmd/dag-to-lua && git checkout  api/cmd/dag-to-lua
```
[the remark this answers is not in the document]
bisakhmondal: Yeah sure

api/internal/route.go
bisakhmondal: Actually its all because Single Page Application strategy followed by React.js. As there is a single `index.html` and each time for a different route the sameDOM gets updated.

Actually for web server the `exists` methods makes sense, as the serving is acted as a middleware, obviously we wouldn't wish to shove HTML content for each request, some condition checking must be there. `exists` check is the required resource file (got through the request path) is present in the HTML directory, if yes, send the file and terminate the request from further propagation, else pass the request to next middleware.

This was okay until I checked with direct route pasted on the browser for eg, (localhost:9000/route/list), as it was not propagated from index.html the `exists` didn't work. But rest of the cases from login to page visits, logout everything was fine.

So I came up with `isBlacklisted`, if the route prefix is not HTML blacklisted for example `/admin` or `/ping` and the request type is `GET` serve the static HTML.

api/build.sh
bisakhmondal: for two reasons :) 
1. we cant embed an empty directory.
2. the embedded directory must be inside the package directory or subdirectory. for `// go:embed` to works.
>Go source files that import "embed" can use the //go:embed directive to initialize a variable of type string, []byte, or FS with the contents of files read from the package directory or subdirectories at compile time.


[reference](https://golang.org/pkg/embed/)

api/cmd/html/index.html
bisakhmondal: same as https://github.com/apache/apisix-dashboard/pull/1788#discussion_r634528665

As we have to provide an option not to embed the web build. We can't leave an empty directory.

api/test/shell/cli_test.sh
bisakhmondal: Yes, I can add tests for when the web build is not embedded, as comparing less content is easy.
But when web build is embedded, I'm not sure how to do it.
Do you have a plan?

api/cmd/html/index.html
nic-chen: not to embed empty dir, build `html` before embed it

bisakhmondal: hi, are you sure? Won't that be time-consuming during testing and development?
BTW, the file gets replaced by the actual build when we issue a full build.
let me know what you think.

nic-chen: Sure. Maybe we could add a parameter, `make api-run` does not need to embed, `make build` does.

bisakhmondal: Yes, that's exactly what I have added. 
`make build` full build with the web embed.
`make api-run` run without embed.
[new] `make api-build` build(just the manager-api) without embed.

Now as you see, we have 2 cases when we are not building web assets. maybe as a substitute for this current `index.html`, we can create an empty directory `html` and an empty file, maybe `keep.keep`. This would work fine. But I thought, instead of an empty file an HTML with a suitable message (`The build is partial`) would be appropriate.
So should we go with the other option just I mentioned?

starsz (commented):
There is an empty file `api/cmd/dag-to-lua/keep`.

starsz:
Now, we don't have a method that doesn't embed assets in API binary?

bisakhmondal:
> Now, we don't have a method that doesn't embed assets in API binary?

yes, the good old `make api-run` doesn't embed web assets and lua scripts.
And newly introduced `make api-build` creates a build without embedding web assets.



> There is an empty file `api/cmd/dag-to-lua/keep`.

Yes, that's an ongoing discussion, please share your views,
https://github.com/apache/apisix-dashboard/pull/1788#discussion_r634528665
https://github.com/apache/apisix-dashboard/pull/1788#discussion_r634248776
thanks :)

api/cmd/html/index.html
bisakhmondal: Just a thought, How about adding the dag-to-lua package as a git submodule. instead of downloading and unzipping.

nic-chen: > Yes, that's exactly what I have added.

so why do you think that will be time-consuming?

We cannot embed old files, but files built from the current source.

bisakhmondal: Oh, actually I thought for every build, you wish to build HTML
> We cannot embed old files, but files built from the current source.

Yes the current code also ensures that too[ref](https://github.com/apache/apisix-dashboard/pull/1788/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R46-R49)
`	rm -rf api/cmd/html && git checkout api/cmd/html/ && api/build.sh && mkdir -p output/logs`

The `rm -rf` and `checkout` is helpful because I have excluded all build files except `index.html` from the HTML dir.

To be honest, the static index.html serves two purpose
1. preventing the error from go embed the empty directory.
2. Instead of filling the directory with some garbage file, an HTML with a message seems reasonable. As anyways it will get replaced by web build's index.html.

Let me know what changes should I make

PR #1795, opened 2021-04-19
ci: cancel duplicate workflow to reduce CI queue time

Yiyiyimu:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [x] None of above

The Github Actions - cancel workflow have already been applied in apisix and it works well, so try to also import it here.

> We always wait for CI queue time, and we always trying to reduce waiting time. Here is another method to achieve it.
> 
> The method is that whenever a new commit added to PR, the previous running CI could be canceled since it is of no use. This could be achieved by using Github API, find the workflow on the same branch but with different SHA. And we could achieve this with [potiuk/cancel-workflow-runs](https://github.com/potiuk/cancel-workflow-runs) which is also [used in Apache Airflow](https://github.com/apache/airflow/blob/master/.github/workflows/build-images-workflow-run.yml).
> 
> To bypass the limit of third party Actions, [the recommendation provided by Apache Infra](https://cwiki.apache.org/confluence/display/BUILDS/GitHub+Actions+status#GitHubActionsstatus-Security) is to use submodules.
> 
> Since the "cancel" workflow could only be triggered after merged, so I could not test it in this PR. Thus I duplicate the same situation in my own repo, and it works as expect. See https://github.com/Yiyiyimu/GithubActionsTest/pull/5 you could find previous running workflow got cancelled when a new one coming in. Also see https://github.com/Yiyiyimu/GithubActionsTest/runs/2344066993?check_suite_focus=true for cancellation detail.

**Checklist:**


.github/workflows/workflow-cancel.yaml
```diff
@@ -0,0 +1,29 @@
+name: Cancelling Duplicates
+on:
+  workflow_run:
+    workflows:
+      - 'Frontend e2e test'
```
juzhiyuan, line 5: Very Cool!!!

juzhiyuan (approved):
Cool!!!!

Just ignored the Netlify Error

LiteSun (approved):

nic-chen (approved):

Issue #1803, opened 2021-04-20
Question: is it a bug that users can delete all target nodes forms in upstream create page?


ericsyh:
## Issue description

When fixing the Upstream form warning message issue, i found a behavior that users can delete all target nodes forms on the page like 

![](http://piw45adv3.bkt.clouddn.com/%E6%8D%95%E8%8E%B7.PNG)

But this upstream can be created successfully. 

![](http://piw45adv3.bkt.clouddn.com/%E6%8D%95%E8%8E%B72.PNG)

## Expected behavior

Not sure is it a expected behavior.

## How to Reproduce

1. Go to 'Upstream Page'
2. Click on 'Create Button'
3. Delete all targets nodes forms

## Environment

* apisix version (cmd: `apisix version`): the master branch
* OS (cmd: `uname -a`):
* OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
* etcd version, if have (cmd: run `curl http://127.0.0.1:9090/v1/server_info` to get the info from server-info API):
* apisix-dashboard version, if have:
* Browser version, if have:

liuxiran:
Hi @ericsyh thanks for your report.

Acctually from the funcitonal point of view, the empty node list is a feature, see:

 `The nodes can be empty, which means it is a placeholder and will be filled later. Clients use such an upstream will get 502 response.`
(come from https://github.com/apache/apisix/blob/master/docs/en/latest/admin-api.md#upstream)

But from the UE point of view, an upstream can not be used with the empty node list, so keep at least one node in the node list is better.

How about add this restriction to the dashboard? 
Let's wait for others option @juzhiyuan @LiteSun @nic-chen @imjoey

juzhiyuan:
As @liuxiran said, we actually support an empty node list, it will return a 5xx status code to the client. I would prefer keeping the current behaviour, because if users really want to do it, what they can do on the dashboard?

nic-chen:
+1 for  keeping the current behavior

liuxiran:
@juzhiyuan @nic-chen  keep the current behavior is also reasonable for me

If there is no objection, we will close the issue

cc @ericsyh

ericsyh:
ok for close the issue. But i am quite interesting in what scenario that users will create a upstream with empty node list?

juzhiyuan:
> ok for close the issue. But i am quite interesting in what scenario that users will create a upstream with empty node list?

[...]

PR #1812, opened 2021-04-21
fix: can not configure upstream with no nodes

liuxiran:

- [X] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

During rechecking #1803 , there is a problem occured when I try to configure a upstream with no nodes.

Test steps:

1. Create an upstream with no nodes
2. Configure this upstream with no changes, an Submit the data, the error occured

2021-04-21 15-32-32屏幕截图

This pr tried to change the fe logic to solve this problem.

Doc: 
Delete `k8s_deployment_info` in upstream just refer to https://github.com/apache/apisix/blob/359e0906c5a3f0149c0f6a44aff927d4c4eb8c6b/CHANGELOG.md#220

TODO

- [ ] add test case

**Related issues**

None

**Checklist:**

- [X] Did you explain what problem does this PR solve? Or what new features have been added?

### DIFF
--- a/api/internal/core/entity/format_test.go
+++ b/api/internal/core/entity/format_test.go
@@ -101,3 +101,53 @@ func TestNodesFormat_Map(t *testing.T) {
 	assert.Contains(t, jsonStr, `"port":8080`)
 	assert.Contains(t, jsonStr, `"host":"127.0.0.1"`)
 }
+
+func TestNodesFormat_empty_struct(t *testing.T) {
+	// route data saved in ETCD
+	routeStr := `{
+		"uris": ["/*"],
+		"upstream": {
+			"type": "roundrobin",
+			"nodes": []
+		}
+	}`
+
+	// bind struct
+	var route Route
+	err := json.Unmarshal([]byte(routeStr), &route)
+	assert.Nil(t, err)
+
+	// nodes format
+	nodes := NodesFormat(route.Upstream.Nodes)
+
+	// json encode for client
+	res, err := json.Marshal(nodes)
+	assert.Nil(t, err)
+	jsonStr := string(res)
+	assert.Contains(t, jsonStr, `[]`)
+}
+
+func TestNodesFormat_empty_map(t *testing.T) {
+	// route data saved in ETCD
+	routeStr := `{
+		"uris": ["/*"],
+		"upstream": {
+			"type": "roundrobin",
+			"nodes": {}
+		}
+	}`
+
+	// bind struct
+	var route Route
+	err := json.Unmarshal([]byte(routeStr), &route)
+	assert.Nil(t, err)
+
+	// nodes format
+	nodes := NodesFormat(route.Upstream.Nodes)
+
+	// json encode for client
+	res, err := json.Marshal(nodes)
+	assert.Nil(t, err)
+	jsonStr := string(res)
+	assert.Contains(t, jsonStr, `[]`)
+}
--- a/api/internal/core/entity/format_test.go
+++ b/api/internal/core/entity/format_test.go
@@ -151,3 +151,29 @@ func TestNodesFormat_empty_map(t *testing.T) {
 	jsonStr := string(res)
 	assert.Contains(t, jsonStr, `[]`)
 }
+
+func TestNodesFormat_no_nodes(t *testing.T) {
+	// route data saved in ETCD
+	routeStr := `{
+		"uris": ["/*"],
+		"upstream": {
+			"type": "roundrobin",
+			"service_name": "USER-SERVICE",
+			"discovery_type": "eureka"
+		}
+	}`
+
+	// bind struct
+	var route Route
+	err := json.Unmarshal([]byte(routeStr), &route)
+	assert.Nil(t, err)
+
+	// nodes format
+	nodes := NodesFormat(route.Upstream.Nodes)
+
+	// json encode for client
+	res, err := json.Marshal(nodes)
+	assert.Nil(t, err)
+	jsonStr := string(res)
+	assert.Contains(t, jsonStr, `null`)
+}
--- a/web/src/components/Upstream/service.ts
+++ b/web/src/components/Upstream/service.ts
@@ -103,10 +103,14 @@ export const convertToRequestData = (
     return undefined
   }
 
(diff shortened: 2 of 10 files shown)


web/src/components/Upstream/service.ts
```diff
@@ -68,7 +68,6 @@ export const convertToRequestData = (
     type,
     hash_on,
     key,
-    k8s_deployment_info,
```
juzhiyuan, line 71: yes, we don't support k8s_deployemnt_info here

imjoey:
@liuxiran

How about we make it in the backend? Please see https://github.com/apache/apisix-dashboard/blob/master/api/internal/core/entity/format.go#L49, we can initialize the `nodes` as an empty slice instead of a nil one. Just change `var nodes []*Node` to `nodes := make([]*Node, 0)` .

Looking forward to your insights.

liuxiran:
> @liuxiran
> 
> How about we make it in the backend? Please see https://github.com/apache/apisix-dashboard/blob/master/api/internal/core/entity/format.go#L49, we can initialize the `nodes` as an empty slice instead of a nil one. Just change `var nodes []*Node` to `nodes := make([]*Node, 0)` .
> 
> Looking forward to your insights.

@imjoey Thanks for your advice, and it really make sense for me.

before this change, when get info about an upstream with no nodes
fe will get
WechatIMG5


actually `nodes:null` causes the problem

after this change, fe will get
WechatIMG4

which is matches expectation.

BTW, I have a question in the process of debugging:
there are some logs I added to watch the changes of `upstream.Nodes` before the advice
WechatIMG7
I found even if I did not do this change, the `upstram.Nodes` will always be `[]`, but what Fe got is different, I haven't thought about it for a long time. 

Also @nic-chen @Jaycean @starsz to take a look, thanks a lot

starsz:
> > @liuxiran
> > How about we make it in the backend? Please see https://github.com/apache/apisix-dashboard/blob/master/api/internal/core/entity/format.go#L49, we can initialize the `nodes` as an empty slice instead of a nil one. Just change `var nodes []*Node` to `nodes := make([]*Node, 0)` .
> > Looking forward to your insights.
> 
> @imjoey Thanks for your advice, and it really make sense for me.
> 
> before this change, when get info about an upstream with no nodes
> fe will get
> WechatIMG5
> 
> actually `nodes:null` causes the problem
> 
> after this change, fe will get
> WechatIMG4
> 
> which is matches expectation.
> 
> BTW, I have a question in the process of debugging:
> there are some logs I added to watch the changes of `upstream.Nodes` before the advice
> WechatIMG7
> I found even if I did not do this change, the `upstram.Nodes` will always be `[]`, but what Fe got is different, I haven't thought about it for a long time.
> 
> Also @nic-chen @Jaycean @starsz to take a look, thanks a lot

Hi, @liuxiran 
In golang, you should distinguish `nil slice` and `empty slice`

```
var x []*entity.Node
```
is a nil slice.

```
y := make([]*entity.Node, 0)
```
is an empty slice. The length and value are 0.
If you use printf or println to print them, the result is both "[]".

In this situation, because the upstream.Nodes' type is interface{}.So it's a little more complex.
You can use 
```
reflect.ValueOf(upstream.Nodes).IsNil()
```
to see the true value.

liuxiran:
> Hi, @liuxiran
> In golang, you should distinguish `nil slice` and `empty slice`
> 
> ```
> var x []*entity.Node
> ```
> 
> is a nil slice.
> 
> ```
> y := make([]*entity.Node, 0)
> ```
> 
> is an empty slice. The length and value are 0.
> If you use printf or println to print them, the result is both "[]".
> 
> In this situation, because the upstream.Nodes' type is interface{}.So it's a little more complex.
> You can use
> 
> ```
> reflect.ValueOf(upstream.Nodes).IsNil()
> ```
> 
> to see the true value.

@starsz Thanks for your detailed explanation, after added `reflect.ValueOf(upstream.Nodes).IsNil()` and debug again, I find out the confusion before:
debug step:
1. create an upstream with no nodes
2. visit uptream page, call `GET /apisix/admin/upstreams`
3. config the upstream ,call `GET /apisix/admin/upstream/:id`

before the advice change, I got the logs:
2021-04-23 14-32-00屏幕截图

after the advice change, I got the logs:
2021-04-23 14-39-50屏幕截图

The conclusion is that: 
the original `upstream.nodes` type got from etcd is `[]interface{}`, and it is an empty slice, 
it will be changed to type nodes , `[]*Node`, in FormatNodes `case []interface{}` when I visit upstream page(step2), 
and  the when I got the upstream info(step3), the data come from catch instead of stroge, that is to say, in this scene, the `upstream.nodes` type is already `[]*Node`.

so the advice will solve the problem. 
Thanks again for all your help, then I will try to add related test cases

juzhiyuan (approved):

api/internal/core/entity/format_test.go
```diff
@@ -101,3 +101,53 @@ func TestNodesFormat_Map(t *testing.T) {
...
+
+	// nodes format
+	nodes := NodesFormat(route.Upstream.Nodes)
+
+	// json encode for client
+	res, err := json.Marshal(nodes)
+	assert.Nil(t, err)
+	jsonStr := string(res)
+	assert.Contains(t, jsonStr, `[]`)
+}
```
starsz, line 153: Need to test without the "nodes" field.

starsz (commented):
Backend LGTM.

api/internal/core/entity/format_test.go
liuxiran: done.

LiteSun (approved):
FE LGTM.

nic-chen (approved):

imjoey (approved):
/lgtm

imjoey:
@liuxiran FE ci failed after this pr got merged. Could you please take a look, please refer to https://github.com/apache/apisix-dashboard/runs/2430236911. Thanks.

liuxiran:
> @liuxiran FE ci failed after this pr got merged. Could you please take a look, please refer to https://github.com/apache/apisix-dashboard/runs/2430236911. Thanks.

got it, I will fix it later, thanks

Issue #1931, opened 2021-06-04
feat: migrate to viper library for config manage


bzp2010:
# Feature request

## Please describe your feature

At present, the original yaml file is directly read and unmarshal mode is used for configuration management. If a new configuration item is added, multiple codes will be modified, and if the old configuration is changed, it will be incompatible.

## Describe the solution you'd like

Use `github.com/spf13/viper` to manage configurations. (It uses MIT license, we should be able to import it.)

It supports the use of yaml and JSON format configuration files, or configuration based on environment variables. At the same time, it can also provide etcd based configuration management function.

## Improvement Roadmap

1. First, only change `conf.go` to unmarshal the configuration file into the structure pointer by viper, which will not destroy the compatibility of the old program.

2. Then, use Viper to get the configuration file in the newly written code.

3. Finally, gradually modify the old code to read the configuration code.

PR #1932, opened 2021-06-04
feat: split cmd package code

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

The current program instructions are written in the same file in the CMD package, and the code structure is not clear. In this PR, I split the `managerapi.go` file.

A little bit of an update has been introduced. (`manager-api -v` and `manager-api version` to get version of Manager API)

**Changes**

1. split `managerapi.go` to multiple files in `cmd` packages
2. move `./cmd/manager/main.go` to `./main.go`
3. change github workflows and test script's golang source path
4. optimize imported libraries

**Related issues**

prechange #1931 

**Checklist:**


api/go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/apisix/manager-api
 
-go 1.13
+go 1.16
```
tokers, line 3: IMHO We should not change the lowest GO version, do you use any new features in Go/1.16?

api/cmd/root.go
```diff
@@ -42,30 +42,23 @@ import (
...
-	fmt.Fprintf(os.Stdout, "%-8s: %s\n", "Loglevel", conf.ErrorLogLevel)
-	fmt.Fprintf(os.Stdout, "%-8s: %s\n\n", "Logfile", conf.ErrorLogPath)
-}
-
-func printVersion() {
-	fmt.Fprintf(os.Stdout, "%-8s: %s\n", "Version", Version)
-	fmt.Fprintf(os.Stdout, "%-8s: %s\n", "GitHash", GitHash)
+var rootCmd = &cobra.Command{
+	Use:   "manager-api [flags]",
+	Short: "Apache APISIX Manager API",
```
tokers, line 50: We may don't need the RunE handler as we put all functions into a bunch of subcommands.

api/go.mod
bzp2010: This is an unexpected change. Maybe due to some functions of IDE, I will change it back.

api/cmd/root.go
bzp2010: This part completely refers to the code segment of the original root command and has not been modified. Should I delete it?

iamayushdas (approved):

api/cmd/root.go
bzp2010: @tokers 
Here I create a new `serve` command. It allows us to run the Manager API directly from the command line, rather than in the form of system services. This is similar to the original root command.
The `RunE` part of the root command has been removed.

bzp2010:
### Change progress
- [x] split `managerapi.go` to multiple files in `cmd` packages
- [x] move `./cmd/manager/main.go` to `./main.go`
- [x] change github workflows and test script's golang source path
- [x] optimize imported libraries

### About root command's `RunE` (Indicated by the review of @tokers)
Due to various tests, the deployment script still relies on the `RunE` code in the root command. I restored some of the changes, such as the added `serve` command. For the following reasons.

> When I read some of the command code introduced by the previous PR, I realized that when we introduce a new command, we should design it carefully. Such as the built-in daemon code, it uses some syscalls and destroys the compatibility of the operating system.
> 
> (When using Windows, we cannot use `syscall.Kill`. Although we may not use Windows in the deployment environment, we still need to ensure the operating system compatibility provided by golang as much as possible.)
>
> By the way, I think it's necessary to use `systemd` or `supervisor` to implement operating system daemons instead of built-in command, which brings unnecessary complexity.

### Final things
Currently, only a few updates have been made to PR, and no incompatible updates have been introduced.

membphis:
ping @nic-chen

api/cmd/remove.go
```diff
@@ -0,0 +1,38 @@
...
+	Short: "remove Apache APISIX Dashboard service",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		serviceState.removeService = true
+		status, err := service.manageService()
+		fmt.Println(status)
+		return err
+	},
+}
+
+func init() {
```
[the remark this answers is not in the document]
bzp2010: @nic-chen fixed

nic-chen (approved):

tokers (approved):

starsz (approved):

PR #1946, opened 2021-06-21
refactor: migrate to viper configure manage

bzp2010:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

The configuration management is implemented by `viper` library instead of reading yaml file directly.
It will provide support for YAML, JSON, TOML and other configuration file formats, and even remote loading through etcd, which will be controlled by `-c` or `--config` parameters.

And no incompatible updates introduced.

**Related issues**

part of #1931

**Checklist:**

- [x] Did you explain what problem does this PR solve? Or what new features have been added?
- [x] Have you added corresponding test cases?
- [x] Have you modified the corresponding document?
- [x] Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

PR #1947, opened 2021-06-22
fix: logger on Windows

bzp2010:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

The following errors will appear when the current Manager API is running on windows.

> `panic: couldn't open sink "C:\\apisix-dashboard\\api\\logs\\error.log": no sink found for scheme "c"`

`uber-go/zap` can't directly handle the windows path starting with, for example, `C:\`. This PR adds a new scheme processor, which supports accessing log files on Windows system by using `winfile:///C:\`. At the same time, the program will automatically add `winfile` when generating the absolute path on Windows.

**Checklist:**


iamayushdas (approved):

juzhiyuan (approved):

nic-chen (approved):
Please test manually to confirm that there is no problem, since CI does not support Windows currently.

bzp2010:
@nic-chen 
> Please test manually to confirm that there is no problem, since CI does not support Windows currently.

I found that when I remove the code in `conf.go` to get the absolute path of the file, I can directly use the relative path to write the log on windows.

But in order to be compatible with the file path printed in `printInfo`, I added some compromise code, such as `ErrorLogPathAbs`. At the same time, I will refactor the configuration module in another PR (#1946) to solve these problems.

api/internal/conf/conf.go
```diff
@@ -191,22 +192,35 @@ func setConf() {
...
-
-		// access log
-		if config.Conf.Log.AccessLog.FilePath != "" {
-			AccessLogPath = config.Conf.Log.AccessLog.FilePath
+			if runtime.GOOS == "windows" {
+				ErrorLogPath = `winfile:///` + ErrorLogPath
+			}
 		}
 		if !filepath.IsAbs(AccessLogPath) {
+			if strings.HasPrefix(AccessLogPath, "winfile") {
```
tokers, line 214: AFAIK the go package `path` handles the windows path normally. Why just use functions provided by it? Like `path.Abs`.

bzp2010: - First of all, the function of getting the absolute path of a file is the original code, and I don't fully understand why I do it.
After my test, I found if I don't get the absolute path here, I can't test it pass on Linux. But using relative path can work normally on Windows.

- The purpose of detecting winfile prefix here is to define absolute path on Windows system in `conf.yaml`. `uber-go/zap` cannot handle paths that start with Windows drive letters such as `C:\`.

tokers (approved):

PR #1956, opened 2021-06-26
refactor: server manager

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Refactoring the management code of Manager API to help to initialize, start and stop. Reduce redundant code and improve readability.

I created a new `server` module in `internal/core`. 
It will help run according to the process of `NewServer()`->`server.init()`->`Start()`->`Stop()`.
It covers the functions of initializing etcd client, initializing store, initializing route and starting HTTP server.

**Checklist:**

### DIFF
--- a/api/internal/core/server/server.go
+++ b/api/internal/core/server/server.go
@@ -44,29 +44,29 @@ func NewServer(options *Options) (*server, error) {
 	return &server{options: options}, nil
 }
 
-func (s *server) Start() error {
+func (s *server) Start(errSig chan error) {
 	// initialize server
 	err := s.init()
 	if err != nil {
-		return err
+		errSig <- err
+		return
 	}
 
 	// write daemon pid file
 	err = s.writePID()
 	if err != nil {
-		return err
+		errSig <- err
+		return
 	}
 
 	// print server info to stdout
 	s.printInfo()
 
-	// For internal error handling across multiple goroutines.
-	errSig := make(chan error, 1)
-
 	// start HTTP server
 	log.Infof("The Manager API is listening on %s", s.server.Addr)
 	go func() {
-		if err := s.server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		err := s.server.ListenAndServe()
+		if err != nil && err != http.ErrServerClosed {
 			log.Errorf("listen and serv fail: %s", err)
 			errSig <- err
 		}
@@ -81,14 +81,6 @@ func (s *server) Start() error {
 				errSig <- err
 			}
 		}()
-	}
-
-	// handle HTTP(s) server error
-	select {
-	case err := <-errSig:
-		return err
-	default:
-		return nil
 	}
 }
 
--- a/api/internal/core/server/server.go
+++ b/api/internal/core/server/server.go
@@ -1,0 +1,152 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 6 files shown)


iamayushdas (approved):

api/internal/core/server/server.go
```diff
@@ -0,0 +1,152 @@
...
+			if err != nil && err != http.ErrServerClosed {
+				log.Errorf("listen and serve for HTTPS failed: %s", err)
+				errSig <- err
+			}
+		}()
+	}
+
+	// handle HTTP(s) server error
+	select {
+	case err := <-errSig:
```
tokers, line 88: Frankly, the way to detect whether these servers were started is not reliable, as it depends on the behavior the go runtime, like the scheduler. These two goroutines might not get the change to run when the `select` statement is run.

api/internal/core/server/server.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (s *server) init() error {
+	log.Info("Initialize Manager API store")
+	err := s.setupStore()
+	if err != nil {
+		return err
+	}
+
+	log.Info("Initialize Manager API server")
+	s.setupHTTPServer()
```
tokers, line 110: Only setup the HTTP server is enough?

bzp2010: In fact, it initializes routes, HTTP(s) server and any other content related to HTTP API. I don't have a better idea of naming it.
Do you have any suggestions?

tokers: Then the name is confusing, what about just `s.setupServers`?

bzp2010: emmm, If it's `s.setupServers`, it looks like it's initializing the entire `Manager API server`.

api/internal/core/server/server.go
bzp2010: Please forgive me for not knowing the relevant content of go runtime. Can you talk more deeply or have related document content?
I search about some related questions, and the solution given is just like what these codes do now.

> https://stackoverflow.com/questions/26090301/run-both-http-and-https-in-same-program/29468115

tokers: Just like I said:

> These two goroutines might not get the change to run when the select statement is run.

And since the `select` statement is non-block, the select run happens before the two servers try to send error message (if errors really occurred).

starsz: Yes, here we can't get err absolutely even if the error occurs.
Maybe we can return a channel in this function.

api/internal/core/server/server.go
tokers: Or set up API.

api/internal/core/server/server.go
bzp2010: Yes, I understand the problems here.

Therefore, I try to change the `default case` to a 5 seconds timer. The listen operation of most devices should be completed within 5 seconds, so the current select will block start for 5 seconds, but it will not affect the HTTP server in goroutine. If the error of `Listen` does not return within 5 seconds, it is considered that no error has occurred and the timer returns `nil` error.

api/internal/core/server/server.go
bzp2010: I think it's good

api/internal/core/server/server.go
bzp2010: **Update:** Yes, I have modified it in this way.

starsz (approved):

nic-chen (approved):

bzp2010:
Hi, @tokers 
If you have free time, can you review it? I have fixed the previous problems. 
Thanks!

tokers (approved):

Issue #1900, opened 2021-05-16
Enable Husky to check in pre-commit


juzhiyuan:
It would be great to have a checker here:
1. We could enable Husky to check pre-commit;
2. When changed files contain local cache paths, like `@/.umi/xx`, just throw errors.

_Originally posted by @juzhiyuan in https://github.com/apache/apisix-dashboard/issues/1899#issuecomment-841829529_

liuxiran:
closed with #1957

PR #1957, opened 2021-06-26
feat: added husky for pre-commit

### LINKED ISSUE #1746: support pre-commit
# Feature request

Hi, I noticed that there have something that could be automatically done before committing:

- [ ] Auto prettier;
- [ ] Auto check commit message;

I would prefer using [husky](https://www.npmjs.com/package/husky) to do this.

## Additional context

Due to we put files under the `web` folder (non-root), husky doesn't support this way months ago, please check if it supports it now.

iamayushdas:

- [x] auto prettier
- [ ] auto-check commit message

- [ ] Bugfix
- [x] New feature provided
- [x] Improve performance
- [ ] Backport patches

Added auto prettier via husky 

**Related issues**

fix/resolve #1746 #1900 

**Checklist:**


web/src/global.tsx
```diff
@@ -25,8 +25,8 @@ import defaultSettings from '../config/defaultSettings';
 
 // see https://github.com/suren-atoyan/monaco-react/issues/168
 loader.config({ paths: { vs: '/monaco-editor/min/vs' } });
-if (getLocale() === 'zh-CN'){
-  loader.config({ "vs/nls": { availableLanguages: { "*": 'zh-cn' } } });
+if (getLocale() === 'zh-CN') {
```
iamayushdas, line 28: staged this file to check the husky

iamayushdas:
cc @juzhiyuan @guoqqqi @liuxiran

juzhiyuan:
Hi, I just checked this PR, and your solution is creating a Husky under the root folder, actually, we only want to format the `web/` folder 🤔 

Do we have some reference or examples using this way?

iamayushdas:
> Hi, I just checked this PR, and your solution is creating a Husky under the root folder, actually, we only want to format the `web/` folder 🤔 
> 
> 
> 
> Do we have some reference or examples using this way?

https://typicode.github.io/husky/#/?id=custom-directory

juzhiyuan:
that's fine, but this PR put package.json under the root folder, right?

I think it should be at the `web` folder just like this example:

https://typicode.github.io/husky/#/?id=custom-directory

```json
// package.json
{
  "scripts": {
    "prepare": "cd .. && husky install front/.husky"
  }
}
```

juzhiyuan:
image

iamayushdas:
If we install something in some package (eg: husky) it will create the package.json, and also IMO that package.json with script consisting in it is the same package.json that will produce while installing husky

iamayushdas:
Okay okay, i guess i understood you now, i am working on it as you want

iamayushdas:
cc @juzhiyuan check it again please

juzhiyuan (approved):

juzhiyuan:
cc @bzp2010 @liuxiran @LiteSun

guoqqqi (approved):

liuxiran:
I tried in my local env, and found that when I tried to commit a `js` file e.g. cypress test files or `.eslintrc.js`, an error will occur:

image

That is because we did not put these js files into `tsconfig.json` include list, or we can also choose to ignore them

I suggest to put `cypress` into include list, as for other config files like `.eslintrc.js`, we can put them into `.eslintignore` file

what do you think?

cc @juzhiyuan @guoqqqi @iamayushdas @LiteSun

iamayushdas:
I agree @liuxiran , it is important for tests, thanks for review

LiteSun:
Agree to put cypress into include list.

iamayushdas:
@liuxiran @LiteSun check once again, is it fine now?

iamayushdas:
search_ route.spec.js test is failing

guoqqqi:
I think we should add some assertions to these open drawer tests to make them as stable as possible.

web/.eslintignore
```diff
@@ -1,4 +1,5 @@
 /lambda/
 /scripts
 /config
-.history
\ No newline at end of file
+.history
+.eslintrc.js
```
liuxiran, line 5: empty line

liuxiran:
image

we can try to import `uuid` which is already one of our dependencies to  generate the `uid` for the file

iamayushdas:
cc 

> image
> 
> we can try to import `uuid` which is already one of our dependencies to generate the `uid` for the file

we can add this in later PR, is that cool for you

iamayushdas:
cc @juzhiyuan @liuxiran @LiteSun @guoqqqi need you review again and lets get this feature merged,

liuxiran:
> we can add this in later PR, is that cool for you

👏  so cool ~, that's a deal

liuxiran (approved):

juzhiyuan (approved):

iamayushdas:
@nic-chen can i have your review here

web/src/constants.ts
```diff
@@ -28,6 +28,7 @@ export const codeMessage = {
   422: '当创建一个对象时，发生一个验证错误。',
   500: '服务器发生错误，请检查服务器。',
   502: '网关错误。',
+
```
nic-chen, line 31: why change this?

iamayushdas: From that i was just testing husky, it will not harm any code

nic-chen: Yes, but please don't make unnecessary changes.

iamayushdas:
cc @LiteSun

LiteSun (approved):

PR #2002, opened 2021-07-25
feat(plugin): allowing  basic-auth to dynamically adapt to the BE rules

### LINKED ISSUE #1588: Optimize front-end validation of fields
## The current front-end approach
To create a route for example, the front-end will be regular [1] check for certain fields, such as name field, if the regular condition is not met, the front-end page will report an error and block the user's next operation.

image

## The disadvantages of the current approach
The current front-end can meet the verification requirements. However, when the back-end changes the verification rules, the front-end has to passively and manually update the verification rules for the corresponding fields. If the back-end does not notify the front-end in time or misses some fields, then the synchronization of the front-end and back-end verification will be a problem, affecting the user's use.

## Improvement measures
After communicating with @membphis , using jsonschema validation would be a good choice.
The front-end calls the API to get the jsonschema of the corresponding module and then maps the corresponding jsonschema rules to the corresponding field checks on the front-end. Do this when the back-end update the validation rules, the front-end can automatically synchronize the latest validation rules, reducing unnecessary development time of the front-end.

[1] regular expressions provided by the back-end, and then manually added by the front-end developers to the corresponding form form for the field validation

cc @juzhiyuan @liuxiran @guoqqqi

wmdmomo:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

The PR will change the plugin under plugin/UI (basic-auth), allowing basic-auth to dynamically adapt to the back-end json rules.

**Related issues**

fix/resolve #1588 

**Checklist:**


web/src/components/Plugin/UI/basic-auth.tsx
```diff
@@ -20,6 +20,7 @@ import { Form, Input } from 'antd';
 
 type Props = {
   form: FormInstance;
+  schema: Record<string, any> | undefined;
```
juzhiyuan, line 23: Hi, @wmdmomo I couldn't find out where this field is used, could you please have a better description?

wmdmomo: Because for the plug-in basic-auth, its back-end jsonschema is not specifically used, so here is just passing parameters.

liuxiran:
hi @wmdmomo please fix conflict files when you have time, thanks

web/src/components/Plugin/UI/plugin.tsx
```diff
@@ -55,7 +62,11 @@ export const PluginForm: React.FC<Props> = ({ name, schema, renderForm, form })
     case 'api-breaker':
       return <ApiBreaker form={form} />;
     case 'basic-auth':
+<<<<<<< HEAD
```
juzhiyuan: Please remove those lines

wmdmomo: I have removed those.

web/src/components/Plugin/UI/basic-auth.tsx
liuxiran: Hi @WillemJiang , sorry for taking so long to reply.
In fact there are two schema for `basic-auth` plugin, one is the `schema` as you pointed in the Screenshot, this is used for `Route` or `Service`. The other is the `consumer_schema`, this is for `Consumer`. 

image

`basic-auth` form UI is used for `Consumer`, so please use the `consumer_schema`.

wmdmomo: @liuxiran Thank you for your reminder. I have modified it. When the plugin is basic-auth, pass it as the value of consumer_schema.

liuxiran: nice, and then @wmdmomo  can try to dynamically adapt to the required props. which  the only rule for `basic-auth` plugin

wmdmomo: @liuxiran  Sorry for taking so long to reply. I have now verified the required items of the basic-auth plugin.

PR #2010, opened 2021-07-28
feat: new authentication framework

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Implement a new user authentication framework to support a variety of different identity source authentication, and strengthen the security of local users.

Migrate the local authentication to the new user authentication framework, and change the password stored in plaintext to bcrypt hash (cost 10).
Add a `type` field to the user information to indicate the authentication source. However, it can still be used normally without setting the `type` field to be compatible with the usage of older versions.

**Checklist:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,13 +16,12 @@
 #
 FROM alpine:latest as pre-build
 
-ARG APISIX_DASHBOARD_VERSION=master
+ARG APISIX_DASHBOARD_GITHASH=00000000
+
+ADD . /usr/local/apisix-dashboard
 
 RUN set -x \
-    && apk add --no-cache --virtual .builddeps git \
-    && git clone https://github.com/apache/apisix-dashboard.git -b ${APISIX_DASHBOARD_VERSION} /usr/local/apisix-dashboard \
-    && cd /usr/local/apisix-dashboard && git clean -Xdf \
-    && rm -f ./.githash && git log --pretty=format:"%h" -1 > ./.githash
+    && echo ${APISIX_DASHBOARD_GITHASH:0:8} > ./.githash
 
 FROM golang:1.14 as api-builder
 
--- a/api/conf/conf.yaml
+++ b/api/conf/conf.yaml
@@ -67,12 +67,11 @@ authentication:
                         # NOTE: Highly recommended to modify this value to protect `manager api`.
                         # if it's default value, when `manager api` start, it will generate a random string to replace it.
   expire_time: 3600     # jwt token expire time, in second
+  data_source: local    # user data store
   users:                # yamllint enable rule:comments-indentation
-    - type: local
-      username: admin   # username and password for login `manager api`
+    - username: admin   # username and password for login `manager api`
       password: admin
-    - type: local
-      username: user
+    - username: user
       password: user
 
 plugins:                          # plugin list (sorted in alphabetical order)
--- a/api/conf/conf.yaml
+++ b/api/conf/conf.yaml
@@ -70,10 +70,10 @@ authentication:
   users:                # yamllint enable rule:comments-indentation
     - type: local
       username: admin   # username and password for login `manager api`
-      password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC   # password: admin (bcrypt-cost10)
+      password: admin
     - type: local
       username: user
-      password: $2a$10$arSuTLefL6mgw8RcE9.ls.pJmnnE9PAYcbI.3POWKyrPyaIZmbeRm   # password: user (bcrypt-cost10)
+      password: user
 
 plugins:                          # plugin list (sorted in alphabetical order)
   - api-breaker
--- a/api/conf/conf.yaml
+++ b/api/conf/conf.yaml
@@ -71,7 +71,8 @@ authentication:
     - type: local
       username: admin   # username and password for login `manager api`
       password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC
-    - username: user
+    - type: local
+      username: user
       password: $2a$10$arSuTLefL6mgw8RcE9.ls.pJmnnE9PAYcbI.3POWKyrPyaIZmbeRm
 
 plugins:                          # plugin list (sorted in alphabetical order)
--- a/api/conf/conf.yaml
+++ b/api/conf/conf.yaml
@@ -70,10 +70,10 @@ authentication:
   users:                # yamllint enable rule:comments-indentation
     - type: local
       username: admin   # username and password for login `manager api`
-      password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC
+      password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC   # password: admin (bcrypt-cost10)
     - type: local
       username: user
-      password: $2a$10$arSuTLefL6mgw8RcE9.ls.pJmnnE9PAYcbI.3POWKyrPyaIZmbeRm
+      password: $2a$10$arSuTLefL6mgw8RcE9.ls.pJmnnE9PAYcbI.3POWKyrPyaIZmbeRm   # password: user (bcrypt-cost10)
 
 plugins:                          # plugin list (sorted in alphabetical order)
   - api-breaker
--- a/api/conf/conf.yaml
+++ b/api/conf/conf.yaml
@@ -68,10 +68,11 @@ authentication:
                         # if it's default value, when `manager api` start, it will generate a random string to replace it.
   expire_time: 3600     # jwt token expire time, in second
   users:                # yamllint enable rule:comments-indentation
-    - username: admin   # username and password for login `manager api`
-      password: admin
+    - type: local
+      username: admin   # username and password for login `manager api`
+      password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC
     - username: user
-      password: user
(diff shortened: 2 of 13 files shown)


api/test/docker/manager-api-conf.yaml
```diff
@@ -36,7 +36,9 @@ authentication:
                              # if it's default value, when `manager api` start, it will generate a random string to replace it.
   expire_time: 3600          # jwt token expire time, in second
   users:                     # yamllint enable rule:comments-indentation
-    - username: admin        # username and password for login `manager api`
-      password: admin
-    - username: user
-      password: user
+    - type: local
+      username: admin   # username and password for login `manager api`
+      password: $2a$10$NjjlIhuj4O6foZhWw96H0uXZQl0na8qg.BSbf/ASwurchFhUW0fzC
```
starsz, line 41: Why we should use the password's hash?

Dockerfile
```diff
@@ -16,13 +16,12 @@
...
+ARG APISIX_DASHBOARD_GITHASH=00000000
+
+ADD . /usr/local/apisix-dashboard
 
 RUN set -x \
-    && apk add --no-cache --virtual .builddeps git \
-    && git clone https://github.com/apache/apisix-dashboard.git -b ${APISIX_DASHBOARD_VERSION} /usr/local/apisix-dashboard \
-    && cd /usr/local/apisix-dashboard && git clean -Xdf \
-    && rm -f ./.githash && git log --pretty=format:"%h" -1 > ./.githash
+    && echo ${APISIX_DASHBOARD_GITHASH:0:8} > ./.githash
```
nic-chen, line 24: why change this file?

api/conf/conf.yaml
```diff
@@ -68,10 +68,12 @@ authentication:
                         # if it's default value, when `manager api` start, it will generate a random string to replace it.
   expire_time: 3600     # jwt token expire time, in second
   users:                # yamllint enable rule:comments-indentation
-    - username: admin   # username and password for login `manager api`
-      password: admin
-    - username: user
-      password: user
+    - type: local
```
nic-chen, line 71: Two user systems: local and etcd?

Dockerfile
bzp2010: - The original dockerfile compilation mechanism is to compile a branch of the current warehouse. Instead of selecting the latest code in the current directory for compilation, it loses the significance of providing dockerfile files.
- The direct reason is that the configuration file with hash password is modified, and the unchanged managerapi code in the master branch cannot pass the CI test.

api/test/docker/manager-api-conf.yaml
bzp2010: Because the configuration files themselves are stored locally, rather than other centralized configuration management software such as etcd, these configuration files are scattered in the file systems of different machines, which are easy to leak and cause harm.
Using a simple bcrypt hash algorithm with time cost to process the password can ensure that even if the configuration file is illegally obtained, the password text will not be leaked.

If you think this part of the function is unnecessary, I can modify it back.

Dockerfile
nic-chen: hi @bzp2010 
should we also update the dockerfile in repo apisix-docker?

api/test/docker/manager-api-conf.yaml
bzp2010: #### Update
I have removed the password hash.

Dockerfile
bzp2010: #### Changes
- Modify dockerfile to compile with local branch code.
- Instead of getting githash from clone's code, pass in githash variable through GitHub actions.

bzp2010: > should we also update the dockerfile in repo apisix-docker?

In fact, I don't think we need to update the `Dockerfile`. The code in that repo is mainly used to compile and upload to the registry, while the code in the current warehouse is more used for development purposes.

What do you think? Thx.

bzp2010:
Hi, @nic-chen, @tokers, @starsz. Sorry for bother you.
There have been some minor changes since the last review. Can you review it again to help me point out the problems? I will continue to improve it.

Thank you.

nic-chen (commented):
@bzp2010 
I thought about it, users should be stored in etcd uniformly, including local users. Otherwise, how to store users from third parties?

bzp2010:
@nic-chen 
1. We maybe able to use email as user ID (both GitHub and Google will provide this column). The user adds the mailbox, etc. to the configuration file or etcd in advance.
2. If we store user information in etcd, should we store the data together in apifix (use `/apisix` prefix)? Or use another prefix?

bzp2010:
### Current Plan
#### Update Time: 
2021-08-10

#### Description:
At present, we plan to provide two types of user storage methods. 

1. The first is a simple user saved in the configuration file and only supports username + password. 
_This is to ensure compatibility with the old version, which will be available for some time in the next version, but there may be no feature updates. It is recommended to migrate to the following second storage.
It is expected that local user storage will not be provided after the installation initialization solution of the second solution is solved. But, there are still some key problems to be solved, so first method will still be used as the default configuration._

2. The second scheme is to store users in etcd in the form of dashboard independent prefix.
_This is a form of user storage that will be supported in the future, and it will provide more flexibility. However, it needs to deal with the initialization of dashboard after installation. At present, it needs to manually add user records to etcd.
In the future, user login methods such as oAuth2 will also be provided._

#### PR Contents:

- This PR is committed to establishing a new universal user framework to support more authentication methods. It will be used as the front work of the second kind mentioned above.
It establishes the location of user storage in the a directory of etcd, it uses a data model similar to the following image.
image
It allows users to set the master account for logging in with username and password, and associate other platform accounts such as oAuth2 with the master account. Just like this.
image
More authentication methods will be added in the form of provider in the future, such as `GitHub`, `Google`

- At the same time, I added a new configuration item datasource in the authentication block in the configuration file. It lets us to specify `local` or `etcd` storage. In the current version, it is still configured as `local`. After the initial user creation problem is solved, switch it to `etcd` and consider removing the local user authentication method.

#### Other
Do you have any other suggestions?
@nic-chen

liuxiran:
Hi @bzp2010 , thanks for this greate change for dashboard. 

It is believed that user authentication is also a concern of many open source users, it would be better to send e-mail to our mail list, that would get more people's attention and discussion. 

You can summarize the current implementation, its shortcomings, areas that need to be improved, and then focus on the latest solutions, and where you want to ask people for advice. 

After much discussion, we can continue to complete the current PR 😊

bzp2010:
> Hi @bzp2010 , thanks for this greate change for dashboard.
> 
> It is believed that user authentication is also a concern of many open source users, it would be better to send e-mail to our mail list, that would get more people's attention and discussion.
> 
> You can summarize the current implementation, its shortcomings, areas that need to be improved, and then focus on the latest solutions, and where you want to ask people for advice.
> 
> After much discussion, we can continue to complete the current PR 😊

Yes, I have sent a discussion email to the mailing list with a link to the link described above.

bzp2010:
### Update
1. Two methods of development have been completed
2. Waiting to write test cases

### Usage
- #### Local Storage

1. In the authentication section of conf.yaml, set data_ The source field is set to local.
2. Add a user record in the following format to the users list in the authentication section of conf.yaml.

> `- username: xxx`

- #### ETCD Storage

shyamalmadura:
Hi team,
can we include a **salt** as well along with user name and password to improve security?

bzp2010:
> Hi team,
> can we include a **salt** as well along with user name and password to improve security?

I'm going to try to use `bcrypt` directly for hash instead of salt.

bzp2010:
Hi, @nic-chen @starsz @tokers @juzhiyuan @liuxiran .

This PR has been hang for a long time. Can we review it again and merge it? Thank you.

api/internal/filter/authentication.go
```diff
@@ -90,7 +93,40 @@ func (mw *AuthenticationMiddleware) Handle(ctx droplet.Context) error {
 		return nil
 	}
 
-	if _, ok := conf.UserList[claims.Subject]; !ok {
+	userExist := false
+	switch conf.DataSourceType(claims.Audience) {
+	case conf.DataSourceTypeEtcd:
```
nic-chen, line 98: In my understanding, `local` should be written into etcd through yaml configuration, and `external` is from third-party. It looks like `local` and `etcd` are still different here? In this way, users configured in yaml cannot be managed on the dashboard.

api/internal/conf/authentication.go
```diff
@@ -0,0 +1,85 @@
...
+	"strconv"
+
+	"github.com/apisix/manager-api/internal/utils"
+	"github.com/apisix/manager-api/internal/utils/consts"
+)
+
+var (
+	// UserList is a mapping of user type and users
+	// like  local  => [User1, User2] google => [User3, User4] github => [User5, User6]
+	UserList   = make([]DashboardUserLocal, 0)
```
starsz, line 29: Why not use map instead ?

api/internal/filter/authentication.go
```diff
@@ -90,7 +93,40 @@ func (mw *AuthenticationMiddleware) Handle(ctx droplet.Context) error {
...
+			return nil
+		}
+
+		if ret.TotalSize > 0 {
+			userExist = true
+		}
+	case conf.DataSourceTypeLocal:
+		fallthrough
+	default:
+		for _, item := range conf.UserList {
```
starsz, line 122: IMO, use map would be better?

api/internal/handler/authentication/authentication.go
```diff
@@ -87,22 +94,61 @@ func (h *Handler) userLogin(c droplet.Context) (interface{}, error) {
...
+		}
 
-	// create JWT for session
-	claims := jwt.StandardClaims{
-		Subject:   username,
-		IssuedAt:  time.Now().Unix(),
-		ExpiresAt: time.Now().Add(time.Second * time.Duration(conf.AuthConf.ExpireTime)).Unix(),
+		if ret.TotalSize == 1 {
+			user := ret.Rows[0].(*entity.DashboardUser)
+			err := bcrypt.CompareHashAndPassword([]byte(user.Password), []byte(password))
```
starsz, line 128: I would prefer the way that uses salt.

iamayushdas (approved):

PR #2054, opened 2021-08-09
fix: perform pid lookup for existing pid file

### LINKED ISSUE #2008: chore: show a friendly prompt, manager-api.pid exists but there's no corresponding process
## Issue description

When you shutdown the machine directly, the manager-api.pid won't be deleted.
After rebooting and start manager-api, it reports “instance of the Manager API already running...” but actually it's not.

we can show a friendly prompt: manager-api.pid exists but there's no corresponding process, and then start the new instance.

bisakhmondal:

- [x] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

https://github.com/apache/apisix-dashboard/issues/2008#issuecomment-894655656

**Related issues**

closes #2008 

**Checklist:**

### DIFF
--- a/api/internal/utils/pid.go
+++ b/api/internal/utils/pid.go
@@ -18,19 +18,62 @@
 package utils
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
 	"strconv"
+	"syscall"
+	"time"
+
+	"github.com/pkg/errors"
 )
 
 // WritePID write pid to the given file path.
 func WritePID(filepath string, forceStart bool) error {
-	if pid, err := ReadPID(filepath); err == nil {
-		if !forceStart {
-			return fmt.Errorf("Instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag\n", pid)
+	pidStatAction := func() error {
+		pid, err := ReadPID(filepath)
+		if err != nil {
+			return nil
 		}
-		fmt.Printf("Force starting new instance. Another instance of Manager API maybe running with pid %d\n", pid)
+		// In unix system this always succeeds, doesn't ensure if the process at all exist or not.
+		proc, err := os.FindProcess(pid)
+		if err != nil {
+			return errors.Errorf("instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag. Error :%v\n", pid, err)
+		}
+
+		// Traditional unix way to check the existence of the particular PID.
+		err = proc.Signal(syscall.Signal(0))
+		// No entry of the process in process control block.
+		if err != nil {
+			return nil
+		}
+
+		// Windows has a tendency to pick unused old PIDs and allocate that to new process, so we won't kill it.
+		if runtime.GOOS != "windows" && forceStart {
+			fmt.Println("Killing existing instance of Manager API")
+			if err := proc.Signal(syscall.SIGINT); err != nil {
+				return errors.Wrapf(err, "failed to kill the existing process of PID %d ", pid)
+			}
+			// Wait for graceful shutdown.
+			for {
+				if err := proc.Signal(syscall.Signal(0)); err != nil {
+					break
+				}
+				time.Sleep(time.Second)
+			}
+			fmt.Println("Existing instance of Manager API successfully killed")
+		}
+		if forceStart {
+			fmt.Printf("Force starting new instance. Another instance of Manager API was running with pid %d\n", pid)
+		}
+
+		return errors.Errorf("instance of Manager API is already running. Either kill the existing instance of PID %d or run Manager API with '-f' or '--force' flag", pid)
+	}
+
+	if err := pidStatAction(); err != nil && !forceStart {
+		return err
 	}
 	pid := os.Getpid()
 	f, err := os.OpenFile(filepath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC|os.O_CREATE, 0600)
@@ -51,7 +94,7 @@ func ReadPID(filepath string) (int, error) {
 	if err != nil {
 		return -1, err
 	}
-	pid, err := strconv.Atoi(string(data))
+	pid, err := strconv.Atoi(string(bytes.Trim(data, " \n")))
 	if err != nil {
 		return -1, fmt.Errorf("invalid pid: %s", err)
 	}
--- a/api/internal/utils/pid.go
+++ b/api/internal/utils/pid.go
@@ -35,12 +35,16 @@ func WritePID(filepath string, forceStart bool) error {
 	pidStatAction := func() error {
 		pid, err := ReadPID(filepath)
 		if err != nil {
-			return nil
+			if _, ok := err.(*os.PathError); ok {
+				return nil // No pid file indicates no running instance of manager-api
+			}
+			return err // Other errors in ReadPID must be reported.
 		}
-		// In unix system this always succeeds, doesn't ensure if the process at all exist or not.
+
+		// In unix system this always succeeds, doesn't ensure if the process at all exist or not. In windows, it returns an error indicating no such process exists.
 		proc, err := os.FindProcess(pid)
 		if err != nil {
-			return errors.Errorf("instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag. Error :%v\n", pid, err)
+			return nil
 		}
 
 		// Traditional unix way to check the existence of the particular PID.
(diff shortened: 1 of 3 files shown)


bisakhmondal:
Pinging @nic-chen for the review.
Hey, could you please review it when you get some time. Thanks.

nic-chen (commented):
@bisakhmondal need test cases for it

liuxiran:
hi @bisakhmondal be-cli failed, please fix it when you have time, thanks

iamayushdas (approved):

api/internal/utils/pid.go
```diff
@@ -18,19 +18,62 @@
...
 
 // WritePID write pid to the given file path.
 func WritePID(filepath string, forceStart bool) error {
-	if pid, err := ReadPID(filepath); err == nil {
-		if !forceStart {
-			return fmt.Errorf("Instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag\n", pid)
+	pidStatAction := func() error {
+		pid, err := ReadPID(filepath)
+		if err != nil {
+			return nil
```
nic-chen, line 38: why return nil here？

bisakhmondal: Hi @nic-chen , I am following a fail-fast approach. The ReadPID returns error means it will most likely be `os.PathError` i.e. the file does not exist.  Instead of writing whole block as `if err==nil` (previous implementation),  I was trying to introduce less spaghetti code. 

If you want, we definitely can improve the err to something like

```go
if err != nil {
  if _, ok := err.(*os.PathError); ok {
     return nil // No pid file --> no running instance
  }
  return err
}
```
Let me know what you think. Thanks.

starsz: LGTM.

api/internal/utils/pid.go
```diff
@@ -18,19 +18,62 @@
...
+		proc, err := os.FindProcess(pid)
+		if err != nil {
+			return errors.Errorf("instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag. Error :%v\n", pid, err)
+		}
+
+		// Traditional unix way to check the existence of the particular PID.
+		err = proc.Signal(syscall.Signal(0))
+		// No entry of the process in process control block.
+		if err != nil {
+			return nil
```
starsz, line 50: Ditto.

api/internal/utils/pid.go
```diff
@@ -18,19 +18,62 @@
...
+				return errors.Wrapf(err, "failed to kill the existing process of PID %d ", pid)
+			}
+			// Wait for graceful shutdown.
+			for {
+				if err := proc.Signal(syscall.Signal(0)); err != nil {
+					break
+				}
+				time.Sleep(time.Second)
+			}
+			fmt.Println("Existing instance of Manager API successfully killed")
```
starsz, line 66:
```suggestion
			fmt.Println("Existing instance of Manager API successfully exited")
```

api/internal/utils/pid.go
bisakhmondal: Sorry couldn't get you. Do you mean it is  okay or the same question "why return nil here？"
If it is the latter one, the pr tries to avoid the false positive that arises due to the stale pid file in `/tmp` (maybe that was left during shutdown and wasn't got cleaned up)

so instead of blindly throwing an error just because of the presence of PID file, the code block looks up if the process at all exists or not. If it doesn't (err is not nil), that's completely fine. We will run a new instance and overwrite the existing pid. That's it.

api/internal/utils/pid.go
bisakhmondal: Cool. Thanks

api/internal/utils/pid.go
nic-chen: OK. 
We could add a comment here, so other people will not be confused when they see it.

nic-chen (approved):

api/internal/utils/pid.go
```diff
@@ -51,7 +94,7 @@ func ReadPID(filepath string) (int, error) {
 	if err != nil {
 		return -1, err
 	}
-	pid, err := strconv.Atoi(string(data))
+	pid, err := strconv.Atoi(string(bytes.Trim(data, " \n")))
```
starsz, line 97: Do we need a blank space before `\n` ?

api/internal/utils/pid.go
```diff
@@ -18,19 +18,62 @@
...
+		// Traditional unix way to check the existence of the particular PID.
+		err = proc.Signal(syscall.Signal(0))
+		// No entry of the process in process control block.
+		if err != nil {
+			return nil
+		}
+
+		// Windows has a tendency to pick unused old PIDs and allocate that to new process, so we won't kill it.
+		if runtime.GOOS != "windows" && forceStart {
+			fmt.Println("Killing existing instance of Manager API")
```
freecodelover, line 55: It would be better to add the pid into the message.

api/internal/utils/pid.go
```diff
@@ -18,19 +18,62 @@
...
-			return fmt.Errorf("Instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag\n", pid)
+	pidStatAction := func() error {
+		pid, err := ReadPID(filepath)
+		if err != nil {
+			return nil
+		}
+		// In unix system this always succeeds, doesn't ensure if the process at all exist or not.
+		proc, err := os.FindProcess(pid)
+		if err != nil {
+			return errors.Errorf("instance of Manager API maybe running with a pid %d. If not, please run Manager API with '-f' or '--force' flag. Error :%v\n", pid, err)
```
freecodelover, line 43: I think `os. FindProcess` will return an error when the process does not exist in windows according to:
https://golang.org/src/os/exec_windows.go

So, we should not return the error here?

api/internal/utils/pid.go
bisakhmondal: Cool

api/internal/utils/pid.go
bisakhmondal: Thanks for pointing it out. Updated

api/internal/utils/pid.go
bisakhmondal: in go1.16 I was getting `invalid pid: strconv.Atoi: parsing "3041138\n": invalid syntax` parsing error ( newline error) without any newline at the end. 

However blank space is not required.

api/internal/utils/pid.go
bisakhmondal: Done

starsz (approved):

bzp2010:
### Nice jobs!

But, with all due respect, the current management mode of service daemon and the use of system calls during use will undermine the compatibility of the operating system.
Therefore, I think our finally goal should be to remove the service daemon management part from the Manager API and transfer it to the container or system D, which is compatible with the container and RPM installation. Let professional software do professional things.

#### What do you think of this?

iamayushdas:
@bzp2010 agree with you, transferring it to container will make it more flexible

starsz:
> ### Nice jobs!
> But, with all due respect, the current management mode of service daemon and the use of system calls during use will undermine the compatibility of the operating system.
> Therefore, I think our finally goal should be to remove the service daemon management part from the Manager API and transfer it to the container or system D, which is compatible with the container and RPM installation. Let professional software do professional things.
> 
> #### What do you think of this?
> > cc @bisakhmondal @starsz @nic-chen @iamayushdas @liuxiran

So we can only provide a service configuration?

bzp2010:
> So we can only provide a service configuration?

I don't understand `service configuration`, can you elaborate?

bisakhmondal:
Hi @bzp2010, thanks for the input.

> But, with all due respect, the current management mode of service daemon and the use of system calls during use will undermine the compatibility of the operating system.

Ideally, it might not be a good practice to use syscalls in the program but I don't think it would be a cause of some problem due to os compatibility (all the stable os kernels are written in C and assembly so libc is already there for handling syscalls).
As for the context of current PR, it uses syscalls to ensure the pid that is written in `/tmp/manager-api.pid` is actually running or not. Please let me know if you have any alternatives.

> Therefore, I think our finally goal should be to remove the service daemon management part from the Manager API and transfer it to the container or systemd, which is compatible with the container and RPM installation. Let professional software do professional things.

I do not fully understand what you meant here. Just FYI and if I remember correctly, the current daemon management package for Manager API internally uses systemd (unit files with go templates) [[link](https://github.com/takama/daemon/blob/496d69192531c6cb1004380a0be0fcf2031b06f4/daemon_linux_systemd.go#L213-L227)].

> I don't understand service configuration, can you elaborate?

I guess @starsz wanted to say configuration unit files (os dependent, like what I shared in the link) for different os service managers as you proposed to remove daemon management part from Manager API. 

Thanks :)

bzp2010:
## Update
In the previous PR we have removed the dependency on the built-in service management library and also removed the code for PID file generation.
## 
So I'll close this PR for now and you can open it again if you need to.

PR #2099, opened 2021-08-26
feat: support proto API

bzp2010:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

Add `proto` entity support to Manager API.

**Related issues**

resolve part of #1378 #1449 

**Checklist:**

- [x] Did you explain what problem does this PR solve? Or what new features have been added?
- [x] Have you added corresponding test cases?
- [x] Have you modified the corresponding document?
- [x] Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

PR #2144, opened 2021-09-20
feat: support proto API (entity)

bzp2010:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

As part of the proto API support, this PR adds data entities about proto to the Manager API. And make some improvements to the data layer.

**Related issues**

part of #2099 

**Checklist:**

### DIFF
--- a/api/internal/core/entity/entity.go
+++ b/api/internal/core/entity/entity.go
@@ -283,3 +283,10 @@ type PluginConfig struct {
 	Plugins map[string]interface{} `json:"plugins"`
 	Labels  map[string]string      `json:"labels,omitempty"`
 }
+
+// swagger:model Proto
+type Proto struct {
+	BaseInfo
+	Desc    string `json:"desc,omitempty"`
+	Content string `json:"content"`
+}
--- a/api/internal/core/store/store.go
+++ b/api/internal/core/store/store.go
@@ -42,6 +42,7 @@ type Pagination struct {
 }
 
 type Interface interface {
+	Type() HubKey
 	Get(ctx context.Context, key string) (interface{}, error)
 	List(ctx context.Context, input ListInput) (*ListOutput, error)
 	Create(ctx context.Context, obj interface{}) (interface{}, error)
@@ -64,6 +65,7 @@ type GenericStoreOption struct {
 	KeyFunc    func(obj interface{}) string
 	StockCheck func(obj interface{}, stockObj interface{}) error
 	Validator  Validator
+	HubKey     HubKey
 }
 
 func NewGenericStore(opt GenericStoreOption) (*GenericStore, error) {
@@ -106,7 +108,7 @@ func (s *GenericStore) Init() error {
 		key := ret[i].Key[len(s.opt.BasePath)+1:]
 		objPtr, err := s.StringToObjPtr(ret[i].Value, key)
 		if err != nil {
-			fmt.Fprintln(os.Stderr, "Error occurred while initializing logical store: ", s.opt.BasePath)
+			_, _ = fmt.Fprintln(os.Stderr, "Error occurred while initializing logical store: ", s.opt.BasePath)
 			return err
 		}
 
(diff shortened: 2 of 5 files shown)


api/internal/core/entity/interface.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (p *PluginConfig) GetBaseInfo() *BaseInfo {
+	return &p.BaseInfo
+}
+
+func (s *ServerInfo) GetBaseInfo() *BaseInfo {
+	return &s.BaseInfo
+}
+
+func (p *Proto) GetBaseInfo() *BaseInfo {
```
nic-chen, line 83: What are the benefits of defining a method `GetBaseInfo` for each struct?

bzp2010: It just replaces the original `BaseInfoGetter` and `BaseInfoSetter`. Because they are all operations on the structure pointer, there is no need to implement them twice. Now use a single method to cover all the data structures, making it easier to get `BaseInfo`.

nic-chen: I think the risk is greater than the benefit. Because when adding a new struct, may forget to add this method for it, resulting in an error when processing the request.

bzp2010: Yes, I can understand.
But I found some problems with the previous code. Just like this,
[https://github.com/apache/apisix-dashboard/blob/a3928d8bbcbaec089458db01eb7a380f20fb33ae/api/internal/core/store/store.go#L174-L175](https://github.com/apache/apisix-dashboard/blob/a3928d8bbcbaec089458db01eb7a380f20fb33ae/api/internal/core/store/store.go#L174-L175)
Although type conversion is performed here, none of the data entities actually implement the getter interface. When executing list query, this will cause panic in the program, so I am very confused.

nic-chen: `entity.BaseInfo` is the implement

bzp2010: Oh, I get it. I will modify it this way.

bzp2010: @nic-chen I have removed the implementation of `GetBaseInfo` in the data entity, and now only the implementation of `GetPlugins` is left.

liuxiran:
Hi @bzp2010 it would be better to add test cases

api/internal/core/entity/entity.go
```diff
@@ -58,14 +58,6 @@ func (info *BaseInfo) KeyCompat(key string) {
 	}
 }
 
-type BaseInfoSetter interface {
-	GetBaseInfo() *BaseInfo
-}
-
-type BaseInfoGetter interface {
-	GetBaseInfo() *BaseInfo
-}
```
nic-chen, line 67: oh, please keep them, their responsibilities are different.

bzp2010: `BaseInfoGetter` and `BaseInfoSetter` are currently replaced by `GetBaseInfo` interface. Their interface definitions are exactly the same. And, the lines of codes is reduced.

bzp2010: ping @nic-chen

bzp2010:
ping @nic-chen @gxthrj @starsz @tokers 

Any more review comments here?

api/internal/core/entity/entity.go
```diff
@@ -290,3 +282,10 @@ type PluginConfig struct {
 	Plugins map[string]interface{} `json:"plugins"`
 	Labels  map[string]string      `json:"labels,omitempty"`
 }
+
+// swagger:model Proto
```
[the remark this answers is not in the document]
bzp2010: I think this can be used to generate API documents and should be retained.

nic-chen (approved):

starsz (approved):

juzhiyuan (approved):

PR #2171, opened 2021-10-12
Adjust the read and write timeout configuration of the http server

### LINKED ISSUE #2165: HTTP server write timeout value too small
### Issue description

When the Dashboard is deployed on a server with limited network bandwidth, the `HTTP server write timeout`([code](https://github.com/apache/apisix-dashboard/blob/f50779a3134da0f97114fb54313e1c77ae042b65/api/internal/core/server/http.go#L54-L54)) value is too small, resulting in a timeout when downloading js files, Chrome reports `ERR_CONTENT_LENGTH_MISMATCH` error, and the page continues to be in the loading state. 
image
image
image
image

### Expected behavior

The page can be displayed normally in a network-restricted environment

### How to Reproduce

1. Deploy dashboard on servers with low bandwidth(You can use linux tc to simulate low bandwidth scenarios)

nic-6443:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Adjust the `read` and `write` timeout configuration of the http server so that the Dashboard can work normally in a slow network environment. Especially write timeout, because the Dashboard has a relatively large front-end static resource file (>2M). When the bandwidth is limited It is easy to trigger the timeout in the scene, which will cause the page display to fail.

**Related issues**

fix/resolve #2165

**Checklist:**

### DIFF
--- a/api/internal/core/server/http.go
+++ b/api/internal/core/server/http.go
@@ -50,8 +50,8 @@ func (s *server) setupAPI() {
 	s.server = &http.Server{
 		Addr:         addr,
 		Handler:      r,
-		ReadTimeout:  time.Duration(1000) * time.Millisecond,
-		WriteTimeout: time.Duration(5000) * time.Millisecond,
+		ReadTimeout:  time.Duration(5000) * time.Millisecond,
+		WriteTimeout: time.Duration(60000) * time.Millisecond,
 	}
 
 	// HTTPS
@@ -60,8 +60,8 @@ func (s *server) setupAPI() {
 		s.serverSSL = &http.Server{
 			Addr:         addrSSL,
 			Handler:      r,
-			ReadTimeout:  time.Duration(1000) * time.Millisecond,
-			WriteTimeout: time.Duration(5000) * time.Millisecond,
+			ReadTimeout:  time.Duration(5000) * time.Millisecond,
+			WriteTimeout: time.Duration(60000) * time.Millisecond,
 			TLSConfig: &tls.Config{
 				// Causes servers to use Go's default ciphersuite preferences,
 				// which are tuned to avoid attacks. Does nothing on clients.


nic-chen:
> > hi, @nic-6443 which file is so large?
> 
> The front-end static files from the `/usr/local/apisix-dashboard/html/` directory. The largest file (`p__Route__Create.8a1b69b7.async.js`) has 2.1M. Under the current 5s timeout setting, this The transmission speed of the network is required to reach 430K/s, otherwise the file will fail to download.
> 
> image When users deploy Dashboard on a public cloud, this problem will be easily triggered.

hi, @bzp2010 
Could we optimize the size of these files? 
I remember there was a saying before, that a single file should be controlled within 100k to ensure user experience. 
The internet speed is relatively high now, but a single file should not be that large.

nic-6443:
60s refers to the default value of `send_timeout` of nginx (although `send_timeout` is not for the transmission of the whole response).
I understand that `WriteTimeout` is used to prevent slow client attacks ([Slowloris Attack](https://en.wikipedia.org/wiki/Slowloris_(computer_security))), and Dashboard is usually not exposed to the public network , So I think there is no risk to increase `WriteTimeout`. Looking forward to your discussion.

zaunist:
> > > Should we add custom configuration timeouts? Allow low bandwidth users access, public network configuration can be limited to IP.
> > 
> > 
> > @nic-6443 we could support configure them in `conf/conf.yaml`
> 
> I think it's ok, but I still insist the default value can be larger so that users will not easily encounter problems when using it for the first time, just like me.Especially when there is no obvious risk of increasing timeout.

My advice is to consider how to compress js files first, and if there is no way to compress them, then discuss increasing the default timeout

nic-6443:
> > > > Should we add custom configuration timeouts? Allow low bandwidth users access, public network configuration can be limited to IP.
> > > 
> > > 
> > > @nic-6443 we could support configure them in `conf/conf.yaml`
> > 
> > 
> > I think it's ok, but I still insist the default value can be larger so that users will not easily encounter problems when using it for the first time, just like me.Especially when there is no obvious risk of increasing timeout.
> 
> My advice is to consider how to compress js files first, and if there is no way to compress them, then discuss increasing the default timeout

ok, no problem

starsz (approved):

nic-6443:
> @nic-6443 Would you like to try using gin `gzip` middleware to solve the problem? If you don't have time to do it, I'll try to fix it.
I can do it, do I need a new pull request?

PR #2238, opened 2021-12-09
feat: Support APISIX_PROFILE for env-specific configuration

### LINKED ISSUE #2194: Should apisix support environment config or profile config?
# Feature request
For example  I want to change etcd  ip address and  apisix  prefix for different environment, according to https://github.com/apache/apisix-dashboard/blob/master/api/conf/conf.yaml#L39, how can I do it?

A clear and concise description of what you want and what your use case is.

kevinw66:

- [*] New feature provided

Support env config just like apisix
https://apisix.apache.org/zh/docs/apisix/profile/

**Related issues**

fix/resolve #2194


juzhiyuan:
Hi @kevinw66, please take a look at the CI 😄

zaunist:
@juzhiyuan  It looks like the CI failed due to unstable CI, I found stream route test untill very unstable, the part of the test is easy to fail.
@bzp2010 Can you take some time to look the stream route test ?

juzhiyuan:
I just rerun the CI, wait for a moment.

kevinw66:
I merged new commits from master.
BTW, the ut is hard to run in local machine because of too many dependencies server, is there any plan to solve this? like mock server or embbed server.

zaunist:
> I merged new commits from master. BTW, the ut is hard to run in local machine because of too many dependencies server, is there any plan to solve this? like mock server or embbed server.

We can use `make api-test` to run unit test, I think you might be talking about integration testing?

kevinw66:
> > I merged new commits from master. BTW, the ut is hard to run in local machine because of too many dependencies server, is there any plan to solve this? like mock server or embbed server.
> 
> We can use `make api-test` to run unit test, I think you might be talking about integration testing?

I tried to run failed test `stream_route_test` in my local machine, but I found I have to start dashboard in local, and then run an upstream, or else it will failed in `connection refused`.

juzhiyuan:
image

Hi @zaunist, could you please give some hints on executing tests locally?

zaunist:
> 

I also have this problem in my PR submission. @bzp2010  more familiar with this, any solutions?

bzp2010:
Hi, @zaunist.
Recently we have not modified the E2E test implemented using the old way, and its instability is perhaps caused by the test environment.
And in the new CI using ginkgo, I have solved one instability issue related to migrate, but it still may fail due to temporary environment exceptions. Also, the UDP case in the Stream Route test does still have some failure issues, which in general are not very common.

bzp2010:
Hi, @kevinw66. It's great to see you contributing to the project. 

But do we really need this feature? As far as I know, we can now rewrite the config file path using the `-c` flag. Can you describe some of the differences between the current PR and `-c` flag scenarios?

Thanks.

kevinw66:
> Hi, @kevinw66. It's great to see you contributing to the project.
> 
> But do we really need this feature? As far as I know, we can now rewrite the config file path using the `-c` flag. Can you describe some of the differences between the current PR and `-c` flag scenarios?
> 
> Thanks.

When we download bin file and use it directly, -c is enough. But if we have to redevelop from source code, mostly we will create env-specific config, and the -c may not so quite useful.
On the other hand, I think it's better to keep dashboard project and main project have same behavior, like config-default 、environment variable like `etcd.host: ${{ETCD_HOST}}` and so on.

bzp2010:
Hi, @kevinw66.

Sorry to reply to your delay, I miss it when dealing with PR, I have learned the purpose of this function, this feature is also used in Apache APISIX, I think it is no problem, but can you add some tests to it?

bzp2010 (commented):
Still need some testing to ensure it works fine. 🤔

Issue #2148, opened 2021-09-25
Proposal: support Plugin Metadata API


bzp2010:
# Feature request

## Background
At present, some plugins in Apache APISIX have metadata configuration items. The current dashboard project cannot support setting these contents. We need to support it. This requires adding APIs at the backend and pages at the frontend.

### About metadata
Metadata is a mechanism provided by Apache APISIX to store some global configuration content for plugins. It is different from plugin configuration on route or other entities. It is globally effective for plugins and takes plugins as dimensions. (for example, after configuring the metadata of `authz-caspin` plugin, the plugin on any route or service can obtain the content stored in the metadata)

### A plugin can support metadata?
We can easily distinguish whether a plugin needs metadata configuration. Plugins with the following structure in the plugin list of the built-in schema.json file in dashboard can support metadata configuration.
<img src="https://user-images.githubusercontent.com/8078418/134772532-e20aaec2-4655-4642-99a9-ea56910625db.png"/>

### Backend
1. Plugin List API
Read the metadata in the plugin information by traversing schema.json‘s schema property to generate a list of plugins that need metadata configuration and the property rules of plugins.
    - URI: `/apisix/admin/plugin_metadata`
    - Method: `GET`

2. Plugin Metadata Update API
According to the characteristics of plugin metadata, I think we only need to obtain and update APIs, rather than create API. When users want to create metadata, they can directly update the data JSON and store it in etcd
    - URI: `/apisix/admin/plugin_metadata/{plugin_name}`
    - Method: `PUT`

3. Plugin Metadata Delete API
Provides the function of clearing metadata data of plugin
    - URI: `/apisix/admin/plugin_metadata/{plugin_name}`
    - Method: `DELETE`

### Frontend
1. Plugin List page
Obtain plugins and their data from the plugin list API (the effect is shown below, but only the plugins that need to configure metadata are displayed)
![image](https://user-images.githubusercontent.com/8078418/134772915-85e0e402-fd8f-4d4b-84ee-8eb350c59ad1.png)

2. Plugin Metadata Setting page
Provide an input box that supports JSON and YAML, run users to input metadata data, and support input prompts.

## Subtask decomposition
1. Add Plugin Metadata data entity and storage layer code
2. Add backend APIs and related test code
3. Add frontend page and E2E test code

bzp2010:
Hi, everyone.
If you have any suggestion, you can send comments here.

tzssangglass:
[...]

PR #2267, opened 2021-12-27
feat: add plugin_metadata

TakumiInui5432:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

support plugin-metadata for manager-api

**Related issues**

part of #2148 

**Checklist:**

- [ ] Did you explain what problem does this PR solve? Or what new features have been added?

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -6,6 +6,8 @@ replace google.golang.org/grpc => google.golang.org/grpc v1.26.0
 
 replace github.com/coreos/bbolt => go.etcd.io/bbolt v1.3.5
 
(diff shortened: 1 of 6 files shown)


zaunist:
During the development process, I found that the structure of plugin_metadata in the plugin is uncertain, and the structure is similar to this:
```
{
    "skey": "val",
    "ikey": 1
}

```
but if we use current storage(store.go) to GET、PUT、and DELETE `plugin_metada`，it need be like this structure:
```
{
"plugin_metadata":{
  "ikey":1,
  "skey":"val"
  }
}
```
If we don’t make a lot of changes to the current storage(store.go), it will be difficult to support plugin_metadata in the dashboard, but the v3 version is already planned. I don’t think it is necessary to spend more time on this function. When designing the v3 version, add this Features plus will be more suitable

PR #2385, opened 2022-03-17
fix: The bug of missing secret field and algorithm when creating jwt-auth has been solved

### LINKED ISSUE #2384: request help: ensure explicitly sets the secret field when using jwt-auth on consumer
# Feature request

**The current jwt-auth plugin configuration in APISIX has special logic. If a user uses `HS` mode but does not set a secret, APISIX will generate one and write it to etcd, but the dashboard does not have the corresponding logic.**

Before we switch to using the Admin API directly, we need some means to circumvent this problem, and I think we can add a form UI to it that helps users avoid this problem by forcing input requirements and automatically generating random characters.

## Describe alternatives you've considered

The above solution is a frontend solution, we can also extend the logic in the backend to merge a copy of the patched jsonschema into the actual version to ensure we can enforce the secret presence check. (We can't directly modifying the jsonschema data exported through the APISIX Control API)

## Additional context

https://github.com/apache/apisix/issues/6564
https://github.com/apache/apisix-dashboard/issues/2383
https://github.com/apache/apisix-dashboard/issues/2176#issuecomment-944127485

kingmouse-yx:
…empty, it will be assigned as HS256. If the secret is empty, a secret will be automatically generated.

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

The bug of missing secret field and algorithm when creating jwt-auth has been solved

**Related issues**

resolve #2384

**Checklist:**

### DIFF
--- a/api/internal/handler/consumer/consumer.go
+++ b/api/internal/handler/consumer/consumer.go
@@ -17,6 +17,8 @@
 package consumer
 
 import (
+	"encoding/base64"
+	"math/rand"
 	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


quanbisen:
Hey, man. Can you resolve this merge conflict？

PR #2386, opened 2022-03-17
fix: Fix Copyright notice

nfrankel:
From [Wikipedia](https://en.wikipedia.org/wiki/Copyright#Copyright_notice):

  1) United States law required the use of a copyright notice, consisting of the
  copyright symbol (©, the letter C inside a circle), the abbreviation "Copr.",
  or the word "Copyright", **followed by the year of the first publication** of the
  work and the name of the copyright holder.

  2) In 1989 the United States enacted the Berne Convention Implementation Act,
  amending the 1976 Copyright Act to conform to most of the provisions of the
  Berne Convention. As a result, the use of copyright notices has become
  optional to claim copyright, because the Berne Convention makes copyright
  automatic.

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**Checklist:**

- [X] Did you explain what problem does this PR solve? Or what new features have been added?

### DIFF
--- a/web/src/components/Footer/index.tsx
+++ b/web/src/components/Footer/index.tsx
@@ -20,7 +20,7 @@ import { DefaultFooter } from '@ant-design/pro-layout';
 
 export default () => (
   <DefaultFooter
-    copyright="2021 Apache APISIX"
+    copyright="2019 Apache APISIX"
     links={[
       {
         key: 'GitHub',


bzp2010 (approved):

moonming:
I checked https://www.apple.com/ and https://github.com/, both of them are using `© 2022 xxx, Inc.`
So I'm not sure if the statement on the Wikipedia is correct, after all both apple and github have their own team of lawyers

nfrankel:
@moonming You're correct.

As for me, I trust Wikipedia as it references its sources. On the other hand, I don't assume that a team of lawyers I don't know did their job. I leave it up to you.

SkyeYoung:
@nfrankel 
I checked apache.org and I found that they are also using `Copyright © 2022`, and I thought, since we are an Apache project, let's keep it consistent with the ASF website
image

lilien1010 (approved):

1502shivam-singh (approved):
Regarding the conflict created in the file.
I am not sure whether copyright is something that needs to be dynamically updated, as in keeping the year recent. AFAIK The copyright date should include the year when the content was [first created and published on the site](https://www.copyrightlaws.com/copyright-symbol-notice-year/). Hence constantly evolving websites have something like `© <creation year>–<current year> Org`.
People often get quite flexible on this, but I think maybe following the rules would be nicer.

PR #2440, opened 2022-04-24
fix: advance matching in the route create page causes the page to crash

### LINKED ISSUE #2438: bug: advance matching in the route create page causes the page to crash
### Issue description

When we create an advanced match rule in a route, the page will crash if the reverse setting is not adjusted.

Based on the occurrence conditions, it is presumed that the default value of the reverse field is not set correctly. At that point, executing toString on a `null` value will result in a JS error.

https://github.com/apache/apisix-dashboard/blob/1fa5df3618708b392bb5fc2c866419b406cc43a7/web/src/pages/Route/components/Step1/MatchingRulesView.tsx#L148-L152

### Expected behavior

correct operation

### How to Reproduce

1. go to route create page
2. create a advance matching rule that not change `reverse` field (**don't click it**)
3. click `confirm`
4. page crashed

### Screenshots

_No response_

### Environment

- apisix-dashboard version, if have: 2.11.0

### Additional context

_No response_

Si-ege:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**Related issues**

fix/resolve #2438
1.fix matching in the route create page causes the page to crash

**Checklist:**

### DIFF
--- a/web/cypress/integration/route/can-skip-upstream-when-select-service-id.spec.js
+++ b/web/cypress/integration/route/can-skip-upstream-when-select-service-id.spec.js
@@ -36,6 +36,7 @@ context('Can select service_id skip upstream in route', () => {
     value: '#value',
     operator: '#operator',
     rowcard: '.ant-table-row-level-0',
+    reverse: '#reverse',
   };
 
   const data = {
@@ -103,21 +104,6 @@ context('Can select service_id skip upstream in route', () => {
     cy.contains('None').click();
     cy.contains(data.serviceName).click();
     cy.get(selector.enable_websocket).click();
-
-    cy.get(selector.addbtn).contains('Add').click();
-    cy.get(selector.position).click();
-    cy.get(selector.selectItem).within(() => {
-      cy.contains('HTTP Request Header').click();
-    });
-    cy.get('.ant-form-item-control-input-content > #name').type(data.parameterName);
-    cy.get(selector.operator).click();
-    cy.get(selector.selectItem).within(() => {
-      cy.contains('IN').click();
-    });
-    cy.get(selector.value).type(data.value);
-    cy.contains('Confirm').click();
-    cy.get(selector.rowcard).should('be.visible');
-    cy.get(selector.rowcard).get('tr>td').eq(2).should('have.value', '');
     cy.contains('Next').click();
 
     // make sure upstream data can be saved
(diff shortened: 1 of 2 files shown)


Si-ege:
My understanding of this issue is:
1. the method toString is not executed when the default value of the reverse field is undefined

Baoyuantop:
Hi @Si-ege, we need E2E test for this change:
1. Go to Route List and click Create
2. Fill in the necessary information and click the `Add` button in the Advanced Routing Matching Conditions section.
3. Fill in everything in the Create Rule except `Reverse the result(!)` field
4. Click Confirm and we can see the data in the table and `Reverse the result(!)` field is false.
5. We can create this route successfully

Si-ege:
> ok

Baoyuantop:
Hi @Si-ege, thank you for your contribution. For the E2E test section, we would prefer to create a new test case, meaning to write the test in a new `it()` instead of changing the previous test.

web/cypress/integration/route/can-skip-upstream-when-select-service-id.spec.js
```diff
@@ -103,12 +112,45 @@ context('Can select service_id skip upstream in route', () => {
...
+    });
+    cy.get('.ant-form-item-control-input-content > #name').type(data.parameterName);
+    cy.get(selector.operator).click();
+    cy.get(selector.selectItem).within(() => {
+      cy.contains('IN').click();
+    });
+    cy.get(selector.value).type(data.value);
+    cy.contains('Confirm').click();
+    cy.get(selector.rowcard).should('be.visible');
+    cy.get(selector.rowcard).get('tr>td').eq(2).should('have.value', '');
```
Baoyuantop, line 143: Should it show up as false here?

web/src/pages/Route/components/Step1/MatchingRulesView.tsx
```diff
@@ -148,7 +148,8 @@ const MatchingRulesView: React.FC<RouteModule.Step1PassProps> = ({
     {
       title: formatMessage({ id: 'page.route.reverse' }),
       key: 'reverse',
-      render: (text: RouteModule.MatchingRule) => text.reverse.toString(),
+      render: (text: RouteModule.MatchingRule) =>
+        text.reverse === undefined ? text.reverse : text.reverse.toString(),
```
Baoyuantop, line 152: I think it is more appropriate to set an initial value for reverse in the Form.

guoqqqi:
ping @Si-ege  any update?

Baoyuantop (approved):

guoqqqi (approved):

LiteSun (approved):

PR #2460, opened 2022-05-30
refactor: OpenAPI 3 parse and convert

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Improved backend implementation of OpenAPI3 import to enable import of standard OpenAPI3 minimum use cases. In the current PR I will not do the integration work, this separate module will be checked by way of unit tests and in the next PR it will be integrated into the data import interface.

**Checklist:**

### DIFF
--- a/api/internal/handler/data_loader/loader/openapi3/import.go
+++ b/api/internal/handler/data_loader/loader/openapi3/import.go
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package openapi3
+
+import (
+	"errors"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/getkin/kin-openapi/openapi3"
+
+	"github.com/apisix/manager-api/internal/core/entity"
+	"github.com/apisix/manager-api/internal/handler/data_loader/loader"
+	"github.com/apisix/manager-api/internal/utils/consts"
+)
+
+func (o Loader) Import(input interface{}) (*loader.DataSets, error) {
+	if input == nil {
+		return nil, errors.New("input is nil")
+	}
+
+	// load OAS3 document
+	swagger, err := openapi3.NewSwaggerLoader().LoadSwaggerFromData(input.([]byte))
+	if err != nil {
+		return nil, err
+	}
+
+	// no paths in OAS3 document
+	if len(swagger.Paths) < 1 {
+		return nil, consts.ErrImportFile
+	}
+
+	if o.TaskName == "" {
+		o.TaskName = "openapi_" + time.Now().Format("20060102150405")
+	}
+
+	data, err := o.convertToEntities(swagger)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func (o Loader) convertToEntities(s *openapi3.Swagger) (*loader.DataSets, error) {
+	var (
+		// temporarily save the parsed data
+		data = &loader.DataSets{}
+		// global upstream ID
+		globalUpstreamID = o.TaskName + "_global"
+		// global uri prefix
+		globalPath = ""
+	)
+
+	// create upstream when servers field not empty
+	if len(s.Servers) > 0 {
+		var upstream entity.Upstream
+		upstream, globalPath = generateUpstreamByServers(s.Servers, globalUpstreamID)
+		data.Upstreams = append(data.Upstreams, upstream)
+	}
+
+	// each one will correspond to a route
+	for uri, v := range s.Paths {
+		// replace parameter in uri to wildcard
+		realUri := regURIVar.ReplaceAllString(uri, "*")
+		// generate route name
+		routeID := o.TaskName + "_" + strings.NewReplacer("/", "-", "{", "", "}", "").Replace(strings.TrimPrefix(uri, "/"))
+
+		// decide whether to merge multi method routes based on configuration
+		if o.MergeMethod {
+			// create a single route for each path
+			route := generateBaseRoute(routeID, v.Summary)
+			route.Uris = []string{globalPath + realUri}
+			route.UpstreamID = globalUpstreamID
+			for method := range v.Operations() {
+				route.Methods = append(route.Methods, strings.ToUpper(method))
+			}
+			data.Routes = append(data.Routes, route)
+		} else {
+			// create routes for each method of each path
+			for method, operation := range v.Operations() {
+				subRouteID := routeID + "_" + method
+				route := generateBaseRoute(subRouteID, operation.Summary)
+				route.Uris = []string{globalPath + realUri}
+				route.Methods = []string{strings.ToUpper(method)}
+				route.UpstreamID = globalUpstreamID
+				data.Routes = append(data.Routes, route)
+			}
+		}
+	}
+
+	return data, nil
+}
+
+// generate upstream from OpenAPI servers field
+func generateUpstreamByServers(servers openapi3.Servers, upstreamID string) (entity.Upstream, string) {
+	upstream := entity.Upstream{
+		BaseInfo: entity.BaseInfo{ID: upstreamID},
+		UpstreamDef: entity.UpstreamDef{
+			Name: upstreamID,
+			Type: "roundrobin",
+		},
+	}
+
+	var (
+		nodes  = make(map[string]float64)
+		scheme string
+		path   string
+	)
+	for _, server := range servers {
+		u, err := url.Parse(server.URL)
+		if err != nil {
+			continue
+		}
+
+		// save the scheme and path of the first valid server
+		// path may be empty
+		if scheme == "" {
+			scheme = u.Scheme
+			path = u.Path
+		}
+
+		nodes[u.Host] = 1
+	}
+
+	upstream.Scheme = scheme
+	upstream.Nodes = entity.NodesFormat(nodes)
+
+	return upstream, path
+}
+
+// generate base sample route for customize
+func generateBaseRoute(id string, desc string) entity.Route {
+	return entity.Route{
+		BaseInfo: entity.BaseInfo{ID: id},
+		Name:     id,
+		Desc:     desc,
+		Plugins:  make(map[string]interface{}),
+	}
+}
--- a/api/internal/handler/data_loader/loader/openapi3/import.go
+++ b/api/internal/handler/data_loader/loader/openapi3/import.go
@@ -17,14 +17,13 @@
 package openapi3
 
 import (
-	"errors"
-	"fmt"
 	"net/url"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/pkg/errors"
 
 	"github.com/apisix/manager-api/internal/core/entity"
 	"github.com/apisix/manager-api/internal/handler/data_loader/loader"
@@ -38,7 +37,7 @@ func (o Loader) Import(input interface{}) (*loader.DataSets, error) {
 
 	d, ok := input.([]byte)
 	if !ok {
-		return nil, fmt.Errorf("input format error: expected []byte but it is %s", reflect.TypeOf(input).Kind().String())
+		return nil, errors.Errorf("input format error: expected []byte but it is %s", reflect.TypeOf(input).Kind().String())
 	}
 
 	// load OAS3 document
@@ -49,7 +48,7 @@ func (o Loader) Import(input interface{}) (*loader.DataSets, error) {
 
 	// no paths in OAS3 document
 	if len(swagger.Paths) <= 0 {
-		return nil, consts.ErrImportFile
+		return nil, errors.Wrap(errors.New("OpenAPI documentation does not contain any paths"), consts.ErrImportFile.Error())
 	}
 
 	if o.TaskName == "" {
--- a/api/internal/handler/data_loader/loader/openapi3/import.go
+++ b/api/internal/handler/data_loader/loader/openapi3/import.go
@@ -62,7 +62,7 @@ func (o Loader) convertToEntities(s *openapi3.Swagger) (*loader.DataSets, error)
 		// temporarily save the parsed data
 		data = &loader.DataSets{}
 		// global upstream ID
-		globalUpstreamID = o.TaskName + "_global"
+		globalUpstreamID = o.TaskName
 		// global uri prefix
 		globalPath = ""
 	)
@@ -83,7 +83,7 @@ func (o Loader) convertToEntities(s *openapi3.Swagger) (*loader.DataSets, error)
 
 		// decide whether to merge multi method routes based on configuration
 		if o.MergeMethod {
-			// create a single route for each path
+			// create a single route for each path, merge all methods
 			route := generateBaseRoute(routeID, v.Summary)
 			route.Uris = []string{globalPath + realUri}
 			route.UpstreamID = globalUpstreamID
@@ -107,7 +107,9 @@ func (o Loader) convertToEntities(s *openapi3.Swagger) (*loader.DataSets, error)
 	return data, nil
 }
 
-// generate upstream from OpenAPI servers field
+// Generate APISIX upstream from OpenAPI servers field
+// return upstream and uri prefix
+// Tips: select only first server in servers field
 func generateUpstreamByServers(servers openapi3.Servers, upstreamID string) (entity.Upstream, string) {
 	upstream := entity.Upstream{
 		BaseInfo: entity.BaseInfo{ID: upstreamID},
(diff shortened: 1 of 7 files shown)


starsz (commented):
Do we have any proposal on this PR?

bzp2010:
> Do we have any proposal on this PR?

@starsz Yes, I was created a proposal for these works, you can see it here. 

https://github.com/apache/apisix-dashboard/issues/2465

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,180 @@
...
+		upstream, globalPath = generateUpstreamByServers(s.Servers, globalUpstreamID)
+		data.Upstreams = append(data.Upstreams, upstream)
+	}
+
+	// each one will correspond to a route
+	for uri, v := range s.Paths {
+		// replace parameter in uri to wildcard
+		realUri := regURIVar.ReplaceAllString(uri, "*")
+		// generate route name
+		routeID := o.TaskName + "_" + strings.NewReplacer("/", "-", "{", "", "}", "").Replace(strings.TrimPrefix(uri, "/"))
```
starsz, line 97: Why not use `base64` or hash to do this?

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,180 @@
...
+
+		// decide whether to merge multi-method routes based on configuration
+		if o.MergeMethod {
```
starsz, line 100: Why do we need a configure to do this ?

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,180 @@
...
+func generateUpstreamByServers(servers openapi3.Servers, upstreamID string) (entity.Upstream, string) {
+	upstream := entity.Upstream{
+		BaseInfo: entity.BaseInfo{ID: upstreamID},
+		UpstreamDef: entity.UpstreamDef{
+			Name: upstreamID,
+			Type: "roundrobin",
+		},
+	}
+
+	u, err := url.Parse(servers[0].URL)
```
starsz, line 158: Only parse servers[0] ?

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: Oh, Thanks for your reminder.

What do you think about using the base64 or hashed content as `ID` and the original uri as `Name`?

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: As I mentioned in the Proposal, different methods under the same URI may have different properties, for example:

```text
GET /user/{id}    Get a user's infomation
POST /user/{id}    Modify a user's infomation.

On POST method that need a Admin permission session, but GET not need it, so the `/user/{id}`'s method has different `securitySchema`.
```

To handle this case, we need to allow the user to choose not to merge methods.

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: This is based on the following considerations:

- The first servers are usually the node that is logically "used first".
- Different servers may have different global paths
```text
What's "global paths"?

servers:
- url: https://z.com/test/
- url: https://x.com/non-test/
- url: https://c.com/test-now/

According to the principle of APISIX, we need to set /test, /non-test,
/test-now, respectively, as the prefix of the routing uri of APISIX,
for example, set the original /user/* to /test/user/*, at this point
we can't operate.
```

Do you have any better suggestions?

api/internal/handler/data_loader/loader/openapi3/import.go
starsz: Maybe base64 is better. That's more easily to debug (we can decode from the id to get the info)

api/internal/handler/data_loader/loader/openapi3/import.go
starsz: Oh, get it.

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: After testing, using URI + base64 will generate some very long IDs, which may exceed the schema check limit of APISIX on the one hand, and confuse the format of ID display in the route list on the other. The symbols in base64 will also trigger the schema ID check of APISIX, while for Name only the length is required and the content is completely customizable.

image

So I took the following approach, instead of generating IDs, which are automatically generated by the dashboard storage layer, we store the information in the route name, which has this format `TaskName_URI[_Method]`. For a route with a URI of `/consumer/{cid}`, the method is GET and POST, it generates this `TaskName_consumer/{cid}[_GET/POST]`

```text
taskName: test

methodMerge: true
Method + URI:
GET|POST /consumer/{cid} => test_consumer/{cid}

methodMerge: false
Method + URI:
GET|POST /consumer/{cid} => test_consumer/{cid}_GET  +  test_consumer/{cid}_POST
```

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: I don't think this is appropriate, it generates a lot of useless routes. 🤔

starsz: Why the other routes are useless?

api/internal/handler/data_loader/loader/openapi3/import_test.go
```diff
@@ -0,0 +1,113 @@
...
+	// Upstream
+	assert.Equal(t, "https", data.Upstreams[0].Scheme)
+	assert.Equal(t, float64(1), data.Upstreams[0].Nodes.(map[string]float64)["api-101.glitch.me"])
+	assert.Equal(t, "test", data.Upstreams[0].Name)
+	assert.Equal(t, "roundrobin", data.Upstreams[0].Type)
+
+	// Route
+	assert.Equal(t, data.Upstreams[0].ID, data.Routes[0].UpstreamID)
+	for _, route := range data.Routes {
+		fmt.Println(route.Name)
```
starsz, line 96: Debug line?

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return data, nil
+}
+
+func (o Loader) convertToEntities(s *openapi3.Swagger) (*loader.DataSets, error) {
+	var (
+		// temporarily save the parsed data
+		data = &loader.DataSets{}
+		// global upstream ID
+		globalUpstreamID = o.TaskName
```
starsz, line 65: Use the taskName as upstream ID ?
It looks strange.

bzp2010: The taskname can be set to a service name, such as `user` or `order`, so that it will generate an upstream named `order` and generate a route name similar to `order_user/list`. I think this is clearer.

api/internal/handler/data_loader/loader/openapi3/import_test.go
bzp2010: Yes, removed

bzp2010:
### Update

The code for the auto-configuration plugin has been removed to reduce the complexity of the review. It will be implemented in a later PR.

api/internal/handler/data_loader/loader/openapi3/import.go
starsz: > This means that for a URI, when we have multiple servers fields, it will generate at most (number of servers * number of paths * number of methods); for example with 5 servers, 10 paths, each with 3 methods, at most it will generate 5 * 10 * 3 = 150 routes, which I think is a bad use of experience

I think the servers in open API should be treated equally.
If the user doesn't want to use the server[0], he should change all the path in the route.It's also a bad experience.

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/apisix/manager-api/internal/utils/consts"
+)
+
+func (o Loader) Import(input interface{}) (*loader.DataSets, error) {
+	if input == nil {
+		return nil, errors.New("input is nil")
+	}
+
+	// load OAS3 document
+	swagger, err := openapi3.NewSwaggerLoader().LoadSwaggerFromData(input.([]byte))
```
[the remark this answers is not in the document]
bzp2010: Get it

api/internal/handler/data_loader/loader/openapi3/import.go
nic-chen: I am a little confused about `servers`, should they be the URL exposed by the gateway or the URL of the API backend service, please confirm, thanks.

bzp2010:
> maybe we should create 10 routes only? each route has 3 methods and should create an upstream with 5 nodes.

Hi, @nic-chen.

Why you can't create multiple node upstreams is mentioned in another comment above. The short description is that the servers field in the OAS documentation supports not only filling in the domain, but also configuring the path while allowing to configure multiple addresses with different paths, we don't actually have the ability to handle this scenario, so only the first servers entry is used. https://github.com/apache/apisix-dashboard/pull/2460#discussion_r890004354

```text
servers:
- url: https://z.com/test/
- url: https://x.com/non-test/
- url: https://c.com/test-now/
```

api/internal/handler/data_loader/loader/openapi3/import.go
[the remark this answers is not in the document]
bzp2010: improved

api/internal/handler/data_loader/loader/openapi3/openapi3.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+type OpenAPISpecFileType string
+
+type Loader struct {
+	// MergeMethod indicates whether to merge routes when multiple HTTP methods are on the same path
+	MergeMethod bool
+	// TaskName indicates the name of current import/export task
+	TaskName string
```
tokers, line 31: How to decide on this field? Two scenarios should be confirmed:

1. Convert the same OAS file twice, with different `TaskName`, then duplicated routes will be created but their functionalities are the same.
2. Convert the two OAS files with the same `TaskName`, and the path and method are the same, will the routes conflict with each other?

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+func (o Loader) Import(input interface{}) (*loader.DataSets, error) {
+	if input == nil {
+		return nil, errors.New("input is nil")
+	}
+
+	d, ok := input.([]byte)
+	if !ok {
+		return nil, fmt.Errorf("input format error: expected []byte but it is %s", reflect.TypeOf(input).Kind().String())
```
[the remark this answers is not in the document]
bzp2010: @tokers 🤔

For the input data error, should we classify it as a `400 Bad Request` instead of a 5xx server-related error? I don't think this should be a programming error on the server.
From a practical point of view, if we use `panic` here, the Manager API will simply treat it as a `500 Internal Server Error` and not automatically select the appropriate response code or have us specify it.

api/internal/handler/data_loader/loader/openapi3/openapi3.go
bzp2010: @tokers 

For this module, it is not concerned with duplicate routes, but only with the parsing and conversion of OAS documents, and for the parts of the duplicate routes that you pointed out as limitations, you can see them here #2474 : https://github.com/apache/apisix-dashboard/blob/7fb05943de573d782b42c18728e4dc9990e04cf3/api/internal/handler/data_loader/route_import.go#L149-L197

api/internal/handler/data_loader/loader/openapi3/import_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/apisix/manager-api/internal/core/entity"
+)
+
+var (
+	TestAPI101 = "../../../../../test/testdata/import/Postman-API101.yaml"
```
nic-6443, line 29: We can use `go:embed` to avoid reading file manually.

api/test/testdata/import/Postman-API101.yaml
```diff
@@ -0,0 +1,152 @@
...
+      responses:
+        '200':
+          description: Successful response
+          content:
+            application/json: {}
+  /customer/{customer_id}:
+    put:
+      tags:
+        - default
+      summary: Update customer
```
starsz, line 102: Do we support parse `summary` ?

api/internal/handler/data_loader/loader/openapi3/import_test.go
tao12345666333: +1

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,149 @@
...
+		upstream, globalPath = generateUpstreamByServers(s.Servers, globalUpstreamID)
+		data.Upstreams = append(data.Upstreams, upstream)
+	}
+
+	// each one will correspond to a route
+	for uri, v := range s.Paths {
+		// replace parameter in uri to wildcard
+		realUri := regURIVar.ReplaceAllString(uri, "*")
+		// generate route Name
+		routeName := o.TaskName + "_" + strings.TrimPrefix(uri, "/")
```
tao12345666333, line 89: I think we can follow APISIX's schema to check resources.

api/internal/handler/data_loader/loader/openapi3/import.go
nic-chen: hi @bzp2010, please check this, it affects the overall convert

api/internal/handler/data_loader/loader/openapi3/openapi3.go
tokers: OK.

api/internal/handler/data_loader/loader/openapi3/import_test.go
```diff
@@ -0,0 +1,109 @@
...
+			assert.Contains(t, route.Uris, "/customer/*")
+			assert.Contains(t, route.Methods, "PUT")
+			assert.Equal(t, "Update customer", route.Desc)
+			assert.Equal(t, entity.Status(0), route.Status)
+		case "test_customer/{customer_id}_DELETE":
+			assert.Contains(t, route.Uris, "/customer/*")
+			assert.Contains(t, route.Methods, "DELETE")
+			assert.Equal(t, "Remove customer", route.Desc)
+			assert.Equal(t, entity.Status(0), route.Status)
+		}
```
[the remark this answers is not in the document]
bzp2010: default branch added

api/internal/handler/data_loader/loader/openapi3/import.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// load OAS3 document
+	swagger, err := openapi3.NewSwaggerLoader().LoadSwaggerFromData(d)
+	if err != nil {
+		return nil, err
+	}
+
+	// no paths in OAS3 document
+	if len(swagger.Paths) <= 0 {
+		return nil, consts.ErrImportFile
```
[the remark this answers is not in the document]
bzp2010: error context improved

api/internal/handler/data_loader/loader/openapi3/openapi3.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type PathValue struct {
+	Method string
+	Value  *openapi3.Operation
+}
+
+var (
+	regURIVar              = regexp.MustCompile(`{[\w.]*}`)
```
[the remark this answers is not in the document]
bzp2010: The content in the search engine says `\w` contains `.`. However, after testing, if I use the regular expression `{\w*}`, it doesn't actually match, so I need to manually append the symbols `.`.

Use `{\w*}`:
image

Use `{[\w.]*}`
image

So I don't think it needs to be modified? If I have any other mistakes, please point them out.

api/test/testdata/import/Postman-API101.yaml
bzp2010: @starsz This feature is not yet supported, maybe we can put it in the `desc` field.

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: @tao12345666333 Yes, the ability to check the schema is already built into the storage layer, so no matter what APISIX resource we try to create, it will perform the schema check.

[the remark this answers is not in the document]
bzp2010: I took a look at the name length limit in APISIX and it has a maximum value of 100, so it is indeed possible to trigger this limit. do you have a better idea for the route name generation rules.

That's the way now:
https://github.com/apache/apisix-dashboard/pull/2460#issuecomment-1152153563

api/internal/handler/data_loader/loader/openapi3/import.go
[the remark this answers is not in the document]
bzp2010: updated the input error to panic

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: > I am a little confused about servers, should they be the URL exposed by the gateway or the URL of the API backend service, please confirm, thanks.

@nic-chen We can look at it this way, during API debugging it is the address actually requested by the debugger, so it should be the address provided by APISIX to the outside world, and the original value should be upstream.

我们可以这样看待这个问题，在API调试过程中，它是调试器实际请求的地址，因此它应当是APISIX对外提供的地址，而原始值应该是上游的。

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: BTW, I'm still not sure on the route name generation and can see from https://github.com/apache/apisix-dashboard/pull/2460#issuecomment-1152153563, do you have any suggestions?

api/internal/handler/data_loader/loader/openapi3/openapi3.go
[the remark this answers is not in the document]
nic-6443: What I mean is `. ` is enough. You can this: `{.*?}`

api/internal/handler/data_loader/loader/openapi3/import_test.go
nic-6443: OK, let's do it in another PR.

api/internal/handler/data_loader/loader/openapi3/import.go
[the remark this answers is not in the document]
nic-6443: I can not think of any better way for the time being, if you want more readability, it does make sense to generate from the URL, otherwise directly use a hash function to generate a fixed length unique string.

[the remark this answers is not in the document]
bzp2010: Yes, I don't have a better idea, so I'll keep the current Name implementation for now and adjust it later based on feedback.

api/internal/handler/data_loader/loader/openapi3/openapi3.go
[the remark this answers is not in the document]
bzp2010: @nic-6443 Oh, indeed, I will revise it.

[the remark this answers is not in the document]
bzp2010: modified

api/internal/handler/data_loader/loader/openapi3/import.go
nic-chen: @bzp2010 I suggest that this PR not deal with `servers` first, and deal with it later if there is a strong demand; instead of dealing with imperfections now.

nic-6443 (approved):

api/internal/handler/data_loader/loader/openapi3/import.go
bzp2010: @nic-chen Well, I decided to remove this feature and replace it with a **default empty upstream** that the user can simply modify to change the upstream associated with all routes.

nic-chen (approved):

SkyeYoung (approved):

juzhiyuan (approved):

Issue #2465, opened 2022-06-06
Proposal: refactor OpenAPI 3 parse and convert


bzp2010:
### Background

Although APISIX Dashboard supports OpenAPI3 specification, it is actually designed to export from APISIX and then import (even it doesn't do well in this area), it has poor support for importing standard OpenAPI3 documents, and we need to improve this.

The following problems exist in the current implementation:

- Cannot import if `operationId` field does not exist

This is our original implementation of the field added to the "export", in line with the OAS specification, but other documents sometimes do not set this value, resulting in the inability to generate route name, the schema will check failure.

- Uri generation error

When a single path has more than one `{xxx}` variable defined, wildcards are truncated directly after the first uri, and the remaining parts of the uri are ignored

```text
Incorrent:
/test/{root}/test/{sub} => /test/*

Corrent:
/test/{root}/test/{sub} => /test/*/test/*
```

- Different methods of the same path are generated as multiple routes

When there are multiple HTTP Methods on the same path, multiple routes will be generated. (Current improvements may have been made, but they do not fully meet the requirements.)

- OAS's `Servers` field is ignored

Currently only focus on custom configurations like `x-apisix-upstream`, ignoring the `servers` field in the OAS specification.

- Over-coupled code

The current import logic is heavily coupled with the handler and cannot be separated, and is not decoupled using the currently defined `Loader` interface.

Therefore, this **Proposal** will fix the issues.

### Scheme

- Cannot import if `operationId` field does not exist

No longer relies on `operationId` to generate route id and name.

Perhaps we can use uri as a material to generate id and name. For example, the following rules:

```text
/test/{user_id}/info => ID: test-user_id-info
```

No better ideas at the moment, hope to get good methods.

- Uri generation error

Correction generation algorithm

- Different methods of the same path are generated as multiple routes

Provide the option for the user to specify whether to merge or not (merging will result in loss when different methods have different security definitions, so it needs to be allowed without merging)

- OAS's `Servers` field is ignored

Prefer to generate upstreams using the Servers field, and generate empty upstreams when servers do not exist.

- Over-coupled code

Using the Loader interface to implement functionality, the original handler implementation will be preserved in this PR and migrated to the Loader implementation in the next PR and marked as deprecated.

### Timetable

- [ ] Implementing new decoder and converter: Ensure that API 101OAS documents converted by Postman can be parsed
- [ ] Migration of existing implementations to the Loader interface
[...]

PR #2468, opened 2022-06-07
feat: migrate old OpenAPI implementation to Loader interface

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Migrate legacy code to the Loader interface implementation to decouple the handler from the import and export logic

**Related issues**

a part of #2465 

**Checklist:**


bzp2010:
### Update

The original OpenAPI import and export logic has been migrated to the Loader interface implementation, including the associated test cases.
However, it should be noted that only code migration has been done in this PR, not removing the original coupling with the Handler, which is still there; the Handler is too coupled and will be completely rewritten later to adapt to the implementation of the Loader interface.

### Desc

The migration in this PR is consistent with the original code, with no changes to the overall logic and only minor adjustments to adapt the Loader interface; the test cases have also been modified to allow testing the import and export modules separately without including logic at the level of the HTTP application framework (Gin). And, once a huge test case file with 2000 lines of code was split into multiple code files by module.

**Therefore, in this PR, the import and export logic coupled with the Gin handler is split out and provided as a separate loader package.**

nic-chen:
large PR, maybe we could split it into small PRs? for example, at least two, import and export.

starsz:
> large PR, maybe we could split it into small PRs? for example, at least two, import and export.

Agree +1

bzp2010:
> large PR, maybe we could split it into small PRs? for example, at least two, import and export.

Hi, @nic-chen.

This one PR contains only the old code and tests for the migration, which are essentially replicated with no core modification logic and only a few modifications for adaptation reasons.

This PR has a huge number of lines, mainly test cases, the original test cases are up to 2000 lines, I split them into several test files for different purposes; the core code has about 500 lines, only a little traversal code is added in the function entry part; the remaining part is the license header.

So I think it's important to focus only on the part I added, which is [codes](https://github.com/apache/apisix-dashboard/blob/ad888d0753726b017335090c94113651b1d55cc4/api/internal/handler/data_loader/loader/openapi_legacy/export.go#L51-L68); and no need to split the PR.

bzp2010:
### Update

#### I'm even thinking about whether it's necessary for it to continue to exist, it only exists to ensure partial compatibility, and given its poor implementation, it's difficult for me to devote effort to maintaining it (The logic of the code is confusing, containing various global and local variables, nested loops; the tests are even run completely on mocking, and the test data they provide is not executed correctly in real situations at all.), and this old implementation will probably be removed later (when we have a better implementation).
#### This PR is currently low priority, and I will deal with this at the end if the current PR is not available for review.

cc @nic-chen @starsz

starsz:
> ### Update
> #### I'm even thinking about whether it's necessary for it to continue to exist, it only exists to ensure partial compatibility, and given its poor implementation, it's difficult for me to devote effort to maintaining it (The logic of the code is confusing, containing various global and local variables, nested loops; the tests are even run completely on mocking, and the test data they provide is not executed correctly in real situations at all.), and this old implementation will probably be removed later (when we have a better implementation).
> #### This PR is currently low priority, and I will deal with this at the end if the current PR is not available for review. Several other PRs have higher priority, such as #2460 #2474
> cc @nic-chen @starsz

Got it.

PR #2525, opened 2022-07-15
test: move service unit cases to e2e

bzp2010:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

Migrating unit tests to E2E implementations


nic-chen:
It seems that the unit test has been migrated to e2e recently,
The purpose of the two tests is not the same, and the scope of coverage is also different,
I would like to know why are we doing this and is there some background clarification?

bzp2010:
> It seems that the unit test has been migrated to e2e recently, The purpose of the two tests is not the same, and the scope of coverage is also different, I would like to know why are we doing this and is there some background clarification?

Hi, @nic-chen.

### Reasons

If we read those unit tests carefully, we will find that it is actually E2E tests using mock, so why don't we just migrate to E2E?

On the other hand, it costs us more to keep the current unit tests. For example, in order to start a test, we had to complete some of the underlying initialization work, such as configuration files, environment variables, etc. This probably defeats the original purpose of unit testing and also brings a lot of trouble for our CI.

For example, it is difficult to find the corresponding configuration file (because go test actually runs in the same directory as the test.go file, and it cannot search the specified directory). Some hack code for the init function must not be added to the configuration file module.

### So

For a variety of reasons, these virtually meaningless pseudo-E2Es will be migrated to true E2E processes. It will help us to focus on the parts that really need unit testing and enhance the use case coverage of E2E tests.

nic-chen:
> 

Thanks for your sharing!
It would be better if there is an issue and a checklist of what needs to be migrated in it.
Maybe someone else in the community would like to get this done together.

guoqqqi (approved):

Baoyuantop (approved):

PR #2564, opened 2022-08-01
upstream discovery allow server name pass from request host

yangboyd:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

upstream discovery allow server name pass from request host
add service_name_type field with value: pass, rewrite.
The original service_name field is used as rewrite service name.

**Related issues**

fix/resolve #0001

**Checklist:**

### DIFF
--- a/api/conf/schema.json
+++ b/api/conf/schema.json
@@ -574,7 +574,7 @@
 						{
 							"required": [
 								"discovery_type",
-								"service_name",
+								"service_name_type",
 								"type"
 							]
 						}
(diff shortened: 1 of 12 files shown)


api/conf/schema.json
```diff
@@ -574,7 +574,7 @@
 						{
 							"required": [
 								"discovery_type",
-								"service_name",
+								"service_name_type",
```
bzp2010, line 577: Please do not modify this file, it is generated directly by the APISIX Control API and should not be edited by a human.

bzp2010:
Hi, @yangboyd.

I noticed that you updated `schema.json` and some Struct definitions which should not be modified manually, so I think this PR is inappropriate.
If you need this feature, I'm afraid you'll have to do the entire data transformation on the front end instead of passing it to the back end.

juzhiyuan:
Thank you for your PR. We will create the new dashboard to resolve this issue. Refer to https://github.com/apache/apisix-dashboard/issues/2981

PR #2459, opened 2022-05-30
chore: move yarn to pnpm

bzp2010:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

Migrate yarn to pnpm to reduce dependency installation and compilation time.

**Checklist:**

- [x] Did you explain what problem does this PR solve? Or what new features have been added?
- [x] Have you added corresponding test cases?
- [x] Have you modified the corresponding document?
- [x] Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

PR #2569, opened 2022-08-05
chore: pnpm will be an optional manager（WIP）

FangSen9000:

- [ ] Bugfix
- [x] New feature provided
- [x] Improve performance
- [ ] Backport patches

Objective: the purpose of this PR is to introduce the advanced and popular pnpm package management tool, and use pnpm to download dependency packages faster without affecting the performance of the original yarn.

Effect: pnpm will be an optional and supplementary package management tool.

Change: I modified the packgae.json package and added some dependency packages missing from pnpm; Relevant documents are supplemented;

----------------------------------------------------------------------------
image
                                                                        figure1:  Changed build document

                                                                        figure2:  Changed developer document
-----------------------------------------------------------------------------
Supplementary information:

1. Yarn and his former employees will not be affected

2. When people using pnpm add new dependency packages, they will automatically generate pnpm-lock.yaml and packgae.json. Like yarn, when someone adds a new dependency package while working, it is not necessary to submit lock.yaml, just submit package.json like yarn. There is no need to remind and worry, just "work as usual!"

3. People using pnpm can delete yarn.lock or not. Package.json is the basis for synchronizing pnpm and yarn dependent packages, and this work is automated

4. If you want to use pnpm lock, you can convert one from lock.lock.  (even for APISIX developers, this thing does not need to be submitted to the warehouse, and our dependent package anchor is still package.json,, however, this can give some convenience to local developers )

5. The above information has been written in the document

**Checklist:**


FangSen9000:
@guoqqqi @bzp2010 @Baoyuantop 
When free, I hope you can give me some suggestions. The purpose of this PR is to allow local developers or users to use pnpm instead of yarn as a second option.

The document and package.json file have been changed. Lock has uploaded the latest one. You can view the changes through screenshots.

FangSen9000:
My work is carried out on the basis of predecessors. He failed, but I benefited a lot from his work. Then I also carried out more in-depth work.

@bzp2010 ‘s work：
https://github.com/apache/apisix-dashboard/pull/2459/files#
my work：
https://github.com/apache/apisix-dashboard/pull/2555/files

This work consists of two parts, GitHub workflow adjustment and project itself adjustment. The latter is easier, but the former is more difficult. Because our workflow needs to change too much.

After I spent 14 days in-depth, I found that there were too many GitHub workflows to be changed, and many of them could not be replaced and run temporarily. The workload was amazing. So I took this helpless solution: instead of completely replacing the package manager, I used pnpm as a supplementary solution.

FangSen9000:
> 文档和 package.json 文件已更改。锁定已上传最新的一个。您可以通过屏幕截图查看更改。

SkyeYoung:
It doesn't look like it's possible to use pnpm and yarn at the same time 🤔

FangSen9000:
> It doesn't look like it's possible to use pnpm and yarn at the same time 🤔

When users or developers download it, they can use pnpm to download the dependent package. Pnpm uses package JSON. The command uses pnpm (although it essentially uses the package manager of the stored script)

Without the two dependency packages I added, pnpm users will spend a lot of time processing these packages. Those who download apifix dashboard using pnpm for the first time will spend a lot of time on them and will face three issues of pnpm:

1. Npmirc file setting problem
2. Peer dependency will not be automatically installed
3. Finally, there is still a lack of dependent packages

These documents will help them very well. However, the local developers or users of pnpm finally enjoy the fast download speed of dependent packages.

SkyeYoung:
@FangSen9000 But you don't keep the file pnpm-lock.yaml, which leads to two problems for developers who want to use it.

1. there are multiple versions of the dependencies used, which can lead to inconsistent versions being used by developers, making the entire software environment unstable.
2. The need to exclude pnpm-lock.yaml for each commit makes it more difficult for developers who are just starting out to contribute.

The above problems make it difficult to use "pnpm as an optional package manager", which leads to no one using pnpm as a package manager, and ultimately makes the whole PR ineffective. 
On reflection, it is also very difficult to maintain dependencies for both package managers, so I suggest you try replacing yarn with pnpm instead of trying to make them co-exist.

FangSen9000:
> @FangSen9000但是您不保留文件pnpm-lock.yaml，这给想要使用它的开发人员带来了两个问题。

Ans1: I pointed out in the document that the required person can generate a pnpm lock through pnpm import: https://pnpm.io/cli/import
Ans2: it's OK to hand it in, because both yen.lock and pnpm lock will follow the package JSON update. And there will be no conflict between the two in the project. I tried. If not, I don't think it would be strange.

If it is completely replaced, we may have to give up the node JS 12. X support
When I restart the old branch and repeat the error, I will ask BZP for his opinion. However, the 15 or so format detection errors on the road are also a big project.

SkyeYoung:
> Ans1: I pointed out in the document that the required person can generate a pnpm lock through pnpm import: [pnpm.io/cli/import](https://pnpm.io/cli/import)
> Ans2: it's OK to hand it in, because both yen.lock and pnpm lock will follow the package JSON update. And there will be no conflict between the two in the project. I tried. I just think it would be strange for two locks to appear at the same time, but they do not interfere with each other.

This is exactly what I mentioned: it can cause problems with inconsistent dependency versions. The problem is not encountered now only because we are under a stable version and subsequent maintenance would be very difficult.

FangSen9000:
WIP, do not merge yet

PR #2570, opened 2022-08-06
feat: add Admin API storage implementation

bzp2010:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Previously, APISIX Dashboard connected directly to etcd and wrote data, now it supports connecting to APISIX Admin API.

**Checklist:**

### DIFF
--- a/api/internal/core/storage/adminapi.go
+++ b/api/internal/core/storage/adminapi.go
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package storage
 
 import (
--- a/api/internal/core/storage/adminapi.go
+++ b/api/internal/core/storage/adminapi.go
@@ -1,0 +1,173 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/go-resty/resty/v2"
+	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
+
+	"github.com/apache/apisix-dashboard/api/pkg/storage"
+)
+
+const (
+	adminAPIErrorWarp = "failed to request Admin API"
+)
+
+type adminAPIVersion = int
+
+const (
+	adminAPIVersion2 adminAPIVersion = 2
+	adminApiVersion3 adminAPIVersion = 3
+)
+
+// Ensure that the AdminAPI Storage implementation conforms to the storage_api interface definition
+var _ storage_api.Interface = AdminAPIStorage{}
+
+// NewAdminAPIStorage will create an instance of AdminAPIStorage and complete the basic configure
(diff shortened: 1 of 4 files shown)


bzp2010:
### Tips

This PR adds a storage implementation of the Admin API, but there are no tests associated with it because it requires a dependency on APISIX, but functions with the original implementation of etcd, so it is applicable to the original etcd use case and we will switch to this storage implementation in CI in the future.

api/internal/core/storage/adminapi.go
```diff
@@ -0,0 +1,190 @@
...
+var _ storage_api.Interface = AdminAPIStorage{}
+
+// NewAdminAPIStorage will create an instance of AdminAPIStorage and complete the basic configure
+func NewAdminAPIStorage() *AdminAPIStorage {
+	return &AdminAPIStorage{
+		client: resty.New().
+			SetBaseURL("http://127.0.0.1:9080/apisix/admin/").
+			SetHeader("X-API-KEY", "edd1c9f034335f136f87ad84b625c8f1").
+			SetTimeout(1 * time.Second),
+	}
```
Baoyuantop, line 54: Do we need to move them into the config file?

ghost (approved):
LGTM.

api/internal/core/storage/adminapi.go
bzp2010: Hi, @Baoyuantop 

Of course it is, but we can't do them all in one PR, so let's start with the most generic configuration and it will be changed later.

Reference: https://github.com/apache/apisix-dashboard/pull/2570#issuecomment-1207185447

api/internal/core/storage/adminapi.go
```diff
@@ -0,0 +1,190 @@
...
+// Watch watches for changes in data sources and generates event streams
+// Admin API storage implementation does not support Watch capability
+func (s AdminAPIStorage) Watch(_ context.Context, _ string) <-chan storage_api.WatchResponse {
+	return make(chan storage_api.WatchResponse)
+}
+
+// removeKeyPrefix will remove the "/apisix/" prefix of key
+// the pattern of the key is usually like "/apisix/routes"
+func (AdminAPIStorage) removeKeyPrefix(s string) string {
+	return strings.ReplaceAll(s, "/apisix/", "")
```
nic-chen, line 147: the prefix needs to be configurated too.

api/internal/core/storage/adminapi.go
```diff
@@ -0,0 +1,190 @@
...
+		if gjson.Get(resp.String(), "deleted").Int() < 1 {
+			return errors.Errorf("key: %s is not found", key)
+		}
+	}
+
+	return nil
+}
+
+// Watch watches for changes in data sources and generates event streams
+// Admin API storage implementation does not support Watch capability
```
nic-chen, line 139: So how should we synchronize incremental changes? Do we have a proposal for admin api storage?

api/internal/core/storage/adminapi.go
bzp2010: Hi, @nic-chen 

Yes, same with https://github.com/apache/apisix-dashboard/pull/2570#discussion_r939836946, it will be finish in next PR, we only use a default value in this PR.

api/internal/core/storage/adminapi.go
bzp2010: This `Watch` is used to update the cache in the `GenericStore`, and I will add a switch to it to allow it to turn off the cache to fetch data directly from the data source rather than through the Watch's caching mechanism.

bzp2010: Or even if we don't continue to use Etcd directly as a data source, we can just remove it.

bzp2010: Here I only leave an empty implementation, if you must implement it will be more complicated.

Use timers to fetch data in full at regular intervals to update the cache, and when an operation to update the data is triggered, the data is updated in full, compared with the value in memory, and converted to a change event to be returned to the upper layer.

Baoyuantop (approved):

api/internal/core/storage/adminapi.go
nic-chen: > if you must implement it will be more complicated.

no. just make sure you have a plan for it.

nic-chen (approved):

PR #2590, opened 2022-08-14
chore: Move from yarn to pnpm (WIP)

FangSen9000:

- [ ] Bugfix
- [x] New feature provided
- [x] Improve performance
- [ ] Backport patches

Objective: the purpose of this PR is to introduce advanced and popular pnpm package management tools and use pnpm to download dependency packages faster.

Effect: pnpm will be the next generation package management tool of dashboard. Migrate the package manager from the original yarn to pnpm to reduce dependency installation and compilation time.

Changes: package.json, pnpm-lock.yaml, related GitHub workflow files. The relevant documents have been updated, solutions have been written for possible problems in use, and the workflow is running normally.

**The related PR**

https://github.com/apache/apisix-dashboard/pull/2459

**Checklist:**

### DIFF
--- a/web/cypress/integration/consumer/create-consumer-with-api-breaker-plugin-form.spec.js
+++ b/web/cypress/integration/consumer/create-consumer-with-api-breaker-plugin-form.spec.js
@@ -50,7 +50,6 @@ context('Create and delete consumer with api-breaker plugin form', () => {
   it('creates consumer with api-breaker form', function () {
     cy.visit('/');
     cy.contains('Consumer').click();
-    cy.wait(3000);
     cy.get(selector.empty).should('be.visible');
     cy.contains('Create').click();
     // basic information
@@ -111,7 +110,6 @@ context('Create and delete consumer with api-breaker plugin form', () => {
 
   it('delete the consumer', function () {
     cy.visit('/consumer/list');
-    cy.wait(3000);
     cy.contains(data.consumerName).should('be.visible').siblings().contains('Delete').click();
     cy.contains('button', 'Confirm').click();
     cy.get(selector.notification).should('contain', data.deleteConsumerSuccess);
--- a/web/cypress/integration/consumer/create-consumer-with-basic-auth-plugin-form.spec.js
+++ b/web/cypress/integration/consumer/create-consumer-with-basic-auth-plugin-form.spec.js
@@ -47,7 +47,6 @@ context('Create and delete consumer with basic-auth plugin form', () => {
   it('creates consumer with basic-auth form', function () {
     cy.visit('/');
     cy.contains('Consumer').click();
-    cy.wait(3000);
     cy.get(selector.empty).should('be.visible');
     cy.contains('Create').click();
     // basic information
@@ -111,7 +110,6 @@ context('Create and delete consumer with basic-auth plugin form', () => {
 
   it('delete the consumer', function () {
     cy.visit('/consumer/list');
-    cy.wait(3000);
     cy.contains(data.consumerName).should('be.visible').siblings().contains('Delete').click();
     cy.contains('button', 'Confirm').click();
     cy.get(selector.notification).should('contain', data.deleteConsumerSuccess);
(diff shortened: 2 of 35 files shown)


.github/workflows/auto-build-rpm.yml
```diff
@@ -49,7 +56,7 @@ jobs:
         run: |
           export VERSION=${{ steps.branch_env.outputs.version }}
           sudo gem install --no-document fpm
-          git clone https://github.com/api7/apisix-build-tools.git
+          git clone https://github.com/FangSen9000/apisix-build-tools.git
```
FangSen9000, line 59: Note: This needs to be changed before merging. I link to the build script that I have changed. And we must let the API7 community through my changes and upgrades in them to change back.

Dockerfile
```diff
@@ -20,7 +20,7 @@ ARG APISIX_DASHBOARD_VERSION=master
 
 RUN set -x \
     && apk add --no-cache --virtual .builddeps git \
-    && git clone https://github.com/apache/apisix-dashboard.git -b ${APISIX_DASHBOARD_VERSION} /usr/local/apisix-dashboard \
+    && git clone https://github.com/FangSen9000/apisix-dashboard.git -b ${APISIX_DASHBOARD_VERSION} /usr/local/apisix-dashboard \
```
FangSen9000, line 23: If I keep it as it is, the original warehouse that has not been changed will only report an error. In order to prove that my change is effective, I link to my changed warehouse. Just like RPM workflow. Before merging, please remind me to change it back.

.github/workflows/auto-build-rpm.yml
FangSen9000, line 59: Note that we should let the API7 community merge my changes to the build script before merging this pr. (if I keep it as it is, the original warehouse that has not been changed will only report an error. In order to prove that my changes are effective, I link to my modified build script warehouse.)

FangSen9000:
@bzp2010 I have finished all the work based on your work. You can have a look

FangSen9000:
image
@SkyeYoung 
Have you ever encountered this situation? The local test case is running well, and no exception is found by the naked eye. However, the test case is not running, and it is not a failure. If it fails, he should prompt several assertions.

It should not be started, but it seems that it is not different from other test cases. Cypress is useless. It took half a day to find out the problem.

SkyeYoung:
> Have you ever encountered this situation? The local test case is running well, and no exception is found by the naked eye. However, the test case is not running, and it is not a failure. If it fails, he should prompt several assertions.
> 
> He didn't start up, but it seems that it is not different from other test cases. Cypress is useless. It took half a day to find out the problem, but fail. This is the last question in this work.

cc @bzp2010 @Baoyuantop

.github/workflows/frontend-e2e-test.yml
```diff
@@ -49,8 +49,9 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 16.x
-          cache: 'yarn'
-          cache-dependency-path: web/yarn.lock
+
+      - name: Install PNPM
+        run: npm i -g pnpm@6.32.11
```
SkyeYoung, line 54: Have you checked out this https://github.com/pnpm/action-setup? I think this might be a little better.

But, anyway, let's run through the test first.

FangSen9000: Let me first say a good news. Yesterday's problem has been solved. The reason is that he didn't use Chrome browser to run it. Just add a setting.

`"For example, I just found that it seems that many errors in file format are related to yarn. Many workflow files have been perfected around this. Then, the pnpm directly downloaded by GitHub does not seem to support node.js12. It describes that the downloaded pnpm does not support NPM below 14 (in fact, you can install a low-level package and ask him to support it, but it feels that it will be troublesome. I am afraid there are more problems.) Only 14 and 16 can be built when building tests. "
1ec791081ecc16fd7dbe07d687f54ea
0ca067b71192e4dc0c2c31de209e9a0

"When downloading the relevant environment, it seems that there are some restrictions. Maybe it can be solved by storing it in the warehouse through some link, but I haven't started to look at it. I'm afraid this is also very troublesome. This picture is the reason why the three people are useless. Action is prohibited. It clearly meets the requirements he said, but it can't be used."`

@SkyeYoung I mentioned this problem. You can look at the pictures. If you can solve it, it's best.

PR #2601, opened 2022-08-23
fix: change default CSP value

nthsky:

- Bugfix

It changes dashboard default CSP. Add `unsafe-inline` to `script-src` and add `img-src`

Current default csp policy is 
```
"default-src 'self'; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
```

**Problems:**

1. script-src without "unsafe-inline" cause error like this:
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-9B0adsvsNyel0bNudomHjhrSk758zuql+j59Udq5xCo='), or a nonce ('nonce-...') is required to enable inline execution.
```
2. no img-src cause that the plugin page without icon using base64 img cannot show correctly.

**Changes:**

After, default CSP changes to 
```
"default-src 'self'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data:"
```


bzp2010 (approved):

LiteSun (approved):

Baoyuantop (approved):

bzp2010:
Thanks for your contribution.

Issue #2360, opened 2022-03-03
Cannot get latest data after ETCD recovery


yangvipguang:
### Issue description

dashboard版本：2.10.1
ETCD版本：3.5

目前我测试了一下，如果我要备份etcd，然后删除掉个别配置（比如upsteam等），如果我要恢复之前备份的etcd数据，这个时候dashboard 不能实时更新显示恢复之前的数据，需要重启一下dashboard的服务才可以，不知道这算bug还是就预期内的。

### Expected behavior

目前我测试了一下，如果我要备份etcd，然后删除掉个别配置（比如upsteam等），如果我要恢复之前备份的etcd数据，这个时候dashboard 不能实时更新显示恢复之前的数据，需要重启一下dashboard的服务才可以，不知道这算bug还是就预期内的。

### How to Reproduce

1、备份etcd数据
2、备份后修改一些数据（比如upstream等）
3、恢复etcd数据
4、刷新dashboard页面

### Environment

- apisix version (cmd: `apisix version`):
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `etcd --version`):
- apisix-dashboard version, if have:
- Browser version, if have:

jwrookie:
looks not a dashboard bug, the data is fetched from etcd in real time. 
If you want, we can communicate by voice

Baoyuantop:
This seems to be as expected, there is a caching layer between Dashboard and etcd.

zaunist:
Hi, @yangvipguang. We need to discuss how to solve this problem, anyone is welcome to make suggestions!

yangvipguang:
@zaunist  ok， I'm going to try a few times to see if I can reproduce this problem every time

jwrookie:
> This seems to be as expected, there is a caching layer between Dashboard and etcd.

you are right, we have a layer of cache

bzp2010:
What kind of backup/restore means do you use? I don't know much about etcd and am not sure that data changes in any case would trigger an Watch push.

jwrookie:
According to etcd doc, `Disaster recovery` requires a restart of etcd, which means that all previous client links will be invalidated, and the old data you see is just the program's cache

As far as I know, `grpc-go` will manage failed layer 4 connections, so when `Disaster recovery` occurs, we reload the local cache and the problem will solved

bzp2010:
[...]

PR #2650, opened 2022-10-24
feat: add etcd store auto re-initialize

### LINKED ISSUE #2461: cache failed to update when etcd restarted
### Issue description

when I restart etcd, dashboard occure the error: 
2022-05-31T10:27:32.003+0800    ERROR   storage/etcd.go:245     channel canceled
2022-05-31T10:27:32.003+0800    WARN    store/store.go:124      watch failed: channel canceled

then, cache data is no longer updated

### Expected behavior

os: linux
version: 2.11

### How to Reproduce

just restart etcd

### Screenshots

_No response_

### Environment

- apisix version (cmd: `apisix version`):
- OS (cmd: `uname -a`):
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`):
- etcd version, if have (cmd: run `etcd --version`):
- apisix-dashboard version, if have:
- Browser version, if have:

### Additional context

_No response_

bzp2010:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Resynchronize when the etcd local cache does not match the data source.

**Related issues**

fix #2461 #2360 and more same problem

**Checklist:**


bzp2010:
Updated:

api/cmd/root.go
```diff
@@ -81,3 +89,57 @@ func manageAPI() error {
...
+
+				// After multiple failures, the connection is restored
+				if unavailableTimes >= 1 {
+					log.Warnf("etcd connection recovered, but after several connection losses, reinitializing stores, times: %d", unavailableTimes)
+					unavailableTimes = 0
+
+					// When this happens, force a full re-initialization of the store
+					store.RangeStore(func(key store.HubKey, store *store.GenericStore) bool {
+						log.Warnf("etcd store reinitializing: resource: %s", key)
+						_ = store.Init()
```
nic-chen, line 119: should also log the err

bzp2010:
### Updated

- Update etcd dependencies to v3.5.5
- Add passive live check for etcd watch
- Add active connection check for all etcd stream
- Add cli test for active check
- Improve some codes

api/cmd/root.go
```diff
@@ -81,3 +89,57 @@ func manageAPI() error {
...
+	ctx, cancel := context.WithCancel(context.TODO())
+	unavailableTimes := 0
+
+	go func() {
+		etcdClient := storage.GenEtcdStorage().GetClient()
+		for {
+			select {
+			case <-time.Tick(10 * time.Second):
+				sCtx, sCancel := context.WithTimeout(context.TODO(), 5*time.Second)
+				err := etcdClient.Sync(sCtx)
```
nic-chen, line 103: oh, I think it is unnecessary. Because the client has a retry mechanism, even if the connection is interrupted during operation, reusing the client will automatically recover, only the `watch` cannot recover automatically in some cases

bzp2010: The connection can be automatically restored, but the behavior of watch is unobservable, and sometimes watch is incapacitated, but it still does not actively quit. There is no API to actively know which watchers are currently available and what is the state of those watchers.
This is a complete black box for its applicators, the entire documentation of etcd mentions that its connections are self-healing and watch will not lose any data, but it does happen in our environment.

I think this is really necessary, and if you don't think it's appropriate, we can use a less expensive way than sync requests, which is to implement it using etcd client's internal connection state monitor, client.getActiveConnection(), which will not initiate the request, but still get the connection state.
**IMHO, for the uncertain behavior in the etcd client (it is indeed a black box), where all the grpc stream details are hidden and the documentation does not match the actual behavior, an active health checking mechanism independent of the internal mechanism is a must.**

bzp2010: @nic-chen 

The connection can be automatically restored, but the behavior of watch is unobservable, and sometimes watch is incapacitated, but it still does not actively quit. There is no API to actively know which watchers are currently available and what is the state of those watchers.
This is a complete black box for its applicators, the entire documentation of etcd mentions that its connections are self-healing and watch will not lose any data, but it does happen in our environment.

I think this is really necessary, and if you don't think it's appropriate, we can use a less expensive way than sync requests, which is to implement it using etcd client's internal connection state monitor, client.getActiveConnection(), which will not initiate the request, but still get the connection state.
**IMHO, for the uncertain behavior in the etcd client (it is indeed a black box), where all the grpc stream details are hidden and the documentation does not match the actual behavior, an active health checking mechanism independent of the internal mechanism is a must.**

api/cmd/root.go
bzp2010: ok

bzp2010: @nic-chen fixed

bzp2010:
ping @nic-chen @starsz again

api/cmd/root.go
```diff
@@ -81,3 +89,59 @@ func manageAPI() error {
...
+func etcdConnectionChecker() context.CancelFunc {
+	ctx, cancel := context.WithCancel(context.TODO())
+	unavailableTimes := 0
+
+	go func() {
+		etcdClient := storage.GenEtcdStorage().GetClient()
+		for {
+			select {
+			case <-time.Tick(10 * time.Second):
+				sCtx, sCancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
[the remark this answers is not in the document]
bzp2010: This means applying a separate timeout for each request instead of using that global one ctx, which is used to turn off the timer.

api/test/shell/cli_test.sh
```diff
@@ -478,7 +477,39 @@ stop_dashboard() {
...
+  sleep 30
+
+  [ "$(grep -c "etcd connection loss detected" ${LOG_FILE})" -ge '1' ]
+
+  run docker start etcd
+
+  sleep 20
+
+  [ "$(grep -c "etcd connection recovered" ${LOG_FILE})" -ge '1' ]
+  [ "$(grep -c "etcd store reinitializing" ${LOG_FILE})" -ge '1' ]
```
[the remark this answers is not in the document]
bzp2010: This is not possible in principle, because each connection exception triggers a re-initialization of the entire abstract storage layer, which involves reloading the "list" in full, and creating a new watcher. As soon as the full load succeeds, the existing data in the cache will be overwritten by the new data, and what is not there will be added.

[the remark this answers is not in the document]
bzp2010: If the load fails, an error will be output to the console and log.

api/cmd/root.go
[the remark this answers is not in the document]
starsz: What I mean is that you can use ctx as the parent of sCtx rather than context.TODO().

nic-chen (approved):
I checked out your branch and tried to reproduce it, but failed.
I think we could merge it first. 
Before that, it is recommended to run it with APISIX for a long time to avoid causing other hidden issues. @bzp2010

bzp2010:
@nic-chen Thanks to your support, I have added a new use case to the CLI test to ensure that data changes during a connection outage will be resynchronized after the data is reinitialized. We can merge and continue to see the impact of this fix.

starsz (approved):

PR #2777, opened 2023-03-18
fix: locked input to a min of "0" for Retries

### LINKED ISSUE #2775: Negative value option reports an error.
### Issue description

Since you cannot set a negative value, why do you have such a scroll wheel option? ? ? Ha ha
image

### Expected behavior

1

### How to Reproduce

1

### Screenshots

_No response_

### Environment

`docker-compose`   deploy the latest version.

### Additional context

_No response_

Neilblaze:

- [x] Bugfix
- [x] Improve performance ?

In the Apisix dashboard, in the inputfield of `Retries` under [Services](https://apisix-dashboard.apiseven.com/service/create), we cannot set a negative value as of currently it doesn't supports it. Hence the up/down arrows have no function if the value is in the negative range.

RE: Primarily we can apply a `visibility: hidden` for `.ant-input-number-handler-wrap`, but I think the better way is to not remove those ` up/down ` arrows and instead, we can lock the input to a *minimum value of* **0**.

**Related issues**

resolves #2775

**Checklist:**

cc: @Baoyuantop, @freemankevin

### DIFF
--- a/web/src/components/Upstream/components/Retries.tsx
+++ b/web/src/components/Upstream/components/Retries.tsx
@@ -30,7 +30,7 @@ const Component: React.FC<Props> = ({ readonly }) => {
       tooltip={formatMessage({ id: 'component.upstream.fields.retries.tooltip' })}
       name="retries"
     >
-      <InputNumber disabled={readonly} />
+      <InputNumber disabled={readonly} min={0} />
     </Form.Item>
   );
 };


Baoyuantop (commented):
We need an E2E case to prove that this change is effective.

Neilblaze:
@Baoyuantop Will this work or should I have to refactor it a little bit?

```js
it('should not allow negative values for retries', () => {
  cy.visit('/routes/list');
  cy.get('[name="Retries"]').type('-1').blur();
  cy.contains('Retry times should not be less than 0');
});
```

Baoyuantop:
You can just commit your code directly.

juzhiyuan:
Thank you for your PR. We will create the new dashboard to resolve this issue. Refer to https://github.com/apache/apisix-dashboard/issues/2981

PR #3020, opened 2025-05-07
feat: plugin_configs pages

SkyeYoung:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches


Baoyuantop (approved):

bzp2010 (approved):
Probably later we could add a prominent hint to inform the user that service should be used in preference to plugin config unless he knows for sure that it is necessary to use it, as a superset it would cover most use cases.

bzp2010:
CI seems broken, we should also remove the CodeQL for golang.

SkyeYoung:
> Probably later we could add a prominent hint to inform the user that service should be used in preference to plugin config unless he knows for sure that it is necessary to use it, as a superset it would cover most use cases.

I think this should be written into the document...

bzp2010:
I've seen too many people who don't read the docs, no one cares what you say in the docs if you don't "get a gun aimed at them".

PR #3028, opened 2025-05-09
docs: update README and related docs

SkyeYoung:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches


docs/en/latest/config.json
bzp2010, line 1: Documentation should be deleted and we will no longer provide a separate documentation build for this project, but should be added to the APISIX project if needed.

Since we will no longer be doing releases on the current project, the mechanism in apisix-website that extracts documents based on a specific tag format will no longer work on this project.

bzp2010: If you need to add some documentation that is not needed in APISIX but only in this repository, you can drop the existing i18n structure and use the traditional DEVELOP.md or DEVELOPMENT.md. A multi-language version is also available, we can discuss it later.

README.md
```diff
@@ -1,54 +1,30 @@
...
+[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/apache/apisix-dashboard/blob/master/LICENSE)
+[![Slack](https://badgen.net/badge/Slack/Join%20Apache%20APISIX?icon=slack)](https://apisix.apache.org/slack)
+
+<p align="center">
+  <a href="https://apisix.apache.org/">Website</a> •
+  <a href="https://github.com/apache/apisix/tree/master/docs">Docs</a> •
+  <a href="https://twitter.com/apacheapisix">Twitter</a>
+</p>
+
+- The master version should be used with Apache APISIX master version.
```
bzp2010, line 12:
```suggestion
- The master version should be used with Apache APISIX master version.
- The project will not be released independently but will use a fixed git tag for each APISIX release.
```

docs/en/latest/config.json
SkyeYoung: In fact, the latest documentation will be deployed even if it is not published. 

To disable rendering, we need to do some processing in the APISIX website.

SkyeYoung: Removed the latest folder and docusaurus related configuration.

bzp2010 (approved):
The `docs/en/` prefix needs to be retained?

SkyeYoung:
> The `docs/en/` prefix needs to be retained?

Removed "en", but it's best to keep "docs" to reduce the number of files in the project's root directory.

PR #3100, opened 2025-06-04
feat: support rendering nested routes in service

SkyeYoung:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Allow rendering nested routes under the service detail page.

After clicking the tabs, the corresponding routes will be navigated to, and the state of the tabs themselves will also be consistent with the routes.

CleanShot 2025-06-04 at 17 20 35@2x

CleanShot 2025-06-04 at 17 21 02@2x

### DIFF
--- a/src/components/form-slice/FormPartCredential.tsx
+++ b/src/components/form-slice/FormPartCredential.tsx
@@ -19,16 +19,11 @@ import { useTranslation } from 'react-i18next';
 import { FormItemPlugins } from './FormItemPlugins';
 import { FormPartBasic } from './FormPartBasic';
 import { FormSection } from './FormSection';
-import {
-  FormSectionGeneral,
-  type FormSectionGeneralProps,
-} from './FormSectionGeneral';
 
-export const FormPartCredential = (props: FormSectionGeneralProps) => {
+export const FormPartCredential = () => {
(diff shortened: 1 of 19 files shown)


src/routes/services/detail.$id/index.tsx
SkyeYoung, line 1: This file is all copied from `src/routes/services/detail.$id.tsx`, no need to be reviewed.

src/routeTree.gen.ts
SkyeYoung, line 1: This file is auto-generated, no need to be reviewed.

src/routes/services/detail.$id/index.tsx
bzp2010: If it doesn't need to be modified in any way, why copy it instead of referencing it.
Is a higher level of abstraction necessary? 🤔

SkyeYoung: > If it doesn't need to be modified in any way, why copy it instead of referencing it. Is a higher level of abstraction necessary? 🤔

The original file needs to be used as a layout. This is the disadvantage of this structure.

Check https://tanstack.com/router/latest/docs/framework/react/routing/file-based-routing#directory-routes if u like

PR #3109, opened 2025-06-09
feat(workflow/e2e): allow `workflow_dispatch` trigger tests

SkyeYoung:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

This is part of the https://github.com/apache/apisix/issues/12234

According to [Events that trigger workflows](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch) , this will allow us to trigger the test manually.

**Why**

This is very suitable for manually running a test to ensure that the dashboard can adapt to the latest APISIX, either on the eve of APISIX's release or after the `APISIX:dev` image is updated.

PR #3110, opened 2025-06-09
feat(workflow/e2e): allow `repository_dispatch` trigger tests

SkyeYoung:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

This is part of the https://github.com/apache/apisix/issues/12234.

Depend on: #3109

According to [Events that trigger workflows](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#repository_dispatch) , this will allow us to trigger the test in other repositories.

**Why**

I see in https://github.com/apache/echarts/blob/master/.github/workflows/nightly-next.yml that `repository_dispatch` seems to be available in the Apache repositories.

But I haven't found any information about this feature on the ASF-related websites yet.

Because "This event will only trigger a workflow run if the workflow file exists on the default branch.", I need to merge this PR into the master for testing in the APISIX repository.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -13,6 +13,7 @@ on:
     # This should be later than the update time of `apache/apisix:dev`
     # Ref: https://github.com/apache/apisix-docker/blob/master/.github/workflows/apisix_dev_push_docker_hub.yaml#L7C15-L7C16
     - cron: '0 2 * * *'
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


juzhiyuan (approved):
Let's have a try

bzp2010 (approved):

LiteSun (approved):

PR #3111, opened 2025-06-10
feat(services): filter routes/stream_routes by `service_id`

SkyeYoung:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Depend on https://github.com/apache/apisix/pull/12291

According to https://github.com/apache/apisix/pull/12291, we can now filter out all routes/stream_routes that are not for the current service in the service's routes/stream_routes.

This can reduce user confusion.

However, those containing `service_id` in the routes/stream_routes page list cannot be filtered out for now, because the API does not yet support providing multiple `service_id`s or other methods.

**Screenshots**

create one route `test` in services, create one route `test_without_id` with upstream in routes.

image

image

### DIFF
--- a/src/apis/hooks.ts
+++ b/src/apis/hooks.ts
@@ -82,14 +82,16 @@ export const genUseList = <
   routeKey: T,
   listQueryOptions: ReturnType<typeof genListQueryOptions<P, R>>
 ) => {
-  return (replaceKey?: U) => {
+  return (replaceKey?: U, defaultParams?: Partial<P>) => {
     const key = replaceKey || routeKey;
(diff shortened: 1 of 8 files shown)


juzhiyuan (approved):
I understand why it's written this way (it could be optional), but the name `WithServiceIdFilter` makes me think it must contain the `ServiceId`. No better suggestions.

SkyeYoung:
> I understand why it's written this way (it could be optional), but the name `WithServiceIdFilter` makes me think it must contain the `ServiceId`. No better suggestions.

Let's keep pushing forward, not being blocked by a better type name

LiteSun (approved):

PR #3380, opened 2026-04-14
fix(header): guard missing translation progress in language menu

liangguifeng:
Please answer these questions before submitting a pull request, **or your PR will get closed**.

**Why submit this pull request?**

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

**What changes will this PR take into?**

This PR fixes a runtime error in `src/components/Header/LanguageMenu.tsx` when switching languages.

Previously, `TranslationProgress` accessed `i18nProgress[lang].percent` directly. When the translation progress entry for a language was missing, this caused a `TypeError` (`Cannot read properties of undefined
(reading 'percent')`) and broke the page.

This change adds a safe guard with optional chaining when reading the translation progress value, so the component gracefully renders nothing if the progress data for the selected language is unavailable.

**Related issues**

fix/resolve #3300

**Checklist:**

- [x] Did you explain what problem does this PR solve? Or what new features have been added?
- [ ] Have you added corresponding test cases?
- [ ] Have you modified the corresponding document?
- [x] Is this PR backward compatible? If it is not backward compatible, please discuss on the mailing list first

Fixes apache/apisix#13497

PR #3382, opened 2026-04-14
fix(header): fix crash on language switch by adding missing 'tr' lang and guard

### LINKED ISSUE #3300: 修复切换语言报错
### Issue description

When switching languages in the dashboard, clicking on any language other than Chinese (zh) causes a runtime error: "Cannot read properties of undefined (reading 'percent')"

### Expected behavior

Users should be able to switch between all available languages (en, de, zh, es, tr) without any errors.

### How to Reproduce

Image

1.Open the APISIX Dashboard
2.Click on the language icon in the header
3.Select any language other than Chinese (e.g., English, Deutsch, Español, Türkçe)
4.Observe the error in the console: "Cannot read properties of undefined (reading 'percent')"

### Screenshots

N/A

### Environment

apisix version: N/A
OS: Windows (win32)
OpenResty / Nginx version: N/A
etcd version: N/A
apisix-dashboard version: Latest (master branch)
Browser version: Any modern browser

### Additional context

Root Cause: The issue occurs in two places:

vite.config.ts: The i18nProgress plugin was configured with langs: ['zh'], but the language menu supports 5 languages (en, de, zh, es, tr). When accessing progress data for non-Chinese languages, i18nProgress[lang] is undefined.

src/components/Header/LanguageMenu.tsx: The TranslationProgress component directly accesses i18nProgress[lang].percent without checking if the data exists.

Baoyuantop:
## Problem

When switching languages in the dashboard, selecting Turkish (tr) causes a runtime error: `TypeError: Cannot read properties of undefined (reading 'percent')`.

## Root Cause

`vite.config.ts` configures the i18nProgress plugin with `langs: ['en', 'es', 'de', 'zh']`, but `LangMap` in `LanguageMenu.tsx` defines 5 languages including `'tr'` (Turkish). When the component accesses `i18nProgress['tr']`, it returns `undefined`, causing the crash when reading `.percent`.

## Changes

1. **Root cause fix**: Add `'tr'` to the `langs` array in `vite.config.ts` so the i18nProgress plugin generates progress data for all supported languages.
2. **Defensive guard**: Add optional chaining (`?.`) in `LanguageMenu.tsx` to prevent crashes if a language is missing from the progress data in the future.

## Related Issues

Fixes #3300

Supersedes #3380 (which only addressed the defensive guard without fixing the root cause)

## Checklist

### DIFF
--- a/src/components/Header/LanguageMenu.tsx
+++ b/src/components/Header/LanguageMenu.tsx
@@ -30,7 +30,7 @@ const LangMap: Record<keyof Resources, string> = {
 };
 
 const TranslationProgress = ({ lang }: { lang: string }) => {
-  const percent = i18nProgress[lang].percent;
+  const percent = i18nProgress[lang as keyof typeof i18nProgress]?.percent;
   if (typeof percent === 'number' && percent < 100) {
     return (
       <span
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -91,7 +91,7 @@ export default defineConfig({
       semicolons: false,
     }),
     i18nProgress({
-      langs: ['en', 'es', 'de', 'zh'],
+      langs: ['en', 'es', 'de', 'zh', 'tr'],
       baseLang: 'en',
       getTranslationDir: (lang) => `./src/locales/${lang}`,
     }),


Baoyuantop:
Closing in favor of pushing the fix directly to #3380.

PR #3396, opened 2026-05-25
fix(consumer): makes the username field read-only on the consumer edit page

okaybase:
username is the unique identifier of the Consumer. When editing the Consumer, username should not be editable. Currently, it can be edited and saved to create a new Consumer.

This PR makes the username field read-only on the consumer edit page.

**Checklist:**

- [x] I have added tests corresponding to this change


Baoyuantop (commented):
Please add coverage for cases: the username input is read-only in Consumer detail edit mode page.

okaybase:
> Please add coverage for cases: the username input is read-only in Consumer detail edit mode page.

tests added

Baoyuantop (approved):

LiteSun (approved):

Issue #3417, opened 2026-07-10
Tracking: verified findings from a full frontend review (master @ 9979b31)


moonming:
This is the tracking issue for a full review of the rewritten dashboard (`master @ 9979b31`), done with static analysis cross-checked against a live APISIX instance. Every item below was verified by reading the code at the cited location and/or reproducing at runtime — items are grouped by theme, each with its pointer. Dedicated issues/PRs are linked where they exist; unlinked items are up for grabs (several are `good first issue` material).

## Already tracked separately

- [ ] #3414 — **CRITICAL** no-op Edit→Save deletes inline upstream health checks (silent data loss; runtime-reproduced)
- [ ] #3415 — **CRITICAL** 401 fabricated into a successful `{data:{}}` response (false success toasts, crashes)
- [ ] #3416 — secrets/private keys/tokens plaintext in view mode
- [ ] #3412 — Plugin Metadata page false "Key not found" toasts (root cause identified in the issue: skip list says `'500'`, Admin API returns `404`)
- [ ] #3296 — `zOneOf` inverted xor (dead code today; fix or remove)
- [ ] #3205 — no search on list pages. Verified scope: **all 12** list pages set `search={false}`; worse, 5 pages (consumers, services, plugin_configs, consumer_groups, credentials) declare `sorter: true` with no handler — the sort arrows render but do nothing. The plumbing already exists end-to-end (`pageSearchSchema` reserves `name`/`label`; the request interceptor already serializes Admin API `filter`) — only the input box is missing.
- [ ] #3322 — dark mode. Cost note: nothing in `src/` is theme-aware today; the expensive part is the antd+Mantine dual-theming — nearly free after UI-library consolidation (below).

## Error handling / resilience

- [ ] Network-level failure (Admin API unreachable/timeout) of any mutation gives **zero feedback**: the interceptor only handles `err.response` errors (`src/config/req.ts`), the shared `QueryClient` has no `MutationCache.onError` (`src/config/global.ts`), and 25 of 26 `useMutation` sites have no local `onError`. Fix in one place: `new QueryClient({ mutationCache: new MutationCache({ onError }) })` toasting on `!err.response`.
- [ ] No `defaultErrorComponent` on the router: 11 of 12 resources render loader/render failures as the dev-styled red stack dump with no retry (only stream_routes registers a custom one — and that one dereferences `err.response?.data.error_msg` without guarding `data`, the exact #3370 shape).
[...]

PR #3418, opened 2026-07-10
fix: reject 401 responses instead of resolving fabricated empty data

### LINKED ISSUE #3415: bug: axios interceptor converts 401 into a fake successful response, causing false success toasts and crashes
### Issue description

The global axios response interceptor converts every **401** into a *successful* response carrying fabricated empty data:

```ts
// src/config/req.ts
if (res.status === HttpStatusCode.Unauthorized) {
  getDefaultStore().set(isSettingsOpenAtom, true);
  return Promise.resolve({ data: {} });   // ← auth failure becomes success
}
```

Every request helper then runs `.then((v) => v.data)`, so all downstream callers receive `{}` as a **success**. Depending on which page you're on when the admin key expires or is wrong, this produces four different (and contradictory) behaviors:

1. **False success on mutations** — a 401'd `DELETE` still shows the green "delete success" toast (`DeleteResourceBtn` chains `onSuccess` off the resolved promise) and invalidates queries; a 401'd `PUT` shows "Edit … Successfully". Nothing was actually deleted/saved.
2. **Crash after create** — the add pages' `onSuccess` handlers read `data.data.value.id` for navigation (e.g. `src/routes/upstreams/add.tsx`); with the fabricated `{}` this throws a TypeError.
3. **Crash on detail pages** — suspense queries resolve with `{}`, then e.g. `routes/detail.$id.tsx` dereferences `routeData.value.upstream` → TypeError → error boundary.
4. **Empty-cluster illusion on list pages** — the list renders as if the cluster had zero routes/services, which is indistinguishable from real data.

### Expected behavior

A 401 should open the settings modal (as it does today) **and reject the promise**, so callers take their normal error path: no success toasts, no fabricated data flowing into `onSuccess`/render.

### How to Reproduce

1. Log into the dashboard with a valid admin key; open any resource list.
2. Change the admin key on the APISIX side (or corrupt the stored key in localStorage).
3. Delete any resource from the list → observe the green success toast while the resource still exists; or open a detail page → observe the crash/error boundary.

### Suggested fix

```ts
if (res.status === HttpStatusCode.Unauthorized) {
  getDefaultStore().set(isSettingsOpenAtom, true);
}
return Promise.reject(err);
```

One-line change; the settings modal behavior is preserved. I'm preparing a PR with an e2e asserting a 401'd mutation shows no success toast and no crash.

### Environment

- apisix-dashboard: master (`9979b31`)

moonming:
## What

Fixes #3415.

The axios response interceptor converted every **401** into a *resolved* response carrying fabricated `{ data: {} }`. Every caller treated that as success:

- a 401'd DELETE showed the green **delete-success toast** (and invalidated queries) while the resource still existed
- add pages' `onSuccess` handlers read `data.data.value.id` from the fabricated body → TypeError
- detail pages' suspense queries resolved with `{}` → crash on `data.value.*`
- list pages rendered an empty-cluster illusion

## How

- `src/config/req.ts`: on 401, still open the settings modal, but **reject** so callers take their normal error path (one-line behavioral core of this PR)
- `src/routes/__root.tsx`: add a root `errorComponent` that keeps `<SettingsModal/>` mounted with a localized message + Retry button — on a fresh install every request 401s until the admin key is entered, and the modal is the only way in (previously the fabricated success masked this; with an honest reject the default TanStack error dump would have replaced the modal)
- `src/config/global.ts`: queries don't retry on 401 (retries can't succeed until the user fixes the key; failing fast surfaces the modal immediately)
- `src/components/page/DeleteResourceBtn.tsx`: swallow the rejection after the interceptor's toast so a failed delete (401/5xx/network) no longer surfaces as an unhandled promise rejection
- new locale keys `error.title` / `error.retry` in all 5 locales

## Verification

- new regression e2e `e2e/tests/regression/auth.401-no-false-success.spec.ts`: forces a 401 on DELETE → asserts settings modal opens, **no success toast**, row survives
- manually verified the fresh-install flow in a real APISIX environment: no key → localized error + Retry + settings modal; entering the key and clicking Retry recovers the page
- `pnpm lint` clean, `tsc -b` clean, e2e: new spec + `auth.spec.ts` + `form.mutation-failure-error.spec.ts` + `services.empty.spec.ts` all pass locally against a real APISIX

### DIFF
--- a/e2e/tests/regression/auth.401-no-false-success.spec.ts
+++ b/e2e/tests/regression/auth.401-no-false-success.spec.ts
@@ -23,7 +23,7 @@
 // AND reject, so callers take their normal error path.
 
 import { routesPom } from '@e2e/pom/routes';
-import { randomId } from '@e2e/utils/common';
+import { getAPISIXConf, randomId } from '@e2e/utils/common';
 import { e2eReq } from '@e2e/utils/req';
 import { test } from '@e2e/utils/test';
 import { expect } from '@playwright/test';
@@ -44,6 +44,7 @@ test('a 401 response must not produce a success toast', async ({ page }) => {
   await putRouteReq(e2eReq, {
     name,
     uri: `/regression/401/${name}`,
+    methods: ['GET'],
     upstream: {
       type: 'roundrobin',
       nodes: { 'reg-401.local:80': 1 },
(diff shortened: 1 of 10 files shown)


moonming:
Post-push audit findings addressed in 61fa5b03:

- **M1 (Retry no-op for suspense-thrown detail-page errors)** — fixed with the documented TanStack Router+Query pattern: `useQueryErrorResetBoundary().reset()` in the root error component before Retry re-runs the tree.
- **M2 (no automated coverage of the Retry recovery flow)** — added a `fresh install recovery` e2e: empty storage state → error screen + settings modal → enter key → **Retry** → recovered routes list, with no `page.reload()`.
- **M3 (blanket catch swallows non-axios bugs)** — the catch now rethrows anything that isn't an axios error, so failures thrown by `onSuccess`/`invalidateQueries` surface again.
- **L1 (ordering window in the negative toast assertion)** — the spec now anchors on the interceptor's red `failed to check token` toast (emitted strictly before any success-chain toast in both worlds) before asserting zero success toasts.
- **L2 (unhandled rejections from add/edit `mutateAsync`)** — pre-existing and out of scope; covered by the `MutationCache.onError` follow-up tracked in #3417.

LiteSun (approved):

guoqqqi (approved):

PR #3419, opened 2026-07-10
fix(plugin-metadata): stop toasting 404 probes as errors

### LINKED ISSUE #3412: Plugin Metadata page shows "Key not found" error toasts for every plugin without configured metadata
### Issue description

Opening the **Plugin Metadata** page in the dashboard always shows "Key not found"
error toasts, even on a healthy APISIX instance.

The page enumerates every plugin that supports plugin metadata and issues an
individual `GET /apisix/admin/plugin_metadata/{plugin}` request for each one. 
The Admin API returns `404 {"message":"Key not found"}` for any plugin whose
metadata has never been configured — which is the documented/expected Admin API 
behavior. The dashboard, however, surfaces each of these 404s as an error toast 
instead of treating them as "not configured (empty)". 

Since virtually no deployment configures metadata for all ~24 metadata-capable
plugins, this error appears on essentially every installation, which makes users
think something is broken.

### Expected behavior

A 404 from `GET /plugin_metadata/{plugin}` means "no metadata configured for
this plugin" and should be rendered as an empty/unset state, not as an error. 
Error toasts should be reserved for real failures (5xx, network, auth). 

Alternatively, the page could rely on the list endpoint 
(`GET /apisix/admin/plugin_metadata`, which returns `{"list":[],"total":0}`)
instead of probing each plugin individually.

### Actual behavior
Every unconfigured plugin produces a "Key not found" error toast on page load.
Configured plugins (e.g. http-logger above) display correctly, so the page is 
otherwise functional. 

### How to Reproduce

1. Run APISIX 3.17.0 with the embedded dashboard, with no plugin metadata 
 configured (fresh install is enough).
2. Open the dashboard → **Plugin Metadata** menu.
3. "Key not found" error toast(s) appear. 

Admin API access log during a single page load (one GET per plugin):

"GET /apisix/admin/plugin_metadata HTTP/1.1" 200<- list endpoint is fine
"GET /apisix/admin/plugin_metadata/limit-count HTTP/1.1" 404
"GET /apisix/admin/plugin_metadata/kafka-logger HTTP/1.1" 404 
"GET /apisix/admin/plugin_metadata/datadog HTTP/1.1" 404
"GET /apisix/admin/plugin_metadata/cors HTTP/1.1" 404 
... (~24 plugins total, one 404 per unconfigured plugin)
"GET /apisix/admin/plugin_metadata/http-logger HTTP/1.1" 200 <- the one we configured renders fine

### Screenshots

Image

### Environment

- APISIX: 3.17.0 (apache helm chart apisix-2.15.0) with embedded dashboard
- Kubernetes: RKE2, deployment via Helm 
- Admin API version: v3

### Additional context

_No response_

moonming:
## What

Fixes #3412.

Opening the Plugin Metadata page on any deployment that hasn't configured metadata for all ~24 metadata-capable plugins produced a red `Key not found` toast per unconfigured plugin. The page probes `GET /plugin_metadata/{name}` per plugin; the Admin API's documented answer for never-configured metadata is **404**, but the per-request interceptor skip list said `'500'` — the wrong status code.

## How

- `src/components/page-slice/plugin_metadata/hooks.ts`: skip list `['500']` → `['404']`. Unconfigured plugins stay quiet; real failures (5xx, network) now surface via the interceptor toast instead of being silently swallowed.
- Same hook, second defect: any failed metadata GET used to be classified as "not configured" and offered as an empty editable `{}` — for a plugin whose fetch failed with a 500/network error, saving that would overwrite existing metadata. Non-404 failures are now excluded from the plugin map until the fetch succeeds.

## Verification

- new regression e2e `e2e/tests/regression/plugin-metadata.no-false-404-toasts.spec.ts`: waits for the probe storm to settle, asserts **zero** `role=alert` toasts, and confirms the select-plugins drawer still offers plugins (404 = unconfigured still works)
- existing `plugin_metadata.list.spec.ts` + `plugin_metadata.crud-required-fields.spec.ts` pass locally against a real APISIX

### DIFF
--- a/e2e/tests/regression/plugin-metadata.no-false-404-toasts.spec.ts
+++ b/e2e/tests/regression/plugin-metadata.no-false-404-toasts.spec.ts
@@ -69,7 +69,7 @@ test('page load with unconfigured plugins shows no error toasts', async ({
         lastSeen = probes;
         return settled;
       },
-      { timeout: 15000, intervals: [500] }
+      { timeout: 15000, intervals: [1000] }
     )
     .toBe(true);
 
@@ -82,4 +82,37 @@ test('page load with unconfigured plugins shows no error toasts', async ({
   const drawer = page.getByRole('dialog');
   await expect(drawer.getByRole('button', { name: 'Add' }).first())
     .toBeVisible();
+});
+
+test('a plugin whose metadata probe fails is not offered as an empty config', async ({
+  page,
+}) => {
+  // Fault injection: a real Admin API cannot be made to 500 a single
+  // plugin's metadata GET deterministically. A failed (non-404) probe
+  // must NOT be presented as "unconfigured" — saving the offered empty
+  // config would overwrite the plugin's existing metadata.
+  await page.route('**/apisix/admin/plugin_metadata/http-logger', (route) =>
+    route.fulfill({
+      status: 500,
+      contentType: 'application/json',
+      body: JSON.stringify({ error_msg: 'injected metadata failure' }),
+    })
+  );
+
+  await pluginMetadataPom.toIndex(page);
+  await pluginMetadataPom.isIndexPage(page);
+
+  // The real failure must surface to the user ...
+  await expect(
+    page.getByRole('alert').filter({ hasText: 'injected metadata failure' })
+  ).toBeVisible({ timeout: 15000 });
+
+  // ... and the failed plugin must not be offered in the select drawer.
+  await page.getByRole('button', { name: 'Select Plugins' }).click();
+  const drawer = page.getByRole('dialog');
+  await expect(drawer.getByRole('button', { name: 'Add' }).first())
+    .toBeVisible();
+  await expect(
+    drawer.getByText('http-logger', { exact: true })
+  ).toHaveCount(0);
 });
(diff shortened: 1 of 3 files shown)


moonming:
Post-push audit findings addressed in ffb12f13:

- **H1 (failed plugin still offered via Select Plugins drawer)** — fixed: the hook now exposes `failedPluginNames`; the drawer offering subtracts it, and the `curPlugin` fallback only hands out an empty `{}` for plugins genuinely offered as unconfigured. Pinned by a new fault-injection e2e (500 on one probe → toast appears, plugin absent from drawer).
- **M2 (no recovery for transient failures)** — real (non-404) failures now retry up to 2 times; 404 probes never retry, so the probe volume for the normal case is unchanged.
- **M3 (revert coverage)** — the new second test covers the exclusion hunk; reverting either hunk now fails a spec.
- **M1 (network errors don't toast)** — pre-existing interceptor gap, intentionally out of scope here; tracked in #3417 ("Network-level failure … zero feedback") with a `MutationCache.onError`/interceptor follow-up.
- **L1 (toast-storm on per-plugin 500s)** — mitigated by the interceptor's dedup-by-message (`notifications.show({ id: message })`) plus the Mantine `limit={5}` cap; noted for the record.
- **L2 (settle-poll window)** — widened the quiet window to 1s.

LiteSun (approved):

guoqqqi (approved):

Issue #3376, opened 2026-04-10
bug: can not save "Discovery Args"  when create or update a Upstreams


poeticdog:
### Current Behavior

<img width="1251" height="661" alt="Image" src="https://github.com/user-attachments/assets/66ea179c-dc53-4ccc-acaa-8ae62bdacd71" />

### Steps to Reproduce

1. add or update a Upstreams
2. Find Upstream From->Service Discovery->Discovery Args, input a json value
3. click the button "Add"

### Environment

- APISIX version (run `apisix version`):
- Operating system (run `uname -a`):
- OpenResty / Nginx version (run `openresty -V` or `nginx -V`):
- etcd version, if relevant (run `curl http://127.0.0.1:9090/v1/server_info`):
- APISIX Dashboard version, if relevant:
- Plugin runner version, for issues related to plugin runners:
- LuaRocks version, for installation issues (run `luarocks --version`):

Baoyuantop:
The root cause is that the Zod schema is too restrictive.

src/types/schema/apisix/upstreams.ts:

```
const UpstreamDiscovery = z.object({
   discovery_type: z.string().optional(),
   service_name: z.string().optional(),
   discovery_args: z.object({}).optional(),  // ← this is the issue
 });
```

 z.object({}) defines an object in Zod that does not contain any known properties. Zod’s default behavior is to strip unknown properties; that is, when a user enters {“group_name”: “my-group”}, Zod parses it into {}, discarding all key-value pairs.

Welcome to create a PR to fix it.

PR #3420, opened 2026-07-10
fix: zOneOf rejected the valid exactly-one state (inverted xor)

### LINKED ISSUE #3296: zOneOf validation errors when exactly one field is provided
### Issue description

- The `zOneOf` helper incorrectly reports errors when exactly one of two fields is filled. It should only error when both are filled or both are empty.
- Location: [src/utils/zod.ts](src/utils/zod.ts#L23-L52), function `zOneOf(...)`.
- Root Cause: The code uses `xor(arg[key1] as boolean, arg[key2] as boolean)` and treats that as an error case. `xor` is `true` when exactly one input is truthy; the function should error only when “not exactly one” is filled (both or neither). Raw boolean coercion is also fragile for “filled” detection (e.g., empty strings vs. `undefined`).

### Expected behavior

- Validation passes when exactly one of the two keys is “filled” per schema rules.
- Validation fails when both keys are filled or when neither key is filled.

### Code-based reproduction
```ts
import { z } from 'zod';
import { zOneOf } from '@/utils/zod';

// Treat "filled" as non-empty string for this example
const schema = z
  .object({ a: z.string().optional(), b: z.string().optional() })
  .superRefine(zOneOf<{ a?: string; b?: string }, 'a', 'b'>('a', 'b'));

// Case 1: only a filled → should pass, currently fails
const res1 = schema.safeParse({ a: 'x', b: undefined });

// Case 2: only b filled → should pass, currently fails
const res2 = schema.safeParse({ a: undefined, b: 'y' });

// Case 3: both filled → should fail, may pass
const res3 = schema.safeParse({ a: 'x', b: 'y' });

// Case 4: neither filled → should fail, may pass
const res4 = schema.safeParse({ a: undefined, b: undefined });

// Expected:
// res1.success === true
// res2.success === true
// res3.success === false
// res4.success === false
```

### UI-based reproduction (if a form uses one-of semantics)
1. Open any form that requires exactly one of two fields (e.g., a mutually exclusive pair).
2. Fill only the first field and leave the second empty.
3. Submit the form.
4. Observe an error raised on both fields, contradicting the intended “exactly one” rule.

### Screenshots

_No response_

### Environment

- apisix version (cmd: `apisix version`): N/A (services not running locally)
- OS: Microsoft Windows 11 Home Single Language, 10.0.26200 Build 26200
- OpenResty / Nginx version (cmd: `nginx -V` or `openresty -V`): N/A (services not running locally)
- etcd version, if have (cmd: run `etcd --version`): N/A (services not running locally)
- apisix-dashboard version, if have: master branch, commit bbe05ad9 (package.json version 0.0.0)
- Browser version, if have: Google Chrome 144.0.7559.110

### Additional context

I can open a PR to fix this and add unit tests.

moonming:
## What

Fixes #3296.

`zOneOf` used `xor(a, b)` as the *error* condition — but `xor` is `true` when exactly one operand is truthy, i.e. the **valid** state. The helper rejected exactly-one and accepted both-set/none-set: backwards on all four quadrants. The `as boolean` coercion also made a cleared form input (`''`) behave inconsistently.

Note: `zOneOf` has **no call sites** on master today (the existing e2e `validation.zOneOf-single-field.spec.ts` exercises the form behavior, not this helper), so there is no user-facing behavior change — this defuses the trap before the next feature imports it.

## How

- pass when exactly one key is filled; `undefined`/`null`/`''` count as not filled
- one issue per key so both fields display the error (unchanged)
- unit tests over all four quadrants + the empty-string case (`src/utils/zod.test.ts`)
- wire `pnpm test` → `vitest run src`: the repo's only existing unit test (`src/types/schema/apisix/upstreams.test.ts`, guarding the #3376 discovery_args fix) had no runner script and never executed; now both files run in 400ms (scoped to `src` so vitest doesn't pick up the Playwright specs)

## Verification

- `pnpm test`: 2 files, 9 tests pass (5 new)

Follow-up (tracked in #3417): add `pnpm test` to a CI workflow so schema unit tests gate PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "lint": "eslint . --cache --max-warnings=0 --no-warn-ignored",
     "lint:fix": "pnpm lint --fix",
     "preview": "vite preview",
+    "test": "vitest run src",
     "prepare": "husky",
     "e2e": "playwright test"
   },
--- a/src/utils/zod.test.ts
+++ b/src/utils/zod.test.ts
@@ -64,4 +64,15 @@ describe('zOneOf', () => {
     expect(schema.safeParse({ a: '', b: 'y' }).success).toBe(true);
     expect(schema.safeParse({ a: '', b: '' }).success).toBe(false);
   });
+
+  it('treats 0 and false as filled', () => {
+    // 0 is a legitimate port/weight and false a deliberate toggle — a
+    // truthiness-based implementation would wrongly treat them as empty
+    const s = z
+      .object({ n: z.number().optional(), f: z.boolean().optional() })
+      .superRefine(zOneOf<{ n?: number; f?: boolean }, 'n', 'f'>('n', 'f'));
+    expect(s.safeParse({ n: 0 }).success).toBe(true);
+    expect(s.safeParse({ f: false }).success).toBe(true);
+    expect(s.safeParse({ n: 0, f: false }).success).toBe(false);
+  });
 });
--- a/src/utils/zod.test.ts
+++ b/src/utils/zod.test.ts
@@ -1,0 +1,67 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


moonming:
Post-push audit findings addressed in 0635321f:

- **M1 (0/false semantics untested)** — added a test asserting `0` and `false` count as filled; a truthiness-based re-implementation now fails the suite.
- **L1 (object values always count as filled)** — documented at the `isFilled` definition so a future object/string adopter isn't ambushed.
- **L2 (scoping only in the script argument)** — `test.include: ['src/**/*.test.{ts,tsx}']` added to `vite.config.ts`; a bare `vitest` no longer collects the Playwright specs.
- **L3 (type-level narrowing imprecision)** — acknowledged as pre-existing and left as-is per the audit's recommendation (zero call sites).

LiteSun (approved):

guoqqqi (approved):

PR #3422, opened 2026-07-10
fix(settings): debounce the admin-key query refresh

moonming:
## What

The Admin Key field in the settings modal fired `queryClient.invalidateQueries()` + `queryClient.refetchQueries()` on **every keystroke** (inside a bare `setTimeout`). Typing a 32-char admin key = 32 full invalidate+refetch storms of every mounted query, usually against a 401-ing backend. While reviewing this live I captured an 18-request burst of identical `GET /services` calls from one typing session.

## How

- keep the per-keystroke jotai atom update (controlled input; the axios request interceptor reads the atom)
- debounce the query refresh with `useDebouncedCallback` (@mantine/hooks, already a dependency) at 500ms — one refresh when the user pauses, same recovery behavior for paste and for the auth e2e flow

## Verification

- new regression e2e `e2e/tests/regression/settings.admin-key-debounced-refetch.spec.ts`: types the key character-by-character (30ms delay), waits for the request count to settle, asserts the Admin API request count stays far below one-refetch-per-keystroke
- `auth.spec.ts` passes unchanged (fill + reload flow unaffected)

Found during the review tracked in #3417.

### DIFF
--- a/e2e/tests/regression/settings.admin-key-debounced-refetch.spec.ts
+++ b/e2e/tests/regression/settings.admin-key-debounced-refetch.spec.ts
@@ -1,0 +1,64 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


moonming:
Post-push audit findings addressed in 947836a6:

- **HIGH (pending refresh cancelled on modal close)** — fixed with `{ delay: 500, flushOnUnmount: true }`; Mantine's numeric-arg form hard-sets `flushOnUnmount: false` and cancels on unmount, so paste-then-close would have left every query in its 401 error state.
- **MEDIUM (spec never verified the refresh fires)** — the spec now requires a successful (`res.ok()`) Admin API response after typing completes; 401-retry noise alone can no longer satisfy it.
- **MEDIUM (tight <8 bound)** — bound is now relative: `adminRequests < adminKey.length` (old behavior ≥ one refetch per keystroke; fixed behavior ~a handful), with real headroom against retry noise.
- **LOW (double-fetch)** — switched to `invalidateQueries({ refetchType: 'none' })` + a single `refetchQueries()`: one fetch per query, still covering observer-less error-state queries that wrong-key recovery depends on.

LiteSun (approved):

guoqqqi (approved):

PR #3423, opened 2026-07-10
fix: no-op edit-save deleted inline upstream health checks

### LINKED ISSUE #3414: bug: no-op Edit→Save on route/service detail silently deletes inline upstream health checks
### Issue description

Opening a route (or service / stream route) whose **inline upstream has health checks configured**, clicking **Edit**, and clicking **Save without changing anything** silently deletes `upstream.checks` from the resource. The UI shows a green "Edit Route Successfully" toast; the health checks are gone.

Reproduced on `master` (`9979b31`) against a real APISIX instance.

### Root cause

The detail pages' refetch-reset path uses the wrong producer for the health-check UI flags:

- The health-check section is driven by a form-root-level virtual flag: `FormSectionChecks.tsx` registers `__checksEnabled` **unprefixed** and reads it via `useWatch({ name: '__checksEnabled' })`.
- `src/routes/routes/detail.$id.tsx` resets the form with `produceToNestedUpstreamForm(routeData.value)`, but that producer (`src/components/form-slice/FormPartUpstream/util.ts:34-49`) writes the flags **nested** under `upstream.__checksEnabled` (it only sets root-level flags `if (d.checks)`, which routes/services never have at the top level).
- After the reset, the root-level watch reads `undefined` → the checks section renders as "disabled" → its inputs unmount → with `shouldUnregister: true` on the form, `upstream.checks` is omitted from the PUT body → the Admin API PUT replaces the resource without checks.

Affected pages (same reset pattern): `routes/detail.$id.tsx`, `services/detail.$id/index.tsx`, `stream_routes/detail.$id.tsx`.

### How to Reproduce

1. Create a route with an inline upstream carrying active health checks via the Admin API:
   ```bash
   curl -X PUT http://127.0.0.1:9180/apisix/admin/routes/checks-repro \
     -H "X-API-KEY: $KEY" -d '{
       "uri": "/checks-repro",
       "upstream": {
         "type": "roundrobin",
         "nodes": {"127.0.0.1:8000": 1},
         "checks": {"active": {"type": "http", "http_path": "/health",
           "healthy": {"interval": 2, "successes": 2},
           "unhealthy": {"interval": 2, "http_failures": 3}}}
       }}'
   ```
2. Open the route's detail page in the dashboard → **Edit** → **Save** (change nothing).
3. `GET /apisix/admin/routes/checks-repro` → `upstream.checks` is gone.

### Expected behavior

A save that changes nothing must be a no-op: the resource read back after save is identical to what was there before (health checks included).

### Why this is severe

This is silent data loss on the most routine action in the product — a user editing a route's description destroys their health-check configuration with zero feedback. It is also invisible to the current e2e suite: no round-trip test asserts field preservation across a no-op edit-save.

### Suggested fix

Make `produceToNestedUpstreamForm` also derive the root-level flags from the nested upstream:

```ts
if (d.upstream && typeof d.upstream === 'object' && !Array.isArray(d.upstream)) {
  d.upstream = produceToUpstreamForm(d.upstream, d.upstream);
  d.__checksEnabled = !!(d.upstream as Upstream).checks && isNotEmpty((d.upstream as Upstream).checks);
  d.__checksPassiveEnabled = !!(d.upstream as Upstream).checks?.passive && isNotEmpty(...);
}
```

plus a regression e2e: `route detail preserves upstream.checks after no-op edit-save`. I'm preparing a PR.

### Environment

- apisix-dashboard: master (`9979b31`), UI served by `pnpm dev` against APISIX master via the standard `/apisix/admin` proxy

moonming:
## What

Fixes #3414 (**silent data loss**).

Opening a route/service/stream-route whose **inline upstream has health checks**, clicking **Edit**, and clicking **Save without changing anything** silently deleted `upstream.checks`. The green "Edit … Successfully" toast made the loss invisible. Reproduced against a real APISIX before fixing.

## Root cause

The detail pages reset their form with `produceToNestedUpstreamForm`, which wrote the `__checksEnabled` UI flags **nested** under `upstream.` — but `FormSectionChecks` registers and reads those flags at the form **root**. After the reset (triggered by the mount-time background refetch), the root flag read `undefined` → the checks section rendered as "disabled" → its inputs unmounted → `shouldUnregister: true` dropped `upstream.checks` from the PUT body → the Admin API PUT replaced the resource without checks.

## How

- `produceToNestedUpstreamForm` now derives the root-level `__checksEnabled`/`__checksPassiveEnabled` from the nested upstream's checks (`src/components/form-slice/FormPartUpstream/util.ts`)
- `stream_routes/detail.$id.tsx` resets through the producer (it reset with the raw API value, so the flags were never computed there at all)

## Verification — including a RED check

- new regression e2e `e2e/tests/regression/form.noop-edit-preserves-checks.spec.ts` (route + service variants): seeds a resource with checks via the Admin API, **loads the detail page directly** (full page load — the bug only reproduces on a fresh mount), waits for the detail queries to settle (the exact reset that wiped the flags), performs a no-op Edit→Save, then asserts via the Admin API that `upstream.checks` survived
- **verified the spec fails on the unfixed code** (route variant catches the wipe; the PUT body captured during debugging shows the upstream stripped of `checks`) and passes with the fix — the timing subtlety is documented in the spec
- also re-verified manually in a live environment: unfixed → checks GONE after no-op save; fixed → PRESERVED

### DIFF
--- a/e2e/tests/regression/form.noop-edit-preserves-checks.spec.ts
+++ b/e2e/tests/regression/form.noop-edit-preserves-checks.spec.ts
@@ -103,6 +103,10 @@ test('no-op edit-save preserves route inline upstream health checks', async ({
   await uiGoto(page, '/routes/detail/$id', { id });
   await routesPom.isDetailPage(page);
   await waitForDetailQueriesToSettle(() => routeGets);
+  // deterministic discriminator: after the post-refetch reset, the form
+  // must still show health checks as enabled — the bug's first visible
+  // symptom was this switch reading off
+  await expect(page.getByTestId('checksEnabled')).toBeChecked();
 
   await page.getByRole('button', { name: 'Edit' }).click();
   await page.getByRole('button', { name: 'Save' }).click();
@@ -142,6 +146,8 @@ test('no-op edit-save preserves service inline upstream health checks', async ({
   await uiGoto(page, '/services/detail/$id', { id });
   await servicesPom.isDetailPage(page);
   await waitForDetailQueriesToSettle(() => serviceGets);
+  // deterministic discriminator (see the route variant above)
+  await expect(page.getByTestId('checksEnabled')).toBeChecked();
 
   await page.getByRole('button', { name: 'Edit' }).click();
   await page.getByRole('button', { name: 'Save' }).click();
--- a/e2e/tests/regression/form.noop-edit-preserves-checks.spec.ts
+++ b/e2e/tests/regression/form.noop-edit-preserves-checks.spec.ts
@@ -1,0 +1,159 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Regression for apache/apisix-dashboard#3414: opening a route/service whose
+// INLINE upstream has health checks, clicking Edit, and saving WITHOUT
+// changing anything silently deleted `upstream.checks`. Root cause: the
+// detail pages reset the form with `produceToNestedUpstreamForm`, which set
+// the `__checksEnabled` UI flag nested under `upstream.` while the checks
+// section reads it from the form ROOT — the section unmounted and
+// `shouldUnregister: true` dropped `checks` from the PUT body.
+//
+// A save that changes nothing must be a no-op.
+
+import { routesPom } from '@e2e/pom/routes';
+import { servicesPom } from '@e2e/pom/services';
+import { randomId } from '@e2e/utils/common';
+import { e2eReq } from '@e2e/utils/req';
+import { test } from '@e2e/utils/test';
+import { uiGoto } from '@e2e/utils/ui';
+import { expect } from '@playwright/test';
+
+import { deleteAllRoutes } from '@/apis/routes';
+import { deleteAllServices } from '@/apis/services';
+import type { APISIXType } from '@/types/schema/apisix';
+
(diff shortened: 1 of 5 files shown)


moonming:
Post-push audit came back with **no HIGH findings**; disposition of the rest (8dd04e85):

- **M3 (settle heuristic can race on slow runners)** — addressed: after the GET-settle wait, both variants now assert `getByTestId('checksEnabled')` is checked before clicking Edit. The bug's first visible symptom was this switch reading off, so the discriminator is now deterministic rather than timing-based.
- **M1 (no stream_routes e2e variant)** and **M2 (passive-only regression not covered)** — deliberately left as follow-ups per maintainer bandwidth; both are cheap additions inside `form.noop-edit-preserves-checks.spec.ts` and are listed in the test-gap section of #3417 for anyone picking it up. The stream_routes code path itself is exercised manually and shares the fixed producer.
- **L1 (top-level `if (d.checks)` precedence)** / **L2 (flag-derivation duplicated in two producers)** — valid cleanups, no current caller is affected; noted here for a future refactor.
- **L3 (`rmDoubleUnderscoreKeys` crashes on `null`)** — pre-existing, already tracked as a checkbox in #3417 ("Data-integrity" section).

moonming:
CI's e2e failure on `routes.crud-all-fields` turned out to be this PR **exposing a second, older bug** that the #3414 wipe had been masking — fixed in f413fc81:

**What CI caught:** after this PR mounts the checks section correctly, editing a route whose checks only configure the *unhealthy* side (exactly what the dashboard's own create flow stores — no `active.healthy`, no `type`) made Save silently do nothing: the hand-written zod schema declared `checks.active.healthy`/`unhealthy` as **required**, while the Admin API treats both as optional. The resolver error sat on a path with no rendered input, so there was no PUT, no toast, and no visible error. On master this never surfaced because the wiped checks section kept those fields unregistered — the data-loss bug was hiding the validation bug.

**Fixes:**
1. `upstreams.ts`: `active.healthy`/`active.unhealthy` → `.optional()`, matching the Admin API (the passive side already was).
2. `services`/`stream_routes` detail pages: `defaultValues` now run through `produceToNestedUpstreamForm` too — when the root flags only appear at reset time, the late-mounting passive section missed the reset values and the passive block was dropped from the PUT (root+service variants of the regression spec both discriminate this now).
3. The regression seed is now deliberately sparse (mirroring UI-stored shape) and asserts the **passive** block survives — which also covers the audit's M2.

Verified locally: the previously failing `routes.crud-all-fields.spec.ts` now passes, plus upstreams/services crud suites and the strengthened regression spec. (`stream_routes` specs can't run against my local backend — stream mode disabled — CI covers those.)

guoqqqi (approved):

Yilialinn (approved):

PR #3435, opened 2026-07-20
fix: array<string> fields rendered with wrong widgets; SSL edit wiped the mTLS client block

lxbme:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Part of #3417 (Data-integrity section), two widget/type-mismatch items plus a silent data-loss bug discovered while fixing the second one. Three commits, one root cause each:

**1. `checks.active.http_request_headers` rendered with the Labels widget (`d8157e7`)**

The field is `array<string>` (raw header lines, matching the Admin API) but was rendered with the Labels widget, which expects an object. Both directions were broken: stored header lines displayed as an empty field (the widget's array guard returns `[]`), and typed input produced an object that failed the zod array schema on submit — the field was unusable. Now a `FormItemTagsInput`, the same control the sibling `http_statuses` fields already use.

**2. `ssls.client.skip_mtls_uri_regex` rendered as a boolean Switch (`1c0d99c`)**

Same class: the field is `array<string>` (URI regexes) but was rendered as a Switch. A stored regex list displayed as a meaningless "on" toggle with the values invisible, and toggling wrote a boolean that failed the array schema. Also now a `FormItemTagsInput`.

**3. Edit-save silently deleted the SSL's mTLS client block (`997f608`)**

Found while writing the e2e for item 2: on master, opening an SSL that has a `client` block, clicking Edit and saving **silently deletes the whole `client` block** — the PUT succeeds, the success toast shows, and mTLS client verification is gone. (The existing ssls crud spec's "click Cancel instead of Save" workaround hints at this area's fragility.)

Root cause, probe-verified: the SSL detail page created its form **without creation-time `defaultValues`** and populated it only through a later `form.reset()`. Under `shouldUnregister: true`, react-hook-form's `useController` mount effect (v7.56.1) overwrites a field's `_defaultValues` entry from `_options.defaultValues` — i.e. with `undefined` — so every controlled field that mounts **after** that reset loses its re-seed source. The client section is exactly that (it mounts once the reset sets `__clientEnabled`), so the unmount/remount around toggling Edit dropped the values with nothing to restore from: `client.ca` measured 1333 chars in view mode and `undefined` immediately after clicking Edit, while root-level fields that registered before the reset (e.g. `cert`) survived.

Fix: pass `produceToSSLForm(sslData)` as creation-time `defaultValues` (the data is synchronously available via `useSuspenseQuery`) — the exact pattern the routes/services/stream_routes detail pages already use, and the same bug class as #3414. A survey of all detail pages confirmed ssls was the only remaining page combining `shouldUnregister` + reset-only population + a conditionally-mounted controlled section.

**Tests** (each red on the unfixed build at the intended assertion, green after):

- `upstreams.http-request-headers-field.spec.ts` — stored header lines visible, typed line round-trips to storage as an array member.
- `ssls.skip-mtls-uri-regex-field.spec.ts` — stored regexes visible as tags, typed regex round-trips; the private key must be re-entered before saving because the Admin API never returns it (pre-existing, same as the crud spec).
- `ssls.noop-edit-preserves-client.spec.ts` — a no-op Edit → Save preserves `client.ca`, `client.depth` and `client.skip_mtls_uri_regex` verbatim (red reproduced the silent deletion: save "succeeded" with `client` gone from storage).

Blast radius: full local e2e suite — **173 passed**; the only 2 failures are documented environment items unrelated to this change (a Monaco read-back race that reproduces on pristine master, and `stream_routes.show-disabled-error`, which cannot run outside the repo's own compose project). Unit tests, lint and build clean.

**Related issues**

**Checklist:**

- [ ] Have you modified the corresponding document? (no user-facing document covers these form fields)

### DIFF
--- a/src/components/form-slice/FormPartSSL/index.tsx
+++ b/src/components/form-slice/FormPartSSL/index.tsx
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { InputWrapper, Text } from '@mantine/core';
+import { Text } from '@mantine/core';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 
@@ -54,12 +54,13 @@ const FormSectionClient = () => {
             defaultValue={1}
             min={0}
           />
-          <InputWrapper label={t('form.ssls.client.skipMtlsUriRegex')}>
-            <FormItemSwitch
-              control={control}
-              name="client.skip_mtls_uri_regex"
-            />
-          </InputWrapper>
+          {/* array<string> of URI regexes — a TagsInput, not a Switch:
+              the boolean broke this field both ways (#3417) */}
+          <FormItemTagsInput
+            control={control}
+            name="client.skip_mtls_uri_regex"
+            label={t('form.ssls.client.skipMtlsUriRegex')}
+          />
         </>
       ) : (
         <Text c="gray.6" size="sm">
--- a/src/components/form-slice/FormPartSSL/index.tsx
+++ b/src/components/form-slice/FormPartSSL/index.tsx
@@ -55,11 +55,14 @@ const FormSectionClient = () => {
             min={0}
           />
           {/* array<string> of URI regexes — a TagsInput, not a Switch:
-              the boolean broke this field both ways (#3417) */}
+              the boolean broke this field both ways (#3417). splitChars={[]}
+              because a regex can contain a comma (e.g. the quantifier
+              `{1,3}`) and must stay one entry (#3435 review). */}
           <FormItemTagsInput
             control={control}
             name="client.skip_mtls_uri_regex"
             label={t('form.ssls.client.skipMtlsUriRegex')}
+            splitChars={[]}
           />
         </>
       ) : (
--- a/src/components/form-slice/FormPartUpstream/FormSectionChecks.tsx
+++ b/src/components/form-slice/FormPartUpstream/FormSectionChecks.tsx
@@ -18,7 +18,6 @@ import { Text } from '@mantine/core';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 
-import { FormItemLabels } from '@/components/form/Labels';
 import { FormItemNumberInput } from '@/components/form/NumberInput';
 import { FormItemSelect } from '@/components/form/Select';
 import { FormItemSwitch } from '@/components/form/Switch';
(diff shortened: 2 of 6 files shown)


src/components/form-slice/FormPartSSL/index.tsx
```diff
@@ -54,12 +54,13 @@ const FormSectionClient = () => {
...
           />
-          <InputWrapper label={t('form.ssls.client.skipMtlsUriRegex')}>
-            <FormItemSwitch
-              control={control}
-              name="client.skip_mtls_uri_regex"
-            />
-          </InputWrapper>
+          {/* array<string> of URI regexes — a TagsInput, not a Switch:
+              the boolean broke this field both ways (#3417) */}
+          <FormItemTagsInput
```
LiteSun, line 61: Mantine `TagsInput` splits on commas by default, including pasted text. A valid regex such as `^/v[0-9]{1,3}$` is therefore stored as two tags. Please disable comma splitting for this raw string field (for example with `splitChars={[]}`) and cover a comma-containing regex in the round-trip test.

src/components/form-slice/FormPartUpstream/FormSectionChecks.tsx
```diff
@@ -76,7 +75,10 @@ const FormSectionChecksActive = () => {
         name={np('checks.active.http_path')}
         label={t('form.upstreams.checks.active.http_path')}
       />
-      <FormItemLabels
+      {/* array<string> of raw header lines — a TagsInput, not the Labels
+          widget (which produces an object and broke this field both ways,
+          #3417) */}
+      <FormItemTagsInput
```
LiteSun, line 83: `TagsInput` also splits comma-containing header values. For example, `Accept: text/html,application/json` becomes two array entries, with the second entry no longer a valid header line. Please disable comma splitting for this field and add a round-trip case with a comma in the value.

src/components/form-slice/FormPartSSL/index.tsx
lxbme: Fixed in e0850895 — `skip_mtls_uri_regex` now passes `splitChars={[]}`. The round-trip e2e types `^/v[0-9]{1,3}$` with `pressSequentially` (so the comma keystroke exercises the split path that `fill()` bypasses) and asserts it survives as one tag through save.

src/components/form-slice/FormPartUpstream/FormSectionChecks.tsx
lxbme: Fixed in e0850895 — `http_request_headers` now passes `splitChars={[]}`, same as the regex field. The round-trip e2e types `Accept: text/html,application/json` with `pressSequentially` and asserts it stays a single entry through save.

LiteSun (approved):

guoqqqi (approved):

PR #3437, opened 2026-07-21
fix: unify stream-route submit pipelines, drop phantom plugin_config_id

lxbme:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Part of #3417 (Data-integrity section): "Stream routes: create path bypasses `pipeProduce` (only resource whose create/edit run different cleaning pipelines), and the reused plugins section offers a `plugin_config_id` input the resource schema doesn't have — typed values silently vanish."

**Phantom `plugin_config_id` input (the user-visible bug).** The stream-route form reuses the HTTP-route plugins section, which renders a Plugin Config ID input — but the stream_routes resource has no such field. Verified against a live gateway: the Admin API rejects `plugin_config_id` on stream_routes with 400; in practice the dashboard's zod resolver strips the typed value before the request is built, so whatever the user enters there silently vanishes on a "successful" save and never takes effect. `FormSectionPlugins` gains a `showConfigId` flag (default `true`, HTTP routes unchanged) and the stream-route form turns it off. The regression spec asserts the input is absent on the stream-route add and detail pages and still present on the HTTP-route form.

**Pipeline asymmetry (the hardening).** Stream routes were the only resource whose create and edit paths ran different cleaning pipelines:

- create used a bare `pipe` **without** `pipeProduce` — no `__`-flag removal, no empty-value cleaning, no empty-plugin restore. The only thing keeping form-internal `__` flags out of the request body was the zod resolver's unknown-key stripping — and the Admin API rejects unknown root keys with 400 (verified live), so that protection was load-bearing by accident;
- edit borrowed the **HTTP-route** producer (`produceRoute`), whose `vars` JSON.parse stage is dead code for stream routes.

`produceStreamRoute` now wraps `pipeProduce` (gaining the same explicit defenses every other resource has) and **both** paths use it. Unit tests pin the new behavior (`__` flags stripped at root and nested, empty strings cleaned) and the preserved behavior (name/status deletes, empty-protocol cleanup, inline-upstream removal when a reference id is present, empty-config plugins restored).

One more live-verified correction, disclosed for reviewers: the Admin API nowadays *accepts* `name` on stream routes, so the old "Stream Routes do not support name" comment was outdated — the deletes are kept as defense for values arriving through reused generic components, with the comment corrected.

**Tests** (red on the unfixed build at the intended assertions, green after):

- Unit `FormPartStreamRoute/util.test.ts`: 5 cases (2 red pre-fix: `__`-flag stripping, empty-string cleaning; 3 behavior-preservation guards).
- E2E regression `stream-routes.no-phantom-plugin-config-id.spec.ts`: no Plugin Config ID on stream-route add/detail (plugins section itself still present), input still offered on the HTTP-route form.

Blast radius: full local e2e suite — **175 passed**, all stream_routes suites green; the failures are documented environment items unrelated to this change (two Monaco read-back races that reproduce at the same rate on a pristine master build, one late-serial-run pagination load flake green on isolated rerun, and `stream_routes.show-disabled-error`, which cannot run outside the repo's own compose project). Unit tests 38/38, lint and build clean.

**Related issues**

**Checklist:**

- [ ] Have you modified the corresponding document? (no user-facing document covers the stream-route form)

### DIFF
--- a/src/components/form-slice/FormPartStreamRoute/util.ts
+++ b/src/components/form-slice/FormPartStreamRoute/util.ts
@@ -15,28 +15,33 @@
  * limitations under the License.
  */
 import { produce } from 'immer';
-import { pipe } from 'rambdax';
 
 import { produceRmEmptyUpstreamFields } from '@/components/form-slice/FormPartUpstream/util';
 import { produceRmUpstreamWhenHas } from '@/utils/form-producer';
+import { pipeProduce } from '@/utils/producer';
 
 import type { StreamRoutePostType } from './schema';
 
-export const produceStreamRoute = (val: StreamRoutePostType) =>
-  pipe(
-    produceRmEmptyUpstreamFields,
-    (produceRmUpstreamWhenHas('service_id', 'upstream_id') as unknown as (
-      d: StreamRoutePostType
-    ) => StreamRoutePostType),
-    produce((draft: StreamRoutePostType) => {
-      // Stream Routes do not support name and status
-      const d = draft as StreamRoutePostType & { name?: string; status?: number };
-      delete d.name;
-      delete d.status;
+/**
+ * Shared by BOTH the create and edit paths — stream routes used to be the
+ * only resource whose two paths ran different cleaning pipelines (create
+ * skipped pipeProduce entirely; edit borrowed the HTTP-route producer).
+ * pipeProduce supplies the __-flag removal, empty-value cleaning and
+ * empty-plugin restore every other resource gets (#3417).
+ */
+export const produceStreamRoute = pipeProduce(
+  produceRmUpstreamWhenHas('service_id', 'upstream_id'),
+  produceRmEmptyUpstreamFields,
+  produce((draft: StreamRoutePostType) => {
+    // the form never renders name/status for stream routes; strip them in
+    // case values arrive via reused generic components
+    const d = draft as StreamRoutePostType & { name?: string; status?: number };
+    delete d.name;
+    delete d.status;
 
-      // Cleanup protocol if name is missing
-      if (draft.protocol && !draft.protocol.name) {
-        delete draft.protocol;
-      }
-    })
-  )(val);
+    // Cleanup protocol if name is missing
+    if (draft.protocol && !draft.protocol.name) {
+      delete draft.protocol;
+    }
+  })
+);
--- a/src/components/form-slice/FormPartStreamRoute/util.ts
+++ b/src/components/form-slice/FormPartStreamRoute/util.ts
@@ -33,10 +33,11 @@ export const produceStreamRoute = pipeProduce(
   produceRmUpstreamWhenHas('service_id', 'upstream_id'),
   produceRmEmptyUpstreamFields,
   produce((draft: StreamRoutePostType) => {
-    // the form never renders name/status for stream routes; strip them in
-    // case values arrive via reused generic components
-    const d = draft as StreamRoutePostType & { name?: string; status?: number };
-    delete d.name;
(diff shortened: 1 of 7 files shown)


src/components/form-slice/FormPartStreamRoute/util.ts
```diff
@@ -15,28 +15,33 @@
...
-      delete d.name;
-      delete d.status;
+/**
+ * Shared by BOTH the create and edit paths — stream routes used to be the
+ * only resource whose two paths ran different cleaning pipelines (create
+ * skipped pipeProduce entirely; edit borrowed the HTTP-route producer).
+ * pipeProduce supplies the __-flag removal, empty-value cleaning and
+ * empty-plugin restore every other resource gets (#3417).
+ */
+export const produceStreamRoute = pipeProduce(
```
LiteSun, line 32: This switches Stream Route creation to the current `pipeProduce`, which deep-cleans empty members inside partially non-empty plugin configs and drops inline `upstream.discovery_args: {}`. Those are the behaviors fixed separately by #3438. If this PR lands first, creating a Stream Route can silently change valid user input. Please either make this PR depend on/rebase onto #3438, or include the preservation change and a Stream Route create regression test here.

src/components/form-slice/FormPartStreamRoute/util.ts
```diff
@@ -15,28 +15,33 @@
...
+  produceRmUpstreamWhenHas('service_id', 'upstream_id'),
+  produceRmEmptyUpstreamFields,
+  produce((draft: StreamRoutePostType) => {
+    // the form never renders name/status for stream routes; strip them in
+    // case values arrive via reused generic components
+    const d = draft as StreamRoutePostType & { name?: string; status?: number };
+    delete d.name;
```
LiteSun: The comment correctly notes that the current Admin API accepts Stream Route `name`, but the producer still deletes it and the schema still omits it. A Stream Route created through the API with a name will therefore lose that name after an edit-save in the Dashboard. Please preserve/register `name` and only strip fields that the resource actually does not support.

src/components/form-slice/FormPartStreamRoute/util.ts
lxbme: Rebased this branch onto #3438, so `pipeProduce` here already carries the nested-empty / inline-`discovery_args` preservation. Until #3438 merges, this PR's diff shows those commits on top; once it lands I'll rebase onto master so the diff narrows to just the stream-route changes.

src/components/form-slice/FormPartStreamRoute/util.ts
lxbme: Fixed in 77a7b2ad. Verified against a live gateway: the Admin API accepts `name`/`desc`/`labels` on a stream route (201) and rejects `status` (400). The base `StreamRoute` schema now omits only `status` (the edit form resolves against it, which was why the stored name was stripped before submit), the producer keeps `name` and strips only `status`, and the form renders the name input on create and edit. Unit test: producer keeps name / strips status. e2e: an API-created named stream route keeps its name across a no-op edit-save. This also removed a create/edit asymmetry (create resolved against a schema with `name`, edit against one without).

LiteSun (approved):

src/components/form-slice/FormPartStreamRoute/util.ts
lxbme: #3438 has merged, so this branch is now rebased onto master — the diff is back to just the two stream-route commits.

guoqqqi (approved):

PR #3438, opened 2026-07-21
fix: stop deep-cleaning inside plugin configs, keep nested discovery_args

lxbme:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Part of #3417 (Data-integrity section, last remaining item): "Empty-value stripping inside surviving plugin configs: `produceRestoreEmptyPlugins` only restores whole-plugin entries; nested meaningful empties (`{"headers": {}}`, `[]`, `""`) are still silently removed. Inline route/service `upstream.discovery_args: {}` also still dropped (the #3376 merge-back fix covered the upstreams page only)."

The submit pipeline deep-cleans every empty value (`{}` / `[]` / `""` / `null`) and then restored only WHOLE plugin entries that had vanished. Consequences, both verified against a live gateway:

- empty members **inside** a partially-surviving plugin config were silently removed on any edit-save — the Admin API accepts and stores these shapes for loose-schema plugins (`{"key-auth": {"empty_obj": {}, "empty_arr": [], "empty_str": ""}}` → 201, stored verbatim);
- `discovery_args: {}` on an **inline** upstream (routes/services) was dropped — the #3376 fix only restored the root-level field on the upstreams page.

**Fix.** The clean-then-restore pair is replaced by a single stage, `produceCleanPreservingUserValues`: snapshot the `plugins` subtree, detach it, deep-clean the rest, reattach verbatim, and re-add root/inline `discovery_args: {}` when present. The principle: plugin configs are user-authored JSON and the gateway is the only judge of their empties — loose schemas accept them, strict ones reject with a descriptive 400 (`proxy-rewrite` `headers: {}` → "value should match only one schema, but matches none", verified live).

**Behavior change, disclosed:** typing an empty member that a strict plugin schema rejects now surfaces the gateway's 400 (toast via the existing error path) instead of being silently stripped into a "successful" save that stored something other than what the user wrote.

**Why a single-draft stage — two crashier designs died in testing, disclosed for reviewers:** a restore stage that closes over the pipeline's input value breaks under *nested* `pipeProduce` composition (the routes detail page composes `produceRoute` — itself a `pipeProduce` — as a stage of another `pipeProduce`), because inner stages receive the outer **live immer draft** instead of a plain value: assigning the base reference back trips immer's set trap, and `structuredClone` rejects proxies. Both variants passed unit tests with plain inputs and were caught by the full e2e suite (`routes.crud-all-fields`, `routes.empty-plugin-config` went red). The final design does everything inside one draft with `current()` snapshots and references nothing outside it; a dedicated unit test pins the nested composition.

**Tests** (red on the unfixed build at the intended assertions, green after):

- Unit `producer.test.ts` additions: plugin configs pass through verbatim (array-form nodes included), inline + root `discovery_args` restored, nested-`pipeProduce` composition survives.
- E2E regression `form.plugin-config-nested-empties.spec.ts`: a no-op edit-save keeps `{}` / `[]` / `""` members inside a plugin config, and keeps `discovery_args: {}` on an inline upstream.

Blast radius: full local e2e suite — **177 passed** (the run that caught the two intermediate designs' regressions is what forced the final shape); remaining failures are documented environment items unrelated to this change (one Monaco read-back race non-deterministic on the same build and reproducing on pristine master, one bulk-page load flake green on isolated rerun, and `stream_routes.show-disabled-error`, which cannot run outside the repo's own compose project). Unit tests 37/37, lint and build clean.

**Related issues**

**Checklist:**

- [ ] Have you modified the corresponding document? (no user-facing document covers the submit pipeline)

### DIFF
--- a/src/utils/producer.ts
+++ b/src/utils/producer.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 import { clean, type ICleanerOptions } from 'fast-clean';
-import { produce } from 'immer';
+import { current, isDraft, produce } from 'immer';
 import { pipe } from 'rambdax';
 
 import { produceTime } from './form-producer';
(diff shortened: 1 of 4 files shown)


src/utils/producer.ts
```diff
@@ -102,8 +117,7 @@ export const pipeProduce = (...funcs: ((a: any) => unknown)[]) => {
         ...fs,
         produceRmDoubleUnderscoreKeys,
```
LiteSun: `plugins` is snapshotted only after `produceRmDoubleUnderscoreKeys` has recursively traversed it, so plugin configuration fields beginning with `__` have already been deleted. For example, `{ "my-plugin": { "__mode": "strict" } }` is not passed through verbatim. Since the stated contract is that plugin JSON is owned by the gateway, please skip the `plugins` subtree during UI-key cleanup or detach it before all generic cleaning, and add a regression test.

lxbme: Confirmed and fixed in 75df9865. The `__`-flag removal now runs inside `produceCleanPreservingUserValues`, after the `plugins` subtree is detached — so plugin JSON never sees the `__` cleaner or the deep-clean and passes through verbatim. The separate `produceRmDoubleUnderscoreKeys` pipeline stage (and its now-dead export) is removed. Added a unit regression: `{ "my-plugin": { "__mode": "strict" } }` now survives the pipeline unchanged (red before, green after).

LiteSun (approved):

guoqqqi (approved):

PR #3441, opened 2026-07-21
fix: give each page a distinct localized document title

lxbme:

- [x] Bugfix

Part of #3417 (UX/IA section): "one static `document.title` for every page (router `HeadContent` is already mounted — one `head` option per route away)."

Every page shipped the same static `<title>` (Apache APISIX Dashboard), so tabs, history and bookmarks were indistinguishable. Titles are now derived centrally by a `useDocumentTitle` hook from the current path and the nav route table (the single source of truth already used by the sidebar): section name, plus an Add/Detail qualifier, plus the app name — e.g. `Routes - Apache APISIX Dashboard`, `Add Route - …`, `Route Detail - …`.

**Why central derivation rather than a `head` option on each of the 35 route files:** no per-route boilerplate, and it stays reactive to both navigation and language change — the title re-localizes on a language switch, which per-route `head` evaluation would not. Longest-prefix matching so `/consumer_groups` is not mis-resolved as `/consumers`. `Root` is split into `Root` (provides the i18n context) and `RootShell` (runs the hook inside it).

**Tests.** e2e `router.document-title.spec.ts` asserts distinct section titles, the longest-prefix disambiguation, and the Add qualifier.

Blast radius: full local e2e suite — 178 passed; the only failure is `stream_routes.show-disabled-error`, a documented environment item that cannot run outside the repo's own compose project. Unit tests, lint and build clean.

**Related issues**

**Checklist:**

- [x] Did you explain what problem does this PR solve?
- [ ] Have you modified the corresponding document? (no user-facing document covers page titles)

### DIFF
--- a/src/hooks/useDocumentTitle.ts
+++ b/src/hooks/useDocumentTitle.ts
@@ -18,41 +18,81 @@ import { useRouterState } from '@tanstack/react-router';
 import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
+import type { Resources } from '@/config/i18n';
 import { navRoutes } from '@/config/navRoutes';
 
 // matches the static <title> in index.html (the default before any route
 // resolves and the suffix on every page)
 const APP_NAME = 'Apache APISIX Dashboard';
 
-// Longest `to` first so '/consumer_groups' wins over '/consumers'.
-const sectionsByPathLength = [...navRoutes].sort(
-  (a, b) => b.to.length - a.to.length
-);
+type SourceLabel = keyof Resources['en']['common']['sources'];
+
+// path segment -> sources i18n label. Built from the nav table (the single
+// source of truth), plus `credentials`, which is only ever a nested
+// resource (never a top-level nav entry).
+const segmentToLabel: Record<string, SourceLabel> = {
+  ...Object.fromEntries(
+    navRoutes.map((r) => [r.to.replace(/^\//, ''), r.label])
+  ),
+  credentials: 'credentials',
+};
+
+const isParam = (seg: string) => seg.startsWith('$');
+
+/**
+ * Classify the DEEPEST matched route by its pattern (e.g.
+ * `/services/detail/$id/routes/add`). The action is read from the TAIL of
+ * the pattern so a `detail/$id` in the middle (navigation context) does
+ * not misclassify a nested add/detail page as the parent's detail (#3441
+ * review): a nested `…/routes/add` is "Add Route", not "Service Detail".
+ */
+const classify = (
+  routeId: string
+): { label: SourceLabel; action: 'add' | 'detail' | 'list' } | null => {
+  const segs = routeId.split('/').filter(Boolean);
+  if (segs.length === 0) return null;
+
+  const last = segs[segs.length - 1];
+  let action: 'add' | 'detail' | 'list';
+  let resourceIdx: number;
+
+  if (last === 'add') {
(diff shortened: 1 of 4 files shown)


src/hooks/useDocumentTitle.ts
```diff
@@ -0,0 +1,64 @@
...
+      (r) => pathname === r.to || pathname.startsWith(`${r.to}/`)
+    );
+
+    let title = APP_NAME;
+    if (section) {
+      const name = t(`sources.${section.label}`);
+      const rest = pathname.slice(section.to.length);
+      if (rest.startsWith('/add')) {
+        title = `${t('info.add.title', { name })} - ${APP_NAME}`;
+      } else if (rest.startsWith('/detail')) {
```
LiteSun: The current `rest.startsWith` check classifies every nested page under a parent detail route as that parent detail. For example, `/services/detail/s1/routes/add` and `/services/detail/s1/routes/detail/r1` both get `Service Detail`, so these tabs and bookmarks are still indistinguishable. Please derive the title from the deepest matched route (or otherwise recognize the nested resource/action) and add nested add/detail coverage.

lxbme: Fixed in e7a4e07f. The title is now derived from the deepest matched route's pattern (`useRouterState` → last match `routeId`), with the action read from the tail of the pattern (`/add`, `/detail/$param`, else list) and the resource taken as the segment immediately before that action — so a middle `detail/$id` no longer wins. `/services/detail/$id/routes/add` → "Add Route", `/consumers/detail/$u/credentials/add` → "Add Credential". Added a nested-route e2e case (red on the old logic — it produced "Consumers Detail" — green now).

guoqqqi (approved):

src/hooks/useDocumentTitle.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (segs.length === 0) return null;
+
+  const last = segs[segs.length - 1];
+  let action: 'add' | 'detail' | 'list';
+  let resourceIdx: number;
+
+  if (last === 'add') {
+    action = 'add';
+    resourceIdx = segs.length - 2;
+  } else if (isParam(last) && segs[segs.length - 2] === 'detail') {
```
LiteSun: The new classifier still misses the Secret detail route. `/secrets/detail/$manager/$id` ends with a parameter, but the segment immediately before it is `$manager`, not `detail`, so this condition falls through to the list branch. That branch selects `$manager` as the resource, `segmentToLabel` has no match, and `classify()` returns `null`; the page therefore falls back to the generic application title. Please handle detail routes with multiple trailing parameters and add a regression case for this route.

lxbme: Fixed in ae96880f. `classifyRouteId` now strips *all* trailing params before reading the action from the tail, so `/secrets/detail/$manager/$id` resolves to the secrets detail page regardless of how many params trail it.

Confirmed the old logic was the culprit by running it against that pattern in isolation: it returned `null` (hence the generic title), while the single-param form still worked.

The classifier is now exported and unit-tested against every pattern the generated route tree produces — the two-param secret detail, nested `services/detail/$id/routes/*` and `.../stream_routes/*`, nested credentials, the parent detail pages, and the null fallbacks — so a new route shape can't silently fall back to the app title. The e2e also asserts the real `/secrets/detail/$manager/$id` page title.

LiteSun (approved):

PR #3442, opened 2026-07-21
fix: accessible names for header icon buttons, Monaco contrast, drawer Escape

lxbme:

- [x] Bugfix

Part of #3417 (a11y section). Accessibility quick wins — the app had exactly one `aria-label` in all of `src/`:

- the header's icon-only buttons — settings, language, and the mobile navigation burger — had no accessible name; screen readers announced them as a bare "button". Each gets an `aria-label` from new `a11y.*` i18n keys.
- view-mode Monaco text was gray-5 on the gray-0 editor background, about 2:1 — unreadable on a primary read path. Bumped to gray-7 (~7:1, clears WCAG AA).
- the Select Plugins drawer set `closeOnEscape={false}` despite holding no unsaved state; Escape now closes it like any other dialog.

**Scope, disclosed honestly:** the plugin-card action buttons the issue also lists carry visible text (Add/View/Edit/Delete), so they already have an accessible name — adding an `aria-label` would double-label them. The deeper combobox/`role` semantics rework and the e2e a11y assertion suite the issue mentions are a separate, larger change left for follow-up.

**Tests.** e2e `a11y.icon-buttons-and-escape.spec.ts` covers the accessible names and the drawer's Escape-to-close.

Blast radius: full local e2e suite — 179 passed; the only failure is `stream_routes.show-disabled-error`, a documented environment item that cannot run outside the repo's own compose project. Unit tests, lint and build clean.

**Related issues**

**Checklist:**

- [x] Did you explain what problem does this PR solve?
- [ ] Have you modified the corresponding document? (no user-facing document covers these controls)

### DIFF
--- a/e2e/tests/regression/a11y.icon-buttons-and-escape.spec.ts
+++ b/e2e/tests/regression/a11y.icon-buttons-and-escape.spec.ts
@@ -1,0 +1,51 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


LiteSun (approved):

LiteSun:
Reviewed the accessibility changes and regression coverage. I did not find any blocking issues.

guoqqqi (approved):

Issue #3347, opened 2026-03-21
feat: JSON Schema-driven UI for APISIX Plugin Configuration [GSoC 2026 Proposal]


Baluduvamsi2006:
Feature request

## Problem
APISIX Dashboard plugin configuration is still edited mostly as raw JSON.  
This increases cognitive load for users who are not deeply familiar with each plugin schema, and it can cause invalid or incomplete configurations.

A schema-driven UI can improve usability and correctness by rendering form controls directly from JSON Schema metadata, while still keeping raw JSON fallback for advanced users.

## Alignment note
I am aligning this proposal with existing work in #3274 and I am not proposing a parallel SchemaForm engine.

- Existing aligned PoC baseline: #3274
- Earlier prototype reference from my side: #3346

This issue is now my GSoC tracking/alignment plan for incremental and mergeable work.

## Proposed solution (aligned extension)
I propose a production-focused JSON Schema -> Form UI track by extending and hardening the current direction, with correctness and verifiable deliverables as top priorities.

### Reuse from #3274
- Existing SchemaForm foundation and recursive rendering direction
- Existing conditional rendering baseline (oneOf/anyOf/if-then-else)
- Existing AJV validation direction and demo/test setup

### New focus from my side
- Correctness hardening first (branch switching cleanup + payload consistency)
- Edge-case reliability (nested conditionals, repeated switching, dependencies)
- Incremental integration polish for real plugin flows
- Contributor-oriented documentation and maintenance guidance

## Core architecture (aligned)
1. Schema Resolver Layer
- Parse and normalize plugin schema definitions
- Resolve nested paths, defaults, and required metadata
- Handle APISIX-specific schema usage where practical

2. Field Mapping Layer
- string -> text/password/textarea
- number/integer -> numeric input
- boolean -> switch/checkbox
- enum -> select/radio
- array -> tags/list editor
- object -> grouped/nested section

Enhancements:
- Better label/help extraction
- Default value hydration consistency
- Contributor-friendly extension path

3. Conditional Rendering Engine (correctness-first)
- Improve deterministic handling for:
  - if/then/else
  - oneOf/anyOf
  - dependencies
- Ensure inactive branches are fully cleaned from form state
- Ensure final submit payload strictly matches active schema branch

4. Validation Layer
- Integrate AJV compiled validators
- Normalize errors into field-level + form-level UX
- Accurate error mapping to field paths
- Block unsafe submission on invalid state

[...]

Issue #3146, opened 2025-07-17
bug: cannot add or edit a consumer with a username with a hyphen (`-`) in it while using the new admin ui / dashboard


rctxs:
### Issue description

I tried to add or edit a username of a consumer in the new admin ui / dashboard. 

According to the [changelog ](https://github.com/apache/apisix/blob/release/3.13/CHANGELOG.md#3130) of APISIX 3.13 it is possible to add consumers with usernames with a hyphen (`-`) in it. I've tried that out using the Admin-API and it works. Which is great, thanks for the new feature!

However, it does not work using the new admin ui / dashboard. 

### Expected behavior

The new admin ui / dashboard should support adding consumers with a username with a hyphen in it.

### How to Reproduce

1. Go to the new admin ui / dashboard.
2. Try to add a consumer with a username with a hyphen in it.
3. The ui does not accept the input and marks it as invalid.

### Screenshots

<img width="969" height="594" alt="Image" src="https://github.com/user-attachments/assets/e0321c8a-10b8-4c66-972f-a889ec1ae1f0" />

### Environment

- APISIX Docker Image: apache/apisix:3.13.0-debian
- ETCD Docker Image: bitnami/etcd:3.6

PR #3449, opened 2026-07-27
test: add a gateway-schema contract test for the zod resource schemas

lxbme:

- [x] Bugfix

This implements the "cheapest structural guard" proposed in the "hand-written zod transcription" item of the frontend review in #3417: a CI test asserting the dashboard's zod resource schemas are *looser or equal* to APISIX's own schema, so the recurring "form rejects a config the gateway accepts" class (#3146/#3147/#3362/#3376/#3395) can't silently reappear.

**How it works**

- APISIX exposes its own per-resource JSON schema at `GET /apisix/admin/schema/<resource>`. `scripts/refresh-gateway-schema.mjs` (run via `pnpm refresh:gateway-schema`) snapshots the 11 resources that expose one — route, service, upstream, consumer, ssl, global_rule, plugin_config, consumer_group, stream_route, proto, credential — into `src/types/schema/apisix/__fixtures__/gateway/`. (`secret` is omitted: it is schema'd per-manager and the bare endpoint returns 400.) The snapshot is committed and refreshed on APISIX version bumps; `_meta.json` records the version (currently 3.17.0).
- `src/types/schema/apisix/gateway-contract.test.ts` runs in the unit suite (no gateway needed). For each top-level property the gateway defines it asserts, against the merged dashboard schema: (a) the zod schema knows the field (else the form's resolver silently strips it on submit), and (b) the zod field accepts every value the gateway's enum allows.
- Fields the dashboard does not model live in an `ALLOWLIST` with a reason — intentional non-support, or a KNOWN GAP flagged for follow-up; never a silent omission. A stale allowlist entry (a field zod now covers) also fails the test, forcing cleanup.

**What it already caught**

Populating the allowlist surfaced one genuine "form too strict" gap, plus several intentional differences:

- **KNOWN GAP — `upstream.scheme` omits `"kafka"`**: the gateway accepts a kafka-scheme upstream; the dashboard's zod enum does not, so it cannot be created via the form. Allowlisted with a note and flagged for a follow-up fix (which also needs UI support). This is exactly the class the test exists to surface.
- Intentional differences (allowlisted with reasons): `consumer.id` / `credential.name` (keyed by username / id in the dashboard), `proto.{name,labels,desc}` (content-only form), `consumer_group.name` (no separate name field), `upstream.{create_time,update_time}` (server-managed, not modelled), and `stream_route.name` (the base read schema omits it, but it *is* accepted on the write path via `StreamRoutePostSchema`, per #3437).

**Scope** (kept to the "cheapest guard"): top-level properties only — nested objects and plugin configs (which the dashboard passes through, not transcribes) are out of scope, as is required-ness (the full read schema requires server-managed id/timestamps the write path omits, so comparing it produces only false positives). The longer-term JSON-Schema-driven form work (#3347/#3311/#2986) is unaffected.

**Related issues**

Part of #3417

**Checklist:**

This is a test + fixtures + a dev script only — no application/runtime code changes — so there is nothing for the e2e suite to exercise. Verified with `tsc`, ESLint, and the unit suite (`pnpm test`, which CI now runs). The test's sensitivity was confirmed by adding a fake property to a fixture (goes red naming the field) and by the stale-allowlist check.


lxbme:
Follow-up: the `upstream scheme="kafka"` gap this test flagged is fixed in #3450 (zod enum + scheme dropdown). Once both land, the temporary `ENUM_ALLOWLIST` entry here becomes obsolete and should be dropped on rebase.

guoqqqi (approved):

LiteSun (approved):

